Issue #1, opened 2026-07-29
calendar_list_events is scoped to the current contact, so the agent cannot read a blocking calendar (holidays, closures)


otimizzeai:
> **Edited.** This issue was originally filed as "returns an empty list non-deterministically".
> That framing was wrong — see the correction comment below. The observed behaviour is by design;
> what follows describes the actual gap.

## What the tool does

`calendar_list_events` fences every query to the current contact's own events:

```ts
const params = new URLSearchParams({
  singleEvents: "true",
  orderBy: "startTime",
  maxResults: String(...),
  // Server-side fence: only events stamped for THIS customer. The model has no field to widen it.
  privateExtendedProperty: `${SECV4_CONTACT_KEY}=${stamp}`,
});
```

...and re-verifies the stamp client-side afterwards. That isolation is correct and clearly
deliberate — a customer must never see another customer's appointments.

## The gap

There is no way to read a calendar that the agent must *respect* but does not *own*: public
holidays, staff time off, closures, blackout dates. Any event not created by the agent for the
current contact carries no `secv4Contact` stamp, so it is invisible to every tool the model has.

Concretely: we put Brazilian public holidays on a dedicated calendar so the agent would refuse to
book on them. `calendar_list_events` returns `[]` for that calendar every time, and the agent
cannot tell "no holiday" from "I am not allowed to see holidays".

## Why it is worth surfacing

An empty result is not a neutral outcome for an LLM. When the tool returns `[]`, the model
sometimes asserts the holiday anyway from its own world knowledge, producing a *correct-looking*
answer that is not grounded in the calendar. That failure is invisible when reading conversations
— it only shows up in the execution log, where the tool returned nothing.

## Possible directions

1. A read-only tool for operator-designated "blocking" calendars, returning busy periods only
   (no titles, no attendees) — enough to refuse a slot without leaking anything.
2. Or make it explicit in the tool description that the listing is scoped to the customer's own
   appointments, so the model stops treating `[]` as evidence about the calendar as a whole.

## Note on the workaround

`calendar_check_availability` uses `/freeBusy` on the selected calendar with **no** contact fence,
so busy events placed directly on the booking calendar do block slots. That covers the holiday
case as long as the blocking events live on the same calendar the agent books into — which is a
usable workaround, but it does mean holidays and patient appointments have to share a calendar.

otimizzeai:
**Correction to the original report.**

I filed this as non-deterministic behaviour, based on four calls against the same calendar where
[...]

PR #5, opened 2026-07-29
fix(calendar): clear the opposite time field when patching an event (all-day ⇄ timed)

### LINKED ISSUE #2: `calendar_update_event` cannot convert an all-day event into a timed event (HTTP 400)
## Steps

1. Create (or import) an all-day event — Google returns `start: "2026-09-07"`, `end: "2026-09-08"`.
2. Call `calendar_update_event` on that event id passing explicit datetimes:
   - `start = 2026-09-07T00:00:00-03:00`
   - `end   = 2026-09-07T23:59:00-03:00`

## Result

```
Google Calendar rejected the update (HTTP 400).
```

## Expected

Either the conversion succeeds, or the error explains that all-day ↔ timed conversion
is not supported by the tool.

## Context

Google's API requires clearing `start.date` / `end.date` when setting `start.dateTime` /
`end.dateTime`. A patch that sets only the datetime fields is rejected. Creating a new
timed event works normally, so only the update path is affected.

otimizzeai:
Fixes #2.

`toEventTime` returns either `{ date }` or `{ dateTime, timeZone }`. On update the result goes
straight into a PATCH body, and PATCH merges: setting only `dateTime` leaves the event's existing
`date` untouched. Google rejects an event carrying both, so the call fails:

```
Google Calendar rejected the update (HTTP 400).
```

An all-day event can therefore never be converted to a timed one, nor the reverse. Creating a new
timed event works normally, so only the update path is affected.

## The change

A patch-specific variant that nulls the field it replaces — Google treats an explicit `null` as
"clear this" — leaving `toEventTime` alone for the create path:

```ts
function toEventTimePatch(value: string, timeZone: string): Record<string, string | null> {
  const t = toEventTime(value, timeZone);
  return "date" in t ? { ...t, dateTime: null } : { ...t, date: null };
}
```

## Tests

Two cases added to `tests/modules/toolpacks-google-calendar.test.ts`, covering both directions and
asserting on the PATCH body (`calls[1]`, after the ownership re-fetch).

Verified with `bun test tests/modules/toolpacks-google-calendar.test.ts`:

- with the fix: **41 pass, 0 fail**
- with the original source and the new tests: **39 pass, 2 fail** — the two new cases, so they do
  pin the behaviour rather than passing vacuously

## Where this came up

We import public holidays into the professionals' calendars as full-day busy blocks so
`calendar_check_availability` refuses those days. Converting an imported all-day holiday into a
timed 00:00–23:59 block is what first hit the 400.

### DIFF
--- a/src/modules/integrations/toolpacks/google-calendar.ts
+++ b/src/modules/integrations/toolpacks/google-calendar.ts
@@ -268,6 +268,18 @@ function toEventTime(value: string, timeZone: string): Record<string, string> {
   return { dateTime: v, timeZone };
 }
 
+// PATCH merges what we send, so a patch carrying only `dateTime` leaves the event's existing `date`
+// in place — and Google rejects an event holding both (HTTP 400), which makes every all-day ⇄ timed
+// conversion fail. Sending the opposite field as null clears it, so the patch replaces the time
+// representation instead of mixing the two.
+function toEventTimePatch(
+  value: string,
+  timeZone: string,
+): Record<string, string | null> {
+  const t = toEventTime(value, timeZone);
+  return "date" in t ? { ...t, dateTime: null } : { ...t, date: null };
+}
+
 // Flattens a Calendar start/end object to a single string for the model (dateTime or all-day date).
 function flattenTime(t: unknown): string | null {
   if (!t || typeof t !== "object") return null;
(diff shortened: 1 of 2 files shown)


gabrieljablonski:
Merged — thank you for the sharp diagnosis and for tests that pin the PATCH body; that made review straightforward (reproduced locally: 39 pass + 2 fail without the fix, 41 pass with it).

For transparency on how integration works here: this public repo is generated from an internal master, and your change has been reintegrated there with your authorship preserved. The next `Sync free (agents@…)` commit regenerates the exact tree your PR produced, so nothing gets clobbered and your merge stays in this repo's history.

We're also triaging your other issues — #1 (a readable blocking calendar) is a real product gap and is on our radar.

PR #38, opened 2026-08-07
fix(appointments): lembrete e turnos seguintes carregam a identidade do evento

### LINKED ISSUE #22: Proactive appointment triggers lose the appointment context on the next turn
When `appointmentReminders` fires, the runtime knows exactly which appointment it is reminding about — it cannot pick the moment otherwise, since the offsets are computed from the event itself. That context does not survive into the next turn. So the customer replies to the reminder, and the agent has no idea which appointment they mean.

The result reads badly to the customer: the assistant messages you about your appointment, and then asks you which appointment you are talking about.

### What the logs show

One conversation, in order (times local, one instance):

```
day 1  20:00   generate   trigger: appointment_reminder   -> messaged
day 2  19:00   generate   trigger: appointment_reminder   -> messaged
day 2  21:00   generate   trigger: followup (step 1)      -> messaged
day 2  21:07   tool       calendar_list_events            -> refused
day 2  21:07   generate   -> asks the customer which appointment they mean
```

Seven minutes after a follow-up, and two hours after a reminder about that very appointment, the agent could not identify it.

### Why the agent cannot recover it

`calendar_list_events` operates on one calendar per call. With several calendars configured (one per professional, which is the normal shape for a clinic), `calendarId` stops being optional:

```
Multiple calendars are configured; set calendarId.
Allowed calendars: "<A>", "<B>", "<C>", "<D>" (pass calendarId by name or id)
```

There is no cross-calendar lookup. To find "this customer's appointment" the agent has to call the tool once per calendar — four calls out of a `maxToolCalls` budget of 10, before it has cancelled anything, checked availability or created the replacement — or ask the customer which professional it was. It asks.

### Why this cannot be solved in the prompt

I looked for a prompt-level workaround before filing this.

`buildPromptVars` (`src/graph/prompt.ts`) is a closed allowlist: `contact_name`, `contact_first_name`, `contact_email`, `contact_phone`, `inbox_name`, `company_name`, `agent_name`, plus the time variables — with their pt-BR aliases. Unknown placeholders are deliberately left literal, so inventing one prints it to the operator instead of leaking or blanking. There is no appointment, calendar, event or conversation-attribute variable.

Persisting the id at creation time does not work either: `set_custom_attribute` exists, but there is no read counterpart in the native tool set, and contact/conversation attributes are not part of the turn context. The agent would write to a drawer it cannot open.

Carry the triggering event's identity into the turn it starts — calendar id, event id, start time, calendar label — for `appointment_reminder` and for appointment-related `followup` steps. The agent then answers "your appointment on Thursday at 09:00 with Dr. <name> — what time would you like instead?" and reschedule/cancel need no lookup at all, because the event id is already in hand.

That seems cheaper than the alternative, which would be letting `calendar_list_events` search across the allowed calendars: the data already exists at trigger time, it just does not survive one hop.

If either already exists and I missed it, that is the more likely answer and I am happy to be pointed at it.

### Environment

gabrieljablonski:
## Problem

When `appointmentReminders` fires, the runtime knows exactly which event it is reminding about — the `APPOINTMENT_REMINDER` job payload carries `eventId`/`calendarId`/`startISO`, and the handler even GETs the event to verify it is still alive. But that identity died one call short of the model: `reminderNudge(...)` only received `summary` + `startISO`. So the reminder turn had no ids (the final reminder's instruction literally asked the agent to "call calendar_confirm_appointment with the event id" — an id it was never given), and the turn where the customer replies to the reminder had even less. Recovery via `calendar_list_events` is not viable by design: with several calendars `calendarId` is mandatory per call, and the query is fenced to events the agent itself created.

## Fix

Two legs, both fed from data we already store — no Google call is added to any turn:

1. **The nudge carries the identity.** `AgentNudge` gains `refs` (opaque values the agent may need as tool arguments), rendered INSIDE the `⟦external-data⟧` fence with the same sanitization as every fenced field. `reminderNudge` passes `event_id`/`calendar_id` and its instructions now point at those keys by name. Nothing model-facing is interpolated into the trusted instructions lane.
2. **Every later turn re-reads the appointment identity.** `loadAgentConfig` appends a per-turn context block (sibling of the Chatwoot attribute section) listing the conversation's LIVE appointments — `event_id`, `calendar_id`, calendar label, start, summary — projected from the reminder scheduler rows by `threadId`. Liveness handles the tricky case the report is about: after the LAST reminder fires there is no `PENDING` row left, so a `DONE`/`DEAD` row still counts while its start is in the future. The block reaches the reactive turn, the debounce flush and every nudge (shared `loadAgentConfig`), and the tool pointer ("use calendar_update_event with this event_id…") is emitted only when those tools are actually granted.

Supporting changes:

- The job payload is enriched at (re)arm time with `summary` and `calendarLabel` (snapshot; a rename made directly in Google goes stale until a reschedule re-arms — accepted tradeoff, no per-turn Google call).
- `cancelAppointmentReminders` now tombstones every row of the event (`payload.cancelledAt`): cancelling marks jobs `DONE`, which is otherwise indistinguishable from "fired", and without the stamp a cancelled appointment would keep resurfacing in the context block until its start passed. A reschedule re-arm replaces the payload wholesale, clearing the stamp on surviving offsets.
- Confirmation state is deliberately NOT in the block (it lives in the event's private properties on Google; the conversation history already records the `calendar_confirm_appointment` call).

## Tests

Red-first, mirroring the report: the projection (`DONE` + future start ⇒ live; tombstone ⇒ never; freshest payload wins; instant-based, not lexicographic, start comparison), the section builder (escaping, grant-gated tool pointer), the nudge refs (inside the fence, hostile values cannot break out, instructions reference the key but never contain the raw id), payload enrichment, and DB-backed end-to-end: a conversation whose only reminder row is `DONE` with a future start gets the block in `systemPrompt` (the exact "customer replies to the last reminder" turn), and a cancelled appointment does not.

Fixes #22

## Summary by CodeRabbit

* **New Features**
  * Agent conversations can now include relevant upcoming appointment details and calendar guidance.
  * Appointment reminders retain event summaries, calendar labels, and references for confirmation, rescheduling, and cancellation.
  * External appointment references are safely included in agent prompts.

* **Bug Fixes**
  * Cancelled appointments are excluded from conversation context, including previously completed reminders.
  * Appointment data is preserved when live calendar enrichment is unavailable.
  * Appointment context is sorted and filtered for more accurate, timely responses.

### DIFF
--- a/src/graph/nudge.ts
+++ b/src/graph/nudge.ts
@@ -45,9 +45,9 @@ export interface AgentNudge {
   value?: number | null;
   currency?: string | null;
   summary?: string | null;
-  // Opaque external references the agent may need as TOOL ARGUMENTS (event id, calendar id, …).
-  // Rendered INSIDE the data fence as extra k=v facts — sanitized like every fenced field, and never
-  // appended to the instructions lane (which is trusted operator/code text).
+  // NOTE: Opaque external references the agent may need as TOOL ARGUMENTS (event id, calendar id,
+  // …). Rendered INSIDE the data fence as extra k=v facts — sanitized like every fenced field, and
+  // never appended to the instructions lane (which is trusted operator/code text).
   refs?: Record<string, string | null | undefined>;
   instructions?: string;
   // For a follow-up sequence: the 1-based step that fired. Surfaced on the conversation timeline
--- a/src/graph/nudge.ts
+++ b/src/graph/nudge.ts
@@ -45,6 +45,10 @@ export interface AgentNudge {
   value?: number | null;
   currency?: string | null;
   summary?: string | null;
+  // Opaque external references the agent may need as TOOL ARGUMENTS (event id, calendar id, …).
+  // Rendered INSIDE the data fence as extra k=v facts — sanitized like every fenced field, and never
+  // appended to the instructions lane (which is trusted operator/code text).
+  refs?: Record<string, string | null | undefined>;
   instructions?: string;
   // For a follow-up sequence: the 1-based step that fired. Surfaced on the conversation timeline
   // ("Follow-up N enviado") and in the flow log. Undefined for non-sequenced nudges (inbound events).
@@ -152,6 +156,15 @@ export function renderNudge(
     );
   }
   if (n.summary) facts.push(`summary=${sanitizeFreeText(n.summary, 500)}`);
+  if (n.refs) {
+    for (const [key, value] of Object.entries(n.refs)) {
+      if (value) {
+        facts.push(
+          `${sanitizeFreeText(key, 40)}=${sanitizeFreeText(value, 200)}`,
+        );
+      }
+    }
+  }
   const directive = canMessageCustomer
     ? `An external system event just occurred for this conversation. By default, send a brief, warm, helpful proactive message to the customer about it — keep it short and natural, in the conversation's language. Lean toward reaching out: a timely follow-up is usually welcome. Stay silent ONLY if a message would clearly be unhelpful, premature, duplicated, or annoying; in that rare case reply with EXACTLY ${FOLLOWUP_SKIP_SENTINEL} and nothing else.`
     : `A human agent is currently handling this conversation. Do NOT message the customer. If the event is worth flagging, write a short internal note for the human; otherwise reply with EXACTLY ${FOLLOWUP_SKIP_SENTINEL} and nothing else.`;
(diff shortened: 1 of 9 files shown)


gabrieljablonski:
@coderabbitai review

gabrieljablonski:
@coderabbitai full review

gabrieljablonski:
@coderabbitai full review

gabrieljablonski:
@coderabbitai review

PR #43, opened 2026-08-09
fix(followups): supressão por agendamento enxerga o compromisso após o último lembrete

### LINKED ISSUE #39: pauseWhileAppointment suppression goes blind after the last reminder fires
## Description

`followUp.pauseWhileAppointment` (default on) suppresses re-engagement while the conversation has a known FUTURE appointment. Both places that implement it anchor on **`PENDING`** `APPOINTMENT_REMINDER` rows only:

- the sweep's eligibility SQL (`src/modules/followups/handlers.ts`, the `NOT (... EXISTS (SELECT 1 FROM scheduler_jobs ... AND sj.status = 'PENDING' ...))` clause);
- `hasPendingAppointmentReminder` (`src/modules/appointments/reminders.ts`), used by the handler's re-check and the operator-facing estimate.

But firing a reminder marks its row `DONE`. After the **last** configured reminder fires there is no `PENDING` row left, so the suppression silently turns off while the appointment is still in the future — the follow-up ladder resumes chasing a customer who is about to show up.

The window is `[last reminder fires, appointment starts]` — exactly the offset of the smallest configured reminder (a `1`-hour last offset leaves a 1-hour hole; a `24`-hour single offset leaves a full day).

## Observed

The log in #22 shows it happening (times local, one instance):

```
day 2  19:00   generate   trigger: appointment_reminder   -> messaged   (last reminder)
day 2  21:00   generate   trigger: followup (step 1)      -> messaged   (appointment still ahead)
```

## Fix direction

#38 introduced exactly the pieces this needs:

- `projectAppointmentEvents` (`src/modules/appointments/context.ts`) — the reusable liveness predicate: an appointment is live while some row is `PENDING`/`CLAIMED` **or** a `DONE`/`DEAD` row's `startISO` is still in the future, excluding tombstoned rows;
- the `payload.cancelledAt` tombstone stamped by `cancelAppointmentReminders` — without it, "DONE with future start" would also match appointments the customer just cancelled (cancelling marks rows `DONE` too; there is no `CANCELLED` status).

So: `hasPendingAppointmentReminder` becomes "has live appointment" via the shared predicate, and the sweep's `EXISTS` accepts `DONE`/`DEAD` rows with `(sj.payload->>'startISO')` still ahead **and** no `cancelledAt` — with care for the all-day date format in the SQL cast, and red tests for the post-last-reminder window and for the cancelled case.

Note: the same sweep SQL is also touched by the fix for #27, so whichever lands second rebases over the other.

## Environment

Found while investigating #22 on `main`; behavior is version-independent (the suppression has anchored on `PENDING` since it was introduced).

gabrieljablonski:
## Symptom

`followUp.pauseWhileAppointment` (default on) is supposed to suppress re-engagement while the conversation has a FUTURE appointment. Both implementing sites anchored on `PENDING` `APPOINTMENT_REMINDER` rows only — but firing a reminder marks its row `DONE`, so after the **last** configured reminder fires there is no `PENDING` row left and the suppression silently turns off while the appointment is still ahead. The follow-up ladder resumes chasing a customer who is about to show up; the hole is exactly the smallest configured offset (the log in #22 shows a `followup step 1` two hours after the last reminder, with the appointment still in the future).

## Fix

Both sites now share the liveness predicate #38 introduced (`projectAppointmentEvents`): an appointment is live while some non-tombstoned reminder row is still queued (`PENDING`/`CLAIMED`) **or** its `startISO` is still ahead — `payload.cancelledAt` tombstones excluded, so a cancelled appointment (whose rows are also `DONE`) never suppresses.

- **Sweep** (`src/modules/followups/handlers.ts`): the `NOT EXISTS` clause becomes a SQL mirror of the predicate. Three deliberate details:
  - `startISO` can be an all-day date (`YYYY-MM-DD`) — normalized to `T00:00:00Z` so SQL agrees with JS `Date.parse` (UTC midnight) instead of drifting by the session TimeZone;
  - the cast is guarded (`CASE` + `pg_input_is_valid`, available on the pg17 the deploy mandates) because `startISO` can also be model-supplied garbage and an unguarded `::timestamptz` would abort the whole tenant's sweep — one bad payload row would kill follow-ups for everyone. A guard test pins this;
  - invalid/absent start = not-future (fail-safe, same as the JS `Number.isFinite` check): only the queued arm suppresses then.
- **Handler re-check** (`src/modules/appointments/reminders.ts`): `hasPendingAppointmentReminder` → **`hasLiveAppointment`**, now delegating to `loadAppointmentContext` (the shared predicate) instead of a PENDING-only query. Same signature, single call site; the 1h re-check backoff now genuinely holds the sequence until the start passes or the appointment is cancelled.

`FAILED` is deliberately not added to the queued arm: the scheduler never writes it (`failJob` → PENDING/DEAD, reaper → PENDING/DEAD), and the shared predicate treats any non-queued status as live only via the future-start arm — the SQL mirrors that exactly.

## Tests (written red-first)

In `tests/modules/followup-handler.test.ts` (sweep invoked via `getJobHandler("FOLLOWUP_SWEEP")`):

- **(q)** handler: `DONE` + future start → `reschedule`, nothing sent (the #22 log shape) — red before the fix;
- **(r)** sweep: `DONE` + future start AND a `CLAIMED` row both suppress — red before the fix;
- **(s)** sweep: `DONE` + past start is swept (follow-up resumes after the appointment) — boundary guard;
- **(t)** sweep: fully tombstoned event (cancelled, start still ahead) is swept — cancellation guard;
- **(u)** sweep: a garbage `startISO` in the same tenant neither aborts the sweep nor suppresses — the unguarded-cast regression;
- **(v)** sweep: all-day (`YYYY-MM-DD`) future start suppresses — red before the fix.

Existing (m)/(n) keep passing untouched and now pin the queued arm (a `PENDING` row without `startISO` still suppresses) and the opt-out. Two direct `hasLiveAppointment` tests in `tests/modules/appointment-context-db.test.ts` cover the wrapper (DONE+future → true; cancelled → false).

## Notes

- The operator-facing follow-up **estimate** (`conversations/service.ts`) has never applied appointment suppression — a pre-existing, orthogonal divergence (it can promise a follow-up the sweep will suppress). Left out deliberately; the sweep comment no longer claims parity with it. Worth its own issue if the estimate should learn about appointments.
- `DONE` reminder rows accumulate (there is no scheduler_jobs retention); the EXISTS rides the `(tenant_id, kind, dedupe_key)` index prefix exactly as before, with the new per-row expressions only evaluated on that already-scanned range. Pre-existing and shared with the per-turn context loader; also worth a separate look someday.

Fixes #39.

## Summary by CodeRabbit

- **Bug Fixes**
  - Follow-ups remain suppressed for live future appointments after reminders are completed or claimed.
  - All-day and offset-less appointment dates are handled consistently as UTC.
  - Follow-ups resume when an appointment starts or is canceled.
  - Invalid, missing, or impossible appointment dates no longer interrupt processing or incorrectly suppress follow-ups.
  - Fired reminders continue to suppress follow-ups when the appointment has not started.

- **Tests**
  - Added coverage for appointment status changes, cancellations, future dates, all-day appointments, and varied or malformed date formats.

## Validation scope

No live Chatwoot validation in this round: no Chatwoot call changes — the fix is sweep SQL + an internal JS predicate. Unit + DB-backed e2e (the sweep harness above) close the loop to the observable effect (FOLLOWUP enqueued or suppressed).

### DIFF
--- a/src/modules/appointments/context.ts
+++ b/src/modules/appointments/context.ts
@@ -46,6 +46,26 @@ function cleanText(v: unknown, max: number): string | null {
   return s || null;
 }
 
+// NOTE: One shared time-zone rule for startISO values WITHOUT an offset, mirrored by the follow-up
+// sweep's SQL normalization (followups/handlers.ts): all-day dates and offset-less datetimes are
+// pinned to UTC. Date.parse already reads a bare date as UTC midnight but reads an offset-less
+// DATETIME in the host zone — while Postgres would use the session TimeZone — so without the pin the
+// two liveness decisions could diverge when app and DB zones differ. startISO can carry an
+// offset-less datetime only via the model-input fallback in the Calendar toolpack (Google always
+// returns an offset); UTC is arbitrary there, but agreement matters more than the boundary instant.
+export function parseStartMs(startISO: string): number {
+  if (/^\d{4}-\d{2}-\d{2}$/.test(startISO)) {
+    return Date.parse(`${startISO}T00:00:00Z`);
+  }
+  if (
+    /[Tt ]\d{2}:/.test(startISO) &&
+    !/(?:[Zz]|[+-]\d{2}:?\d{2})$/.test(startISO)
+  ) {
+    return Date.parse(`${startISO}Z`);
+  }
+  return Date.parse(startISO);
+}
+
 // Pure: scheduler rows → the LIVE appointments of a conversation. A row counts toward an event when
 // it is not tombstoned; an event is live when some row is still queued (PENDING/CLAIMED — the
 // reminder turn itself runs on a CLAIMED row) or when its start is still ahead of `now`. The freshest
@@ -82,7 +102,7 @@ export function projectAppointmentEvents(
       typeof winner.payload.startISO === "string"
         ? winner.payload.startISO
         : "";
-    const startMs = Date.parse(startISO);
+    const startMs = parseStartMs(startISO);
     const queued = group.rows.some(
       ({ row }) => row.status === "PENDING" || row.status === "CLAIMED",
     );
--- a/src/modules/appointments/context.ts
+++ b/src/modules/appointments/context.ts
@@ -46,6 +46,21 @@ function cleanText(v: unknown, max: number): string | null {
   return s || null;
 }
 
+// NOTE: Date.parse rolls impossible calendar dates over ("2026-02-30" parses as March 2) while the
+// SQL mirror's pg_input_is_valid REJECTS them — without this guard an impossible startISO would look
+// upcoming to the JS re-check but never to the sweep. Reject the roll-over up front: NaN, like garbage.
+function hasImpossibleDateParts(startISO: string): boolean {
+  const m = /^(\d{4})-(\d{2})-(\d{2})(?:[Tt ]|$)/.exec(startISO);
+  if (!m) return false;
+  const [y, mo, d] = [Number(m[1]), Number(m[2]), Number(m[3])];
+  const roundTrip = new Date(Date.UTC(y, mo - 1, d));
+  return (
+    roundTrip.getUTCFullYear() !== y ||
+    roundTrip.getUTCMonth() !== mo - 1 ||
+    roundTrip.getUTCDate() !== d
+  );
+}
+
 // NOTE: One shared time-zone rule for startISO values WITHOUT an offset, mirrored by the follow-up
 // sweep's SQL normalization (followups/handlers.ts): all-day dates and offset-less datetimes are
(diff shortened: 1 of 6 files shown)


gabrieljablonski:
@coderabbitai review

gabrieljablonski:
@coderabbitai review

gabrieljablonski:
@coderabbitai review

gabrieljablonski:
@coderabbitai review

gabrieljablonski:
@coderabbitai review

gabrieljablonski:
@coderabbitai review

PR #44, opened 2026-08-09
fix(flowlog): falha de integração vira linha warn/error alertável no log de tools

### LINKED ISSUE #40: Integration failures are logged as successful tool calls (level: info, status: ok), so no alert channel can fire on them
An integration that fails at runtime is recorded in `execution_logs` as a **successful** tool call — `level: "info"`, `status: "ok"` — with the provider's failure message sitting in `detail.output` as if it were a normal result. Because alert channels filter by level, nothing can be built on top of this: a channel set to `warn`/`error` never fires, and a channel set to `info` fires on every tool call ever made.

### Mechanism

`src/graph/tool-flowlog.ts` has two callbacks, and the distinction between them is whether the tool **threw**:

```ts
override handleToolEnd(output: unknown, runId: string): void {
  ...
  emitFlowEvent(this.flow, {
    stage: "tool",
    level: "info",
    status: "ok",
    detail: { tool: s.tool, args: s.args, output: toolOutputValue(output) },
  });
}

override handleToolError(err: unknown, runId: string): void {
  ...
  emitFlowEvent(this.flow, {
    stage: "tool",
    level: "warn",
    status: "error",
    ...
  });
}
```

The toolpacks, however, handle provider failures by **returning a string**, not by throwing — which is the right call for the model-facing side, since the agent should degrade gracefully rather than blow up mid-conversation:

```ts
// src/modules/integrations/toolpacks/google-drive.ts:206
if (res.status < 200 || res.status >= 300) {
  return `Google Drive returned HTTP ${res.status}.`;
}
```

Same shape at `google-drive.ts:202` (`"Failed to reach Google Drive. Try again shortly."`), `google-drive.ts:264`, and `asaas.ts:218` (`"The payment provider rejected the request (HTTP ${res.status})."`).

So every one of those lands in `handleToolEnd` and is stamped `ok`. The graceful-degradation string is doing double duty: it is the message to the model **and** the only record of the failure, and the record says nothing went wrong.

### The one case that does work, and why it matters

Credential expiry *is* caught correctly. `src/modules/vault/google-oauth.ts:223` throws an `AppError` rather than returning, so it reaches `handleToolError` and is logged `warn`/`error`:

```
google token endpoint error: invalid_grant
```

That is the behaviour the rest of the failure paths seem to intend, which is why this reads like an oversight rather than a decision.

### Evidence from a deployed instance

Every tool-stage failure over a 15-day window, taken from `execution_logs`:

| tool | message | logged as |
|---|---|---|
| `calendar_update_event` | `Google Calendar rejected the update (HTTP 400).` | `info` / `ok` |
| `asaas_create_pix_charge` | `The payment provider rejected the customer (HTTP 400).` | `info` / `ok` |
| `calendar_check_availability` | `google token endpoint error: invalid_grant` | `warn` / `error` |
| `calendar_check_availability` | `google token endpoint error: invalid_grant` | `warn` / `error` |
| `calendar_check_availability` | `Google Calendar returned HTTP 403.` | `info` / `ok` |
| `drive_find_file` | `Google Drive returned HTTP 403.` | `info` / `ok` |
| `drive_find_file` | `Google Drive returned HTTP 403.` | `info` / `ok` |

Five of seven are invisible to any level-based filter. Worth noting the volume: **seven failures in fifteen days**, against thousands of successful tool calls. This is not a noisy signal that needs suppressing — it is a rare, high-value one that currently cannot be subscribed to at all.

### Why it matters

The failure mode is silent, and the customer discovers it first.

On this instance an OAuth credential expired mid-conversation. The agent told a lead twice that it could not check the calendar, the lead waited, then abandoned the conversation. Nobody on the operator side was notified, because the last availability check before the token died had already been logged as `ok`, and the operator had no way to subscribe to the ones that weren't. The failure was fully machine-detectable at the moment it happened; there was simply no level for it to be detected *at*.

Integration failures are also the class of failure an operator most needs pushed to them rather than polled for: they are caused by something outside the product (a revoked token, a provider outage, a rejected payload), they do not self-heal, and every conversation that touches that integration keeps failing until a human intervenes.

The model-facing behaviour should not change — returning a friendly string is correct. What is missing is a way for the toolpack to say "this return value is a failure" so the callback can pick a different level.

A small branded helper would cover every existing call site without touching the graceful-degradation contract, e.g. a `toolFailure(message)` that returns the same string but is recognisable in `handleToolEnd`, which would then emit `level: "warn"`, `status: "error"` and carry the message into `errorMessage`. Any equivalent mechanism works — the requirement is only that the emitted event be distinguishable by level.

That alone makes the existing alert-channel machinery useful for this, with no changes to it.

### Environment

gabrieljablonski:
## Symptom

An integration that fails at runtime is recorded in `execution_logs` as a **successful** tool call — `level: "info"`, `status: "ok"` — with the provider's failure message sitting in `detail.output` as a normal result. Alert channels filter by level, so a channel set to `warn`/`error` never fires on an expired OAuth credential, a provider outage, or a rejected payload. The customer finds out first (issue #40 documents a lead abandoning a conversation after two failed availability checks nobody was alerted about).

## Cause

`ToolFlowLogger.handleToolEnd` stamps every returned output `info`/`ok`; only tools that **throw** reach `handleToolError` (`warn`/`error`). The toolpacks deliberately return friendly strings instead of throwing — correct model-facing behavior (graceful degradation), but the string was doing double duty as the only record of the failure, and the record said nothing went wrong.

## Fix

The model-facing contract is unchanged; the failure is now *marked* on its way out:

- **`src/graph/tools/failure.ts`** (new): `toolFailure(message)` wraps a friendly string as an integration failure; `failableTool(...)` (a `tool()` wrapper) converts it into a `ToolMessage` with `status: "error"` and the same string as content, using LangChain's direct-tool-output passthrough — the message survives intact through `ToolNode` to the model AND reaches `handleToolEnd`. Without a `tool_call` in scope (plain-args invocation, e.g. tests) it degrades to the plain string, exactly today's behavior.
- **`src/graph/tool-flowlog.ts`**: `handleToolEnd` reads `output.status === "error"` → emits the ONE tool line as `level: "warn"`, `status: "error"`, with the string as `errorMessage`. `warn` matches `handleToolError` and the MCP discovery-failure precedent, and passes both the dispatch gate (`warn`/`error` + inbox) and `minLevel` filtering.

### What is marked as a failure

- **google-calendar / google-drive / asaas toolpacks**: network failures, non-2xx provider responses, malformed provider responses, and missing/unresolvable credentials (`NOT_CONNECTED` / "not configured" — operator config, exactly what a channel should surface).
- **Operator-authored HTTP tools**: **every non-2xx** (`HTTP <status>` body unchanged). For a declarative tool, 401/403 mean the operator's credential is broken and 5xx/429 mean the provider is down or throttling — the alert-channel use case. Bursts are contained by the existing coalescing (one pending delivery per channel/stage/level with a `count`). If a deployment's API legitimately uses 404-as-data and this gets noisy, a per-tool "expected statuses" opt-out is the natural follow-up lever — not built now.
- **`react_to_message`**: the `catch` that swallowed a real Chatwoot API failure now marks it.

### What stays `info`/`ok` (deliberately)

Business-level degradation and model input errors are normal operation, not outages: empty availability slots, playground/context gates, invalid ranges, invalid CPF/CNPJ, Asaas id-not-found guidance (recoverable model input), policy limits (blocking-calendar caps, file too large), "already cancelled" idempotency.

MCP tools were already correct (`isError` → `ToolException` → `handleToolError`) and are untouched.

## Tests

- `tests/graph/tool-failure.test.ts`: the wrapper contract (tool_call → `ToolMessage` error with id/name; plain args → plain string; success untouched).
- `tests/graph/tool-flowlog.test.ts`: info/ok pin for plain strings; ONE warn/error line with `errorMessage` for a marked failure; an e2e through `buildAgentGraph` with a scripted model proving the second model call receives the **exact** friendly string (model-facing unchanged) while the flow log records exactly one alertable line; and the issue's full loop — a marked failure produces an `alert_deliveries` row for a `minLevel: warn` channel (before this fix no channel could ever produce one).
- Failure-marking cases per pack (written red-first against the old behavior): calendar HTTP 500 + not-connected, asaas not-configured, drive network + not-connected, HTTP tool 500/404 → error with 200 staying success — plus guards pinning that bad model input (range >24h, invalid CPF) is NOT marked.

## Notes

- Adapter behavior for `ToolMessage.status`: OpenAI-family and Anthropic adapters ignore it (byte-identical requests). `@langchain/google-genai` wraps error-status content as `functionResponse.response.error.details` — same string, semantically coherent framing.
- The playground trace already maps `status === "error"` to its `isError` flag, so marked failures now render as failures there too.
- A sibling class of problems — failures swallowed with a **successful** tool return (orphaned Asaas correlation refs, PIX QR fetch failures, best-effort handoff sends) — needs an explicit extra emit and a `FlowContext` seam into the toolpack context; tracked separately as a follow-up issue.

Fixes #40.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tool failures now return clear, helpful messages while preserving relevant details.
  * HTTP, authentication, network, credential, and provider errors are consistently identified as failures.
  * Invalid integration responses are handled safely, avoiding unnecessary follow-up actions.
  * Successful operations and business-level outcomes continue to work normally.
* **Monitoring**
  * Tool-call failures are recorded with appropriate warning or error severity.
  * Warning-level failures can trigger configured alert notifications.
* **Documentation**
  * Added guidance on tool-call logging and integration failure handling.

## Validation scope

No live Chatwoot validation in this round: no Chatwoot request changes — the only Chatwoot-facing surface touched is the reaction tool, whose request is unchanged (only its failure return is now marked). Unit + DB-backed e2e close the loop to the observable effect (an `alert_deliveries` row for a `minLevel: warn` channel).

### DIFF
--- a/docs/logs.md
+++ b/docs/logs.md
@@ -21,6 +21,10 @@ All three are in `TENANT_SCOPED_MODELS` with the standard `tenant_isolation` RLS
 
 `generate` (runtime + playground graph invoke), `stt` (`transcribeInboundAudio`), `vision` (`extractInboundFile` + the playground's `extractPlaygroundFile`; the two-step playground UI logs it under step 1's `extract` call), `tts` (`synthesizeReply`), `split` (`deliverReply`), `handoff` (assignee re-check in `runLoadedTurn`). **`embed` is in the vocabulary but not yet wired** — RAG search runs inside the `generate` span, so an embedding failure surfaces there; threading a flow context through the shared toolset builder is a deferred follow-up.
 
+### Tool lines and integration failures
+
+`ToolFlowLogger` (`src/graph/tool-flowlog.ts`) emits one `tool` line per tool call. A tool that **throws** is logged `warn`/`error` via `handleToolError`. A tool that degrades gracefully — returns a friendly string to the model instead of throwing (toolpacks, operator HTTP tools) — marks provider/credential failures with `toolFailure(...)` (`src/graph/tools/failure.ts`, built via `failableTool`): the model still sees the exact same string, but the line is logged `level: warn`, `status: error`, with the string as `errorMessage` — so alert channels with `minLevel: warn` fire on integration failures (expired OAuth, provider outage, rejected payloads). Business-level replies ("no free slots", policy limits, bad model input) stay `info`/`ok` — they are normal operation. For operator-authored HTTP tools every non-2xx counts as a failure.
+
 ## Alerting (`alerts.ts`, `alert-worker.ts`)
 
 - `dispatchAlertsForEvent` is called fire-and-forget from `emitFlowEvent` for `warn`/`error` events **on `source==="inbox"` only** (a playground error must not page). It matches channels by `minLevel` + stage allowlist and **coalesces**: a pending delivery for the same `(channel, stage, level)` is bumped (`count++`) instead of inserting a new row.
--- a/src/graph/tool-flowlog.ts
+++ b/src/graph/tool-flowlog.ts
@@ -25,7 +25,7 @@ function toolOutputValue(output: unknown): unknown {
   return output;
 }
 
-// A ToolMessage with status "error" is a tool-marked integration failure (failableTool/toolFailure —
+// NOTE: A ToolMessage with status "error" is a tool-marked integration failure (failableTool/toolFailure —
 // the friendly string went to the model, but the call must be logged as a failure). Thrown errors
 // take the handleToolError path instead; this only classifies returned outputs.
 function isErrorToolOutput(output: unknown): boolean {
@@ -82,7 +82,7 @@ export class ToolFlowLogger extends BaseCallbackHandler {
     this.starts.delete(runId);
     const failed = isErrorToolOutput(output);
     const value = toolOutputValue(output);
-    // Integration failure returned as a friendly string (failableTool): ONE line, level warn —
+    // NOTE: Integration failure returned as a friendly string (failableTool): ONE line, level warn —
     // same level as handleToolError, so alert channels (minLevel warn) can subscribe (issue #40).
     emitFlowEvent(this.flow, {
       stage: "tool",
--- a/src/graph/tool-flowlog.ts
+++ b/src/graph/tool-flowlog.ts
@@ -25,6 +25,18 @@ function toolOutputValue(output: unknown): unknown {
   return output;
 }
 
+// A ToolMessage with status "error" is a tool-marked integration failure (failableTool/toolFailure —
+// the friendly string went to the model, but the call must be logged as a failure). Thrown errors
+// take the handleToolError path instead; this only classifies returned outputs.
+function isErrorToolOutput(output: unknown): boolean {
+  return (
+    !!output &&
+    typeof output === "object" &&
+    "status" in output &&
+    (output as { status?: unknown }).status === "error"
+  );
+}
+
 // Logs each tool call the agent makes during a turn as a `tool` execution-flow line (name + status +
 // duration + the redacted args/result), so the operator can SEE which tools ran AND expand the marker
 // to inspect what was passed and returned (parity with the playground trace). Bound to the running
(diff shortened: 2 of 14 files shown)


gabrieljablonski:
@coderabbitai review

gabrieljablonski:
@coderabbitai review

gabrieljablonski:
@coderabbitai review

gabrieljablonski:
@coderabbitai review

gabrieljablonski:
@coderabbitai review

gabrieljablonski:
@coderabbitai review

PR #48, opened 2026-08-10
fix(flowlog): falha engolida em side effect de tool vira linha warn própria no log

### LINKED ISSUE #46: Swallowed side-effect failures are invisible to the flow log even when the tool line is ok
PR #44 made integration failures visible whenever the TOOL ITSELF fails: `failableTool`/`toolFailure` mark the `ToolMessage` with `status: "error"`, the flow log records a warn/error `tool` line, and alert channels fire.

There is a second family it deliberately did NOT cover: **failures swallowed inside a tool that still returns success**. The tool call genuinely succeeded from the model's point of view, so the tool line is legitimately `ok` — but a side effect failed and today that failure exists only as a server log line, invisible to the execution-flow log and to alert channels.

## Known sites

- **Orphan Asaas correlation refs** (`src/modules/integrations/toolpacks/asaas.ts:251-261` and `:456-466`): the payment link / PIX charge is created in Asaas, but persisting the `IntegrationExternalRef` fails. The tool returns the link (correct — the customer can pay), yet a future webhook can no longer be tied back to the conversation. Today: `logger.error` only, manual reconciliation.
- **PIX QR fetch failure** (`asaas.ts:472-483`): the charge exists, but the `pixQrCode` fetch fails and the tool returns the charge without the copy-and-paste code. Today: `logger.warn`.
- **Handoff best-effort targeting** (`src/graph/tools/native.ts:200-247`): the handoff itself succeeds (status → `open`), but the follow-up agent/team assignment fails. Correctly non-fatal, yet the operator never learns routing is broken. Today: swallowed.
- **`mirrorAttributeWrite`** (`native.ts:359-416`): the Chatwoot write succeeds, but the local DB mirror write-through fails (mirror catches up on the next webhook, so it self-heals — still worth a flowlog line if it repeats).

## Direction

Emit an explicit flow event from inside the tool, the way MCP discovery failures already do (`src/graph/prepare.ts` — `onDiscoverError` emits a warn `tool`-stage line with `emitFlowEvent`). That requires threading a `FlowContext` (or an `emit` callback) into `ToolpackCtx`/`ToolCtx`, which is its own seam — the reason this was split out of PR #44 rather than bolted on.

With that seam in place, each site above becomes one explicit `emitFlowEvent({ stage: "tool", level: "warn", ... })` carrying the side-effect failure, while the tool keeps returning success to the model.

Refs: #40, PR #44.

gabrieljablonski:
Fixes #46.

PR #44 made failures **of the tool itself** visible (ToolMessage `status: "error"` → warn `tool` line → alert channels). This PR covers the family it deliberately left out: **side effects that fail inside a tool that still returns success**. The tool line is legitimately `ok` — the model got what it asked for — but something the operator depends on silently broke, and until now that existed only as a stdout log line.

## The seam

`ToolpackCtx` and `ToolCtx` gain an optional callback:

```ts
onSideEffectError?: (e: { tool: string; phase: string; detail?: Record<string, unknown>; err: unknown }) => void;
```

`prepare.ts` binds it — exactly where the MCP `onDiscoverError` precedent already lives — to `emitFlowEvent(flow, { stage: "tool", level: "warn", status: "error", detail: { tool, phase, …detail }, errorMessage })`. `detail.tool` names the trail card (the MCP precedent leaves it unnamed); `detail.phase` discriminates the side effect. Absent a flow context the callback is undefined and every site degrades to today's log-only behavior. The tool's return value never changes.

## Sites covered

- **Orphan Asaas correlation refs** (`asaas.ts`, payment link + PIX charge): the charge/link exists but the `IntegrationExternalRef` write failed → a future webhook can't be tied back to the conversation. `phase: "persist_ref"`.
- **PIX copy-and-paste code missing** (`asaas.ts`): the `pixQrCode` fetch threw — and also the **non-2xx case, which never even threw** and used to vanish with no log at all. `phase: "pix_qr"`. (A 2xx without a payload stays quiet: that's a legitimate no-PIX-key state with a payable `invoiceUrl`.)
- **Handoff** (`native.ts`): assignment failure (`phase: "assign"`, with the targeting mode) and customer-message failure (`phase: "customer_message"`) — the operator never learned routing was broken.
- **Attribute mirror write-through** (`native.ts` `mirrorAttributeWrite`): `phase: "mirror_write"` with scope+key (self-heals on the next webhook, but repeats are worth seeing).
- **Kanban outbound emit** (`native.ts`): the card moved but the `kanban.card_moved` outbound event wasn't queued → external automation silently misses it. `phase: "outbound_emit"`.
- **Appointment reminders arm/cancel** (`prepare.ts` closures, where `flow` is already in scope): the Calendar event exists but reminders were never armed/cancelled — the customer silently misses them. `phase: "reminders_enqueue"` / `"reminders_cancel"` (reported under the `google_calendar` family name; the closure doesn't know which calendar tool called it).

`docs/logs.md` documents the new line.

## Deliberately out

- **`emitAck` failures** (the "I'll look into that…" nicety): paging warn channels for a missed ack would be noise without an action.
- **Playground**: its flow context is created after the toolset is built (and the graph can outlive a turn), so binding it there risks stale correlation — side effects stay stdout-only on the playground, where nothing pages anyway.

## Validation scope

- Unit (hermetic): every site above exercised with an `onSideEffectError` spy — the tool's return value is asserted to stay a success while the callback reports the right `tool`/`phase` (asaas persist_ref for link and charge, pix_qr non-2xx **and** thrown, handoff assign + customer_message, mirror_write, outbound_emit).
- End-to-end (DB-backed, local — the public CI has no Postgres service, so DB suites skip there): through the real `runAgentTurn` → `prepare` path, a handoff whose assignment fails lands (1) the handoff tool's own `info/ok` line, (2) a separate warn line with `detail.phase: "assign"` named `handoff_to_human`, and (3) an `alert_deliveries` row for a `minLevel: warn` channel — the issue's exact complaint closed end to end.
- No Chatwoot-facing request changes in this PR, so no live Chatwoot validation was run.

Refs #40, PR #44.

## Summary by CodeRabbit

- **Bug Fixes**
  - Tool side-effect failures are now recorded as separate warning events while the primary tool action remains successful.
  - Added error reporting for handoffs, customer messages, attribute updates, Kanban events, appointment reminders, and payment processing.
  - Payment and PIX operations preserve successful results even when follow-up persistence or QR-code requests fail.

- **Tests**
  - Added coverage validating side-effect warnings and preserved tool responses across supported integrations.


gabrieljablonski:
@coderabbitai review

gabrieljablonski:
@coderabbitai review

gabrieljablonski:
@coderabbitai full review

gabrieljablonski:
@coderabbitai full review

gabrieljablonski:
@coderabbitai full review

gabrieljablonski:
@coderabbitai full review

gabrieljablonski:
@coderabbitai full review

PR #51, opened 2026-08-10
feat(chatwoot): salvar nome informado pelo paciente

DrWallx:
## O que mudou

- adiciona a ferramenta nativa `set_contact_name`
- atualiza o nome do contato no Chatwoot com o token administrativo
- atualiza imediatamente o espelho local do contato
- normaliza espaços e rejeita valores semelhantes a telefone
- inclui a ferramenta no catálogo, na interface e nas traduções
- adiciona teste de regressão e documentação

## Por que

Quando um paciente informa o próprio nome, o agente precisa registrá-lo no contato para não perguntar novamente nas próximas conversas e para utilizar o primeiro nome no atendimento.

## Segurança

- somente o nome explicitamente informado é aceito
- não inventa sobrenome
- não sobrescreve um nome confiável sem correção explícita
- nenhuma credencial ou configuração de produção foi incluída
- alteração aditiva no repositório Free

## Validação

- `bun test tests/graph/tools.test.ts`: 37 aprovados, 0 falhas
- `bun check`: lint, TypeScript, worker, OpenAPI e i18n aprovados
- suíte geral: 1.307 aprovados; 3 testes de servidor estático expiraram apenas na execução conjunta no Windows e passaram isoladamente (3/3)


gabrieljablonski:
Thanks for this. The idea is one we want, and the mechanics you chose are the right ones (resolving the mirrored contact inside the tenant scope, updating Chatwoot through `updateContact`, then updating the mirror so the prompt variables follow on the next turn).

Before you invest more in it: we would like the scope generalized, and the version we would merge is meaningfully bigger than what is here. The whole direction is below so you can decide whether to take it. If you would rather not, say so and we will pick it up from here with you as co-author. Either answer is fine.

## The change: one tool for the contact's fields, not one tool per field

`set_contact_name` becomes `update_contact`, over a fixed field set, with the operator choosing per agent which of those fields the agent may write.

**In:** `name`, `email`, and the `additional_attributes` group Chatwoot's own contact form edits: `company_name`, `city`, `country`, `description`.

**Out, deliberately:**

- `phone_number`: unique per account and it is the contact's identity on the channel. It needs live validation before we hand it to a model.
- `identifier`: the WhatsApp to web-widget redirect stitches identities through it (`src/modules/channel-redirect/gate.ts`). A model rewriting it breaks that funnel.
- `avatar_url`: no use case.

One detail on `country`: Chatwoot's own form always writes `country` (display name) together with `country_code` (the ISO id), see `ContactForm.vue`. A tool that writes `country` alone leaves the code stale, so filters and automations keyed on `country_code` will not see it. Simplest acceptable behavior is to write `country` only, tell the model in the field description to use the country name as Chatwoot spells it, and document the caveat.

## 1. The operator's selection lives in `agent.settings`

Per agent, read through a typed reader in the shape of `readAttributeContextConfig` (`src/modules/chatwoot/attributes.ts`) or `readLimitsConfig` (`src/modules/agents/limits.ts`): anything malformed reads as "nothing selected", never as "everything". It needs the same plumbing every other behavior block has: `BEHAVIOR_SETTINGS_KEYS` and `mergeBehaviorSettings` (`src/modules/agents/behavior-settings.ts`), the `agent_settings_set` description in `src/modules/mcp/server.ts`, and the editor.

Two lists, with the invariant that a writable field must also be a visible one (letting the agent write a field it cannot see is blind overwriting):

```
contactFields: {
  context:  ["name", "email", "city"],  // current values injected into the prompt
  writable: ["name"]                    // subset the agent may change
}
```

## 2. The zod schema is built from that selection, not fixed

A field the operator did not enable must not exist in the schema the model sees, so it cannot even be attempted. There is precedent next to your code: `set_custom_attribute` only offers `scope: "task"` when the conversation actually has a linked kanban card (`src/graph/tools/native.ts`). All fields optional, at least one required. Empty selection means the tool is not built at all, fail-closed, same discipline as the native-tool allowlist.

## 3. One PUT, and a rejected write must be visible

`client.updateContact` needs to learn `additional_attributes`. No read-merge-write is needed for those: Chatwoot merges them server-side (`contacts_controller#contact_update_params`), so send only the keys you touch.

Wrap the tool with `failableTool` and return `toolFailure(...)` on a rejected write (`src/graph/tools/failure.ts`). Email is unique per account, so a 422 is a normal outcome, and an integration failure that returns a friendly string while the tool line stays `ok` is exactly what #40 and #46 were about.

## 4. Mirror write-through

Same as yours, extended to the fields you write, plus the `attributes` bag for the additional ones. One caveat to respect: Chatwoot does not deliver `contact_updated` to agent bots, so a late event carrying a pre-write snapshot can clobber what you just wrote. `mirrorAttributeWrite` in the same file solves that for the custom-attribute bag with a source watermark; the identity fields need the same care, or a note explaining why they do not.

## 5. The read side is half the feature

Today the model writes blind: it does not know the current name, and it cannot know the email is still missing. The current values of the selected fields go into the system prompt as a block, mirroring `buildAttributeContextSection` (`src/modules/chatwoot/attributes.ts`):

- appended AFTER `{{var}}` interpolation, so a stored value containing `{{nome_contato}}` reaches the model literally;
- values through `sanitizePromptValue` and XML-escaped, since they are customer-authored;
- framed as DATA, never as instruction;
- an empty field still emitted, flagged `filled="no"`. Knowing what is missing is what lets the agent ask for it at the right moment, which is the whole point.

The source is the mirrored contact row that turn prep already loads (`src/graph/prepare.ts`), so this costs no Chatwoot call.

## 6. Editor UI

It belongs in the "Data in context" section, which already has this shape. The field list is fixed, so no API call is needed to populate it: one row per field, with a checkbox for "the agent can update this one".

## 7. Tests

Schema gating (an unselected field is absent from the schema), the failure path (a rejected write produces a warn tool line, not a silent `ok`), the prompt block (values present, `filled="no"` on the empty ones), and the mirror write-through.

## 8. Docs

`docs/chatwoot.md`, replacing the section you added.

---

Most of what is already here survives: the normalization, the validation that rejects phone-shaped values, the i18n entries and the recording-client test setup all carry over. Let us know which way you want to go.

gabrieljablonski:
Taking this from here with you as co-author, which was the second half of the offer above. Nothing is needed from you; if you would rather have built the generalized version yourself, say so and it is yours.

Two reasons not to ask you to rebase first. `main` has moved 61 commits since this branch, and one of them (#147) removed `NATIVE_TOOL_RISK` outright, so the catalog entry here no longer has a map to land in. Rebasing would be work spent on a shape that is changing anyway.

What carries over from your work, unchanged: the whitespace normalization, the refusal of phone-shaped values, the requirement that a name contain a letter, the write order (Chatwoot first through `updateContact`, then the local mirror so the next turn's prompt variables follow), the i18n entries, and the recording-client test setup.

What changes is the scope described above: one `update_contact` over a fixed field set, with the operator choosing per agent which of those fields go into the prompt and which the agent may write, and the zod schema built from that selection so a field the operator did not enable cannot be attempted.

Closing so this does not sit as an open request on you. Thanks for #94 and #95 as well — both are in.

PR #53, opened 2026-08-10
feat(guardrails): validar respostas com o contexto do cliente

DrWallx:
## O que mudou

- envia a mensagem atual do cliente ao guardrail de saída
- torna explícita a verificação de instruções obrigatórias como `must`, `always`, `never`, `required` e `exactly`
- detecta respostas que omitem elementos exigidos, usam conteúdo proibido ou respondem a outra pergunta
- mantém a moderação de entrada sem duplicar esse contexto
- adiciona testes de regressão e documentação do runtime

## Por que

O verificador de aderência recebia a resposta do agente e o prompt configurado, mas não recebia a pergunta que estava sendo respondida. Sem esse contexto, não conseguia avaliar de forma confiável se a resposta era relevante, completa ou se cumpria uma regra condicional do atendimento.

## Impacto

A mudança é aditiva e só afeta agentes com guardrails de saída e `promptAdherence` habilitados. Agentes com o recurso desligado mantêm o comportamento atual.

## Validação

- teste específico de guardrails executado com sucesso em configuração backend isolada
- a suíte completa não pôde ser executada localmente porque o diretório de dependências estava incompleto e o ambiente isolado bloqueou o download dos pacotes
- a PR permanece em rascunho para a CI oficial validar lint, TypeScript e a suíte completa


gabrieljablonski:
Thank you for this, and sorry it sat for so long.

I read it closely and it turned into two PRs, both crediting you as co-author: **#94** and **#95**. Here is what I found, including the part of this PR that is not in its own description and that I think is the most valuable thing in it.

## The commit that is not in the description is the important one

`b867d835 fix: run OpenAI guardrails with the agent model fallback` fixes a real and serious defect, and the PR body does not mention it. I chased it end to end and it is worse than it looks:

1. `GUARDRAILS_DEFAULTS` ships `{ provider: "openai", model: "" }`.
2. The editor writes the per-provider default only in the provider select's `onChange`, and the provider already **is** `openai`, so an operator who enables guardrails without opening that dropdown saves `model: ""`.
3. The field is not blank while that happens: `ModelPicker` had `placeholder = "gpt-5.4-mini"` hardcoded, and the closed trigger renders the placeholder. The form shows a model that was never saved.
4. Measured on the version we ship: `new ChatOpenAI({ model: "" }).invocationParams({})` returns `{ model: "", … }`. It goes on the wire verbatim and the provider refuses the call.
5. `analyzeGuardrail` fails open, correctly, but returned the same `CLEAN` verdict as a real approval, and logged only to the process log. Nothing reached the Logs page or an alert channel.

So the guardrail reads as enabled, every verdict comes back clean, and there is nothing to notice. That is the **default** configuration. It deserved its own PR, which is #94.

**Two things moved in the fix.** `gr.model.trim() || loaded.mc.model` borrows the agent's model name, and the guardrails agent carries its own provider and credential precisely so it can be a different vendor: guardrails on `openai` with an agent on `gemini-3.5-flash` would send a Google model id to OpenAI and fail the same way with a different error. And the runtime is one call site, while `readGuardrailsConfig` is documented as the single source of defaults and clamping, so resolving it there covers every consumer. #94 resolves against one `PROVIDER_DEFAULT_MODEL` table now shared by the editor, the model picker and the runtime, adds the missing turn-trail line when an analysis fails, and fixes the placeholder that made the trap.

## The feature itself: right idea, and it moved to its own check

The adherence check cannot judge whether the reply answered the question, because it never sees the question. That is a real gap and your fix is the right shape.

What changed is where it lives. In #95 it is a separate check, `answer_relevance`, **off by default**, instead of widening `prompt_adherence`:

- **`prompt_adherence` ships on.** `GUARDRAILS_DEFAULTS.output.checks.promptAdherence` is `true`, so every install with output guardrails enabled already has it. Widening it changes behaviour for all of them without anyone opting in.
- **It is the only check that can trip on a correct reply.** A customer asks about prices, the agent asks which service, the customer says `sim`, and the agent's correct detailed answer is compared against `sim`. The action is not a log line: `template` / `generated` / `silent` means a false positive **replaces a good reply in front of the customer**. The prompt in #95 also tells the reviewer that answering more than was asked, or continuing an exchange already under way, is still an answer.

Two smaller things I would not have found without writing the tests:

- **`turnText` carries the conversation divider.** On the first turn of a returning contact it is prefixed with `CONVERSATION_DIVIDER` (`"(Contexto do sistema: início de uma nova conversa…)"`), so the reviewer would be told the customer said a system marker, on the opening turn of every returning attendance. #95 passes the raw inbound text. The `runGuardrail` parameter is renamed `subject` for a related reason: as `text` it shadowed the enclosing `text`, so the obvious spelling sends the **reply** as the question.
- **The review procedure and the customer message need the same condition.** Here the message block requires `customerMessage?.trim()` and the procedure block does not, so with an empty message the reviewer is told to "compare with the customer message" and there is none. An instruction to compare against something absent invites the model to supply it.

I also left `claims support that the supplied context does not establish` out. That is a grounding check, and `prompt_adherence` is documented as scope and persona; expanding what a checkbox means without renaming it changes what the operator agreed to.

## What the live testing found, which is the part I would want to know

Sending the customer's message to the reviewer is the right idea, and measuring it against a real model (`gpt-5.4-mini`) turned up three things that no amount of code review had caught. They apply to the original patch too, so they are worth passing on:

**1. The customer's message contaminates the other checks when it shares a call with them.** Same reply, same checks, only the message differing:

| | reply flagged `competitor_mention` |
| --- | --- |
| customer's message in the same call | **11 / 16** |
| customer's message absent | **0 / 16** |

The reply was `"Não. Nossa integração de WhatsApp é própria e nativa."`, which names nobody. The **customer** had named the competitor, and the configured action replaces that reply in front of the customer.

**2. Prompt wording could not fix it, and the most explicit wording was the worst.** Three arms, `n=16` each: no scoping sentence 6/16, a sentence scoping the message to one check 3/16, and a variant enumerating the policies that must ignore it (*"can never violate toxicity, competitor_mention…"*) **8/16**. Naming a policy to a model makes it consider that policy. So the fix is structural: `answer_relevance` gets its own model call, and the other policies keep exactly the call they had before. Measured again after the split: **0 / 16**.

**3. A relevance violation must not produce a generated replacement.** With the `generated` action, 10 replacements for one relevance violation: **3 stated a commercial fact the model had no way to know** ("yes, we work with them" / "no, we don't"), and **5 named a competitor from the operator's own list** while the same prompt told them never to. The reason is structural too: a relevance violation means the reply did not *answer*, so there is nothing to rewrite and the model has to invent. Toxicity rewrites what the agent said; relevance would be fabricating what the business does. That check now always falls back to the configured template.

There is one more, from the security side. Putting the customer's message anywhere the model reads as an operator instruction is a bypass: with it interpolated into the system prompt, a payload that closed the delimiter and ordered a clean verdict **got a genuinely abusive reply approved, 3 times out of 3**. In #95 the message travels as its own user message, fenced and named, with the fence delimiter stripped from the payload so it cannot close its own fence. Standard mitigation, not a guarantee.

## Where things stand

- **#94** (the model fix): merged.
- **#95** (the customer-message check): merged.
- **#99**: filed from what the measurements showed. The same fabrication path still exists for the input direction and for `prompt_adherence`, and both get the same battery before anything is changed.

Both carry `Co-authored-by: DrWallx`. Closing this one as superseded rather than leaving it open: everything it asked for is on `main` now, in a shape that came out of it. The defect it quietly fixed, the check it proposed, and three findings that only surfaced because it sent us looking. An open PR that will not be merged is worse than a closed one with the reason written down, so here is the reason. Reopen it or reply if you disagree with any of the calls above, the discussion is welcome and it is your finding either way.

One last note on this PR's CI: the only failure is formatting in `tests/modules/guardrails.test.ts`, which `bun format` fixes.

PR #54, opened 2026-08-10
feat(graph): bound the agent's per-contact memory (context ceiling + resolved-conversation retention)

rronaldovasconcelos-star:
## The problem

The graph checkpointer thread is keyed **per contact-inbox**, so it spans every conversation a contact ever had on that channel — and the agent node hands the model the **entire** history on every call:

```ts
const history = state.messages.filter((m) => m.getType() !== "system");
…
.invoke([new SystemMessage(prompt), ...history])
```

Nothing ever trimmed it and nothing ever dropped a finished attendance, so the context only grows. Measured on a production instance (a school's WhatsApp assistant, ~12 conversations with the same contact):

| | |
|---|---|
| context per generation | **79,862 tokens** |
| floor (system prompt + tool definitions) | 15,806 |
| **difference** | **~64k of already-closed conversations, re-sent every turn** |

A brand-new conversation that opens with "oi" already carries weeks of resolved attendances. That is paid twice:

- **Provider rate limit.** At ~80k tokens/turn, a 200k TPM organization fits **~2.5 turns per minute**. A normal burst throttles the live conversation — the symptom that sent me looking was "the assistant gets slower the more the family interacts", which had been (wrongly) written off as pure quota.
- **Cost.** When the provider's prompt cache expires, one turn goes from US$0.019 to **US$0.060**.

Emptying that one thread by hand dropped the next turn to **15,895 tokens (−80%)** and the turn latency from 11.2s to 2.9s, which is what confirmed the diagnosis.

The design intent — memory across a contact's conversations — is good, and this PR does not change it. What was missing is a bound.

## What this adds

Two knobs under `agent.settings.limits`, **both off by default** (an instance that upgrades must never silently start forgetting):

| knob | what it does | which half |
|---|---|---|
| `forgetResolvedAfterDays` | drops a contact's memory once **all** their conversations have been resolved for that long | the **cause** — cuts on a real semantic boundary (the attendance ended) rather than an arbitrary token offset |
| `maxHistoryTokens` | ceiling on the history sent to the model each turn | the **safety net** — one long-running conversation is never resolved, so retention never sees it |

They are meant to be used together.

### `maxHistoryTokens` — `trimMessages`, and three options that are load-bearing

- **`startOn: "human"`** — a window must never open on a `ToolMessage` whose originating `tool_call` was just dropped. Providers reject an orphan tool result outright (OpenAI 400), so a naive "keep the last N tokens" would fail **hardest on exactly the longest conversations**. The test for this is calibrated so a naive window lands on the `ToolMessage`; without `startOn` it fails.
- **`allowPartial: false`** — half a message reads as the agent misquoting the customer.
- **`includeSystem: false`** — the prompt is prepended per turn by the caller and is never part of `history`; counting it would silently shrink the budget.

A trim emits a `generate` warn (`historyTrimmed` / `historyKept`) so it shows up in the turn trail — trimming that leaves no trace is indistinguishable, from the operator's chair, from the agent forgetting things. If trimming throws, it falls back to the full history: it is an optimization and must never cost a customer their answer.

### `forgetResolvedAfterDays` — `MEMORY_SWEEP`

One self-rearming job per tenant, same shape as `FLOWLOG_SWEEP`. It deletes the checkpointer thread + `AgentThread` row when **both** hold:

1. every conversation on the thread has been resolved for longer than the window, **and**
2. the thread has **no `PENDING`/`CLAIMED` scheduler job**.

**(2) is the guard that matters.** The follow-up ladder runs on this same thread on purpose — `docs/graph.md` records a past bug where the nudge ran on the per-conversation thread and lost that memory — and `pauseWhileAppointment` can push a follow-up well past any day count. So the day count is slack; "nothing scheduled" is the guarantee. Forgetting early would recreate that bug on a timer, and **silently**: the follow-up still fires, just generic. Hence the guidance (in the UI hint and the docs) to keep the window comfortably above the longest configured follow-up delay.

Other decisions worth surfacing for review:

- Two agents on one instance with different windows: **the longest wins**. Forgetting early is the destructive direction, so the tie-break goes to the more conservative agent.
- Checkpointer deleted **before** the marker row: if the second delete fails, the next sweep finds the row again — the reverse order could leave a large orphan thread nobody looks for.
- Per-tick cap (200 threads) so a backlog drains over days instead of holding one long transaction against tables every live turn also writes to.
- `getCheckpointer` imported lazily, only once there is something to forget (it opens a dedicated Postgres pool).
- Armed for every tenant regardless of config — the handler checks whether any agent opted in, so enabling the knob needs no restart.

## Surfaces

Both knobs flow through `readLimitsConfig`, so REST / MCP / UI project the same validated shape: editor Behavior tab (with the operational advice in the hints), `agent_settings_set`, and `docs/graph.md`. `0` and negatives read as **off**, not as "the tightest possible setting".

## Tests

New tests, each verified to **fail against the defect it pins** (including the orphan-tool one, which passes either way if the budget is not calibrated — I checked, and fixed it):

- `tests/graph/agent-node.test.ts` — no ceiling = previous behavior; trims oldest and keeps the current turn; never opens on an orphan tool result; falls back to the full history when trimming throws.
- `tests/modules/agent-limits.test.ts` — defaults off, rounding, clamping, `0`/negative = off.
- `tests/modules/agent-memory-retention.test.ts` — forgets long-resolved threads, **spares recent ones and any with a pending job**, and an agent without the knob forgets nothing.

`bun test` is green (1428 pass / 0 fail), as are the Biome check, both `tsc --noEmit` passes, the OpenAPI check and `i18n:extract` (pt-BR strings translated).

⚠️ **One caveat on my side:** the two DB-backed suites follow the repo's `describe.skipIf(!dbUp)` pattern and I had no Postgres available on this machine, so `agent-memory-retention.test.ts` **was not executed locally** — it type-checks and lints, but CI is the first place it actually runs. Worth a close look in review.

Happy to split this into two PRs (ceiling / retention) if you'd prefer them reviewed separately.


gabrieljablonski:
Thanks for this, and for the measurements above all. The diagnosis is right and it is the reason the design moved: I opened #55 to work through the shape of the fix.

A few things in here you got right that I want on the record, because they are the parts that are easy to get wrong and hard to notice afterwards:

- `startOn: "human"`. A window opening on a `ToolMessage` whose `tool_call` was dropped is rejected outright by the provider, and it would fail hardest on exactly the longest conversations, which are the target. Calibrating the test budget so a naive window lands on the orphan is the right way to pin it.
- The reasoning about the follow-up ladder running on this same thread, and `pauseWhileAppointment` pushing a follow-up past any day count. "No pending job" as the real guarantee and the day count as slack is the correct reading.
- Finding that `APPOINTMENT_REMINDER` links to its conversation only through the payload, since its dedupe key is `reminder:<googleEventId>:<offsetHours>`. Matching the key alone would have left visit reminders unprotected.
- Deleting the checkpointer before the marker row, so a failure leaves a row the next sweep finds again instead of an orphan thread nobody looks for.

## Where the design landed

The short version of #55: instead of deleting a contact's memory once their attendances have been closed for N days, **compact** it. When an attendance closes, its raw turns are replaced by a summary of that attendance, so the thread becomes "N summarized attendances + the current one, raw". Same trigger you chose (the attendance ended, a real semantic boundary), condensing instead of discarding.

Two consequences worth flagging to you specifically:

1. The risk you designed most carefully around goes away in that form. Nothing is deleted, so a follow-up firing months later still finds a summary instead of waking up generic. The `MEMORY_SWEEP` job kind, the enum migration, the per-tenant arming and the sweep query all go with it.
2. Compaction covers something the ceiling cannot. `maxHistoryTokens` bounds what is **sent**; the `messages` channel keeps concatenating and every turn rewrites the whole channel into a new checkpoint. Only `RemoveMessage` bounds what is **stored**, and it is already available on the `@langchain/langgraph` and `@langchain/core` versions we ship.

This is not a rejection of the approach in this PR. It is your diagnosis taken one step further, and the whole first half of your work survives unchanged.

## What I would like to do with this PR

Reduce it to the **`maxHistoryTokens` half**, and merge that. The ceiling is not superseded by compaction: a conversation that never gets resolved is never seen by the attendance boundary, which is the hole you identified yourself, so the ceiling stays as the safety net. Everything in the trim (the three `trimMessages` options, the fallback to the full history when trimming throws, the reader clamps, the editor field, the MCP description) lands as-is.

One change on the way in: the trim currently emits a `generate` **warn**. After #44 and #48, a `warn` with `source: inbox` is dispatched to alert channels, and once a ceiling is configured, trimming is normal operation on every turn of any long-standing contact, not an anomaly. Coalescing dampens the burst but not the stream, so a correctly configured ceiling would become a permanent source of alerts. `info` keeps it on the Logs page and in the turn trail without paging anyone. Your point that a silent trim is indistinguishable from the agent forgetting things is right; it just belongs one level down.

If you would rather not split it, say so and we will take the ceiling from here with you as co-author.

## One question only you can answer quickly

**Were your measurements taken with a hot or a cold prompt cache?**

It matters more than it looks. Prefix caching requires an exact prefix match, so the append-only history we have today is maximally cacheable, and a cached token costs a fraction of a fresh one. But cache TTL is on the order of 5 to 10 minutes, and WhatsApp turns are often further apart than that, which would mean the cache is cold most of the time and the full 79.8k is paid at full price on nearly every turn. Your 11.2s to 2.9s latency figure points that way.

If you can confirm it from the provider dashboard (cached input tokens against total input tokens on that instance over a day), it settles how much of the win is size-shaped and how much is cache-shaped, and it goes straight into #55.

Design questions 2 through 5 in #55 (how many attendances stay raw, which model summarizes, where the summary lives, what the lossiness guardrails are) are open and your input is welcome there.

gabrieljablonski:
Closing this, because it shipped — in two pieces, both crediting you.

- **The ceiling half is #127** (merged 19 Aug). It keeps your diagnosis, your measurements, and the three invariants this PR got right: the trim starting on a human message, whole messages only, and the system prompt outside the count. `Co-authored-by` on the commit.
- **The compaction half is #134** (merged 20 Aug), the shape #55 settled on. Credited there too, on the measurement.

Your answer in #55 on 14 Aug is what settled it. Splitting the cache rate by how long it had been since the previous call on the same thread, instead of averaging across a day, is the thing that made the TTL boundary visible — and that is the reason the ceiling ships off by default while compaction ships on.

`forgetResolvedAfterDays` and the `MEMORY_SWEEP` job kind stayed out, per the design discussion in #55: nothing is deleted any more, so the risk you designed most carefully around does not arise in that form.

One thing I got wrong here, and it is worth naming: my last word on this PR was a proposal to reduce it to the ceiling half and merge it, and then the ceiling landed in a PR of mine while this one stayed open for ten days. The credit is in the commits, but the PR should have been closed with this explanation when #127 merged, not now.

PR #70, opened 2026-08-14
feat: hide WhatsApp group conversations from web

chaostheory53:
## Summary

- classify and persist WhatsApp group conversations from Chatwoot `@g.us` contact metadata
- add an opt-in database/REST filter while preserving the complete default used by MCP
- make only the web conversation list exclude groups, including pagination and realtime reclassification

## Testing

- `bun check` (1246 passed, 573 DB-dependent skipped, 0 failed)
- focused conversation, Chatwoot, client, and realtime suites


chaostheory53:
Opened against the wrong repository. This change will be submitted to the private homelab repository instead.

Issue #27, opened 2026-08-06
Follow-up sweep never fires for bot-owned conversations (assignee_type = 'AgentBot')


rogerheldcorp-ia:
## Description

The follow-up sweep's eligibility SQL requires `assignee_type IS NULL`, but Chatwoot assigns bot-owned conversations `assignee_type = 'AgentBot'` (not `NULL`) as the normal state — the same state `shouldBotHandle` explicitly treats as "still ours, safe to act" (see `docs/chatwoot.md`, the `shouldBotHandle` predicate: `assignee_type != "User"`, with `AgentBot` only excluded when it belongs to a *different* bot).

As a result, `agent.settings.followUp` never enqueues a `FOLLOWUP` job for a conversation that is actively being handled by the bot alone — which is the normal case, not an edge case. Follow-up effectively never fires in ordinary operation.

## Steps to reproduce

1. Enable `followUp` on an agent bound to a WhatsApp (Baileys) inbox, e.g. `{ enabled: true, steps: [{ delayValue: 2, delayUnit: "minutes", instructions: "..." }, { delayValue: 5, delayUnit: "minutes", resolve: true, instructions: "..." }] }`.
2. Have a real customer message the inbox and let the agent reply once (creating a normal bot-owned conversation: `status = 'pending'`, `assignee_type = 'AgentBot'`).
3. Stop replying and wait past the configured first-step delay.
4. Check `scheduler_jobs` for a `FOLLOWUP` job tied to that conversation's thread — none is created. The `FOLLOWUP_SWEEP` job keeps re-arming every 60s but never enqueues anything for this conversation.

Confirmed on a live instance: verified `conversations.assignee_type = 'AgentBot'` for the affected conversation, and confirmed via `scheduler_jobs` that no `FOLLOWUP` row was ever created for its thread, despite well over an hour of inactivity in some cases.

## Expected

The sweep should treat `assignee_type = 'AgentBot'` as eligible (bot-owned, no human yet), consistent with `shouldBotHandle`'s own definition — not require a literal `NULL`.

## Where

`src/modules/followups/handlers.ts`, `sweepHandler`'s raw SQL:

```sql
AND c.assignee_type IS NULL
```

Likely fix: mirror `shouldBotHandle`'s condition, e.g. `AND c.assignee_type IS DISTINCT FROM 'User'` (or explicitly allow `NULL` or `'AgentBot'`).

## Environment

- Self-hosted, Free edition
[...]

PR #75, opened 2026-08-14
fix(chatwoot): order conversation state by the payload's updated_at so a handoff survives its own message tail

### LINKED ISSUE #61: Conversation mirror reverts to pre-handoff state when the stale message_updated tail shares a second with the status change
A handoff leaves the conversation mirror showing `pending` / unassigned, while Chatwoot itself correctly shows `open` / assigned to the human. Whether it happens is decided by **whether the agent's outgoing message and the status change land in the same wall-clock second** — same code path, same event sequence, different outcome.

Two conversations on the same inbox, twenty minutes apart, both handed off by `handoff_to_human` in `pinned` mode:

| | Chatwoot | mirror (`conversations` row) |
|---|---|---|
| conv 6 | `open` · assignee = user 3 | **`pending` · assignee `null`** |
| conv 9 | `open` · assignee = user 3 | `open` · assignee = user 3 |

The mirror row for conv 6 was written at `21:26:59`, five seconds **after** the handoff — with the pre-handoff values. Not a missed update: a stale write that won.

### Mechanism

Chatwoot delivers a burst of events after a handoff. The tail of that burst is `message_updated`, and its embedded conversation snapshot reflects the state **at message creation**, not at dispatch. The handoff tool posts the customer message *before* it changes status (`src/graph/tools/native.ts`), so that snapshot is always the pre-handoff one.

Both conversations received the identical sequence, ending the same way:

```
EVENTO "conversation_updated"        status: "open"
EVENTO "conversation_opened"         status: "open"
EVENTO "conversation_status_changed" status: "open"
EVENTO "conversation_updated"        status: "open"     ← assignee set here
EVENTO "message_updated"             status: "pending"  ← stale
EVENTO "message_updated"             status: "pending"  ← stale
```

So the stale tail is not the anomaly — it is normal. What differs is whether the monotonic guard catches it. `src/modules/chatwoot/mirror.ts:119`:

```ts
// Monotonic guard: only when both timestamps are known. A null new timestamp applies
// best-effort (the next timestamped event reconciles).
const stale =
  newLastEventAt != null &&
  existing?.lastEventAt != null &&
  existing.lastEventAt > newLastEventAt;
```

`newLastEventAt` comes from the payload's `last_activity_at`, which has **one-second resolution** and **does not advance** when only status or assignee changes. The actual timestamps carried by those bursts:

```
conv 6   every event, stale tail included ..... 1786483614
                                                            → equal, `>` is false, stale write APPLIES

conv 9   correct events ......................  1786484801
         stale message_updated tail ..........  1786484800
                                                            → 801 > 800, guard fires, stale write SKIPPED
```

conv 9 was saved by one second of luck.

### Why the write is destructive

In the update branch (`mirror.ts:189`), `status` is guarded against a missing value but the assignee fields are not:

```ts
...(n.status != null ? { status: n.status } : {}),
assigneeId: n.assigneeId,
assigneeType: n.assigneeType,
assigneeName: n.assigneeName,
```

so an event carrying `assignee: nil` clears a real human assignee unconditionally.

### Impact

The attribution gate is unaffected — `shouldBotHandle` runs on the normalized webhook event, not on the mirror, so the bot correctly stays silent on a conversation it has handed off. What breaks is everything that reads the mirror:

- the console / `list_conversations` show a handed-off conversation as still owned by the bot;
- `conversation_reengage` resolves against the mirror;
- `emitMirrorEvent("conversation.handoff")` fires on the `assigneeType !== "User" → "User"` edge, and the stale write silently reverses that edge — so the next legitimate handoff re-fires it, and any alerting built on "conversation has no human owner" produces false positives.

### Suggested direction

Bumping the comparison to `>=` does **not** fix it: every event in these bursts shares one timestamp, so `>=` would discard the correct ones too (in conv 6 the first event of the burst is the `pending` one, and the mirror would end up equally wrong).

The guard cannot order a same-second burst. What can is not treating message events as authoritative for conversation-level state: the conversation snapshot embedded in `message_created` / `message_updated` describes the message's moment, so `status` / `assigneeId` / `assigneeType` would be better mirrored only from conversation-level events (`conversation_updated`, `conversation_status_changed`, `conversation_opened`, `conversation_resolved`) and from creation. Message events would still update `lastEventAt`, `lastInboundAt` and contact data.

Failing that, a tiebreaker with finer resolution than `last_activity_at` — an event sequence number, or the delivery id — would let the guard order the burst.

### Environment

`harbor.fazer.ai/agents/fazer-ai/agents-pro:latest` (package version 1.1.1), Chatwoot Pro `v4.16.2-fazer-ai-pro.112`, `Channel::Email` inbox, agent handoff `mode: "pinned"` with `targetAgentId` set. Observed 2026-08-11 on a single-agent tenant; the conversations differ only in wall-clock timing.

gabrieljablonski:
Fixes #61.

The report is right on every point, including the part that says bumping the comparison to `>=` would not fix it, and its second suggestion is the one this takes: a tiebreaker with finer resolution than `last_activity_at`. That tiebreaker already ships in the payload.

## What was wrong

A message event's conversation snapshot is serialized when the message event fires (`AgentBotListener` builds `message.webhook_data` and only then enqueues the delivery; a failed delivery retries with that same copy), so it describes the conversation as of that moment, not the delivery's. `handoff_to_human` posts the customer message *before* it changes the status, so the tail of every handoff burst carries the pre-handoff snapshot.

There was already a fence for that, but it was written against one symptom rather than the cause:

```ts
const staleMessageReopen =
  n.message !== undefined &&
  !isNewIncomingMessage(n) &&
  (existing.status === "resolved" || existing.status === "snoozed") &&   // <- only these two
  (n.status === "pending" || n.status === "open");
```

It was built for a message delivered after a `conversation_resolved` dragging the mirror back to `pending`. A handoff leaves the conversation `open`, so the fence never fired, and because `assigneeKnown` was derived from the same flag, the stale tail also cleared a real human assignee.

Whether it corrupted the row was decided by the clock, exactly as reported: `last_activity_at` has one-second resolution and does not advance on a status or assignee change, so when the whole burst shares a second the monotonic guard cannot order it.

## The fix

**A message event's embedded `conversation` is a snapshot, not a statement.** Conversation state comes from conversation-level events, ordered by the conversation's own version. Two rules follow from that, and they are the whole change:

**1. Each field carries the version of the payload that wrote it.** Not one watermark, two: `chatwoot_status_at` and `chatwoot_assignee_at`.

```ts
// A mark moves when the field it belongs to is WRITTEN, and only forward.
statusAt: status != null ? advances(row.statusAt) : null,
assigneeAt: assignee ? advances(row.assigneeAt) : null,
```

One mark cannot serve both, because a degraded payload (`meta` absent — the intermittent shape behind #27) carries a trustworthy status and says nothing at all about the assignee. After it lands the two fields legitimately reflect different versions of the source row, and whichever single number were kept, something would be ordered by a value that does not describe it: hold the degraded version and the complete event delivered after it loses the assignee it is the only witness of; withhold it and that same event reopens a conversation resolved after it. Three review rounds found that defect from three different sides before the marks were split.

**2. A snapshot moves no state and claims no version.** The second half is what makes the first safe: if a snapshot moved a mark it would push it past a conversation event still in flight, and that event would arrive already "stale" — precisely how a delayed handoff used to be lost.

`updated_at` is the source row's version: it moves on every write to the conversation, including the status and assignee changes `last_activity_at` ignores; it carries a fraction (`push_timestamps` sends `updated_at.to_f`, the only `to_f` among its `to_i` siblings); and it is serialized together with the state it describes. Upstream since Chatwoot 4.0.2 (`chatwoot/chatwoot#10873`, `#10875`). Stored as the raw double, not a timestamp: two writes land microseconds apart and a `TIMESTAMP(3)` would round them into one version.

The comparison is `>=`, not `>` — Chatwoot emits several events for ONE write, and the one carrying `meta` is frequently the second, so rejecting an equal version would drop the assignee it brought. At an equal version an assignee may be SET but never CLEARED: every payload is serialized from one conversation object, so companions of a single write agree by construction, and a disagreement means one witness is degraded, with `null` being the degraded reading.

### The one thing a message does write

A brand-new incoming message reopens the conversation before the event is dispatched (`Message#execute_after_create_commit_callbacks` runs `reopen_conversation`, then `dispatch_create_events`). Status only, never the assignee — and because it WRITES the status, it also claims the status version, which is safe only because the marks are split: it moves the status mark and leaves the assignee ordered by a mark that snapshot never touched.

It claims that version unconditionally, not only when the status changed. The mirror frequently has not SEEN the change: when the resolve is itself delayed, the row still reads `open` as the reopen lands, and withholding the version on that basis leaves the delayed resolve looking newer than the mark. A guard on "changed" was carried for two rounds and removed once the only test justifying it turned out to describe a burst that cannot occur: the snapshot is read from the row at dispatch (`set_conversation_activity` runs first), so a newer message always saw the newer state, and a message serialized before a conversation event carries a lower version and cannot push the mark past it.

### Why the old code trusted those snapshots, and why that argument dissolves

When the handoff's own event is delayed past the first message the human sends, that message's snapshot is the only witness of the new assignee. Distrusting it looked like it would leave the mirror bot-owned for good — because under the old guard the delayed event LOST on arrival.

That "for good" was an artifact of the guard the argument was written against, not a fact about the problem. Ordering removes the trap: the snapshot claims no version, so the watermark does not advance past the delayed event, and a conversation event carrying a version is judged by that version alone — never by `last_activity_at`, which is exactly the value that does not move on a handoff. The delayed event applies when it lands, which is what the witness argument wanted all along. There is a test pinning it.

### Which axis orders what

The guard picks its axis from what the event offers:

- a conversation event **with** a version → that version, and nothing else. When the row has no version yet, it applies: that is every row the migration touched, and falling back to `last_activity_at` there would recreate the discard above for exactly the conversations that are live at the upgrade;
- everything else → `last_activity_at`. A message, which that value describes exactly, and a conversation event from a Chatwoot too old to send a version, where there is nothing finer to order by and the monotonic guard is the only protection against a status regression.

**Everything a delayed event does NOT speak about keeps a recency fence.** The attribute bags (`customAttributes`, `kanbanAttributes`) and the relations (`contactInboxId`, `inboxId`, `contactId`) are carried by every payload and versioned by none, so they keep the fence the stale check used to give them for free: a delayed conversation event rules on the state it names and says nothing about the rest. Otherwise it rolls a Kanban card back a column, or restores a contact-inbox from before a merge — and the graph's thread key is built from that one, so the agent's next turn lands on another conversation's history.

An earlier revision also fenced CLOSES on that axis (a delayed `resolved` may not close a conversation a customer just wrote into). That is gone: once the reopen claims its status version, the VERSION rejects such a resolve on the fraction, where the fence had to round to the whole second. Removing it deleted a `closesConversation` special case and one more place to get the axis wrong; both cases stay pinned by tests.

**The live probe follows the same rule**, because it is the same kind of payload. `runAgentNudge` re-checks the conversation over REST before any proactive send, and the REST `show` renders the same `updated_at.to_f` the webhook does — so the reconcile writes each field only when the version carrying it is at least as new as that field's mark, advances both marks from it, and writes nothing at all when nothing differs.

### Where there is no version at all

With none on either side a message snapshot cannot be ordered against the conversation, so it moves no state. Trusting it to *promote* an assignee was tried and withdrawn: it fixes a lost handoff and creates its mirror image, since a message the human sent before the conversation went back to the bot arrives with the same `last_activity_at` and re-takes it, firing a false `conversation.handoff`. Neither direction can be ordered, so the mirror keeps what it has. Recorded in `.codex-review-waived`.

### What this does NOT close

A conversation event already in flight can still undo a state write made from our own console. It predates this change (`updateMirror` never moved `lastEventAt` either, so the previous guard did not reject that event) and closing it needs the version our own REST call produced, which neither endpoint returns: `assignments#create` renders the agent, `toggle_status` renders `{success, conversation_id, current_status, snoozed_until}`. Filed as #77, with the fix that is worth having — return `updated_at` from both endpoints in the fork and let the local write claim the watermark.

A local-marker mechanism that tried to cover it without that version was built in this PR and then removed: without the real version it needs a fence, an expiry, a lock and a way to tell the echo of our own write from any other conversation event, and review found four separate races inside it. The snapshot rule then made it unnecessary for the case it was actually protecting — a console take-over undone by a message tail, which no longer has a voice.

## Where the rules live

The ordering is a pure function, `src/modules/chatwoot/state-order.ts`: it takes the payload's ordering facts and the row's marks, and returns what may be written. No DB, no clock. `mirrorChatwootEvent` collects the facts, asks once, and writes what it is told.

That split is not cosmetic. The mirror was deciding five groups of fields on three different axes inside the webhook's hot path, and the explanation had grown longer than the code it explained (124 comment lines around 71 of logic). The reasoning is now the header of the unit it describes, and the unit can be tested as what it is: a decision table.

## Blast radius

`prisma/migrations/20260814191926_conversation_chatwoot_state_versions` adds two nullable `DOUBLE PRECISION` columns. Existing rows start with no version; the first conversation event that lands on them establishes one and ordering runs from there, and until then a versioned conversation event still applies rather than being discarded.

## Validation scope

**Proved by test**, `tests/modules/chatwoot-mirror-handoff.test.ts`, DB-backed, replaying real payload shapes through `normalizeChatwootEvent` + `mirrorChatwootEvent` (not hand-built normalized events). 29 cases, the ones worth naming:

- the burst entirely inside one second (the conversation that broke, `1786483614`) → red before the change with `status: "pending"`, the reported symptom, and the human assignee cleared;
- the burst whose tail is one second behind (the conversation that survived on luck) → still discarded, so it keeps working for a reason rather than by accident;
- a message that overtakes the handoff event: the handoff lands when it arrives and `conversation.handoff` fires once;
- the same delayed handoff on a row with no watermark yet — the migrated-row shape, i.e. every conversation live at the upgrade;
- a degraded payload (no `meta`) still resolves the conversation, and its status is not walked back by an older complete event delivered after it — while that same complete event still owns the assignee it is the only witness of. Both hold only because the marks are split;
- a resolve that predates the customer's message cannot close the conversation, and neither can the resolve's companion arriving after a reopen — including when both share one whole second, where nothing but the version can separate them;
- a message that changes nothing does not outrank a conversation event still in flight (removing that guard left the suite green until this case was written);
- a delayed conversation event does not roll back the attribute bags, and reports the timestamp the row actually kept, since the webhook broadcasts it and the console sorts on it;
- with no version anywhere, a late message can neither re-take a conversation nor take it back;
- two writes inside the same millisecond stay ordered (red when the version is rounded to milliseconds, which is what a `TIMESTAMP(3)` would do);
- a take-over through the real `handoffConversation` and a resolve through the real `setConversationStatus` both survive a message tail serialized before the click, with no marker of any kind, because a tail no longer speaks;
- a new incoming message still reopens a conversation resolved in the console, and a resolved conversation reopened by a customer does not regress;
- a delayed event does not restore the relations a newer message already mirrored, nor close a conversation on the strength of the agent's own reply;
- the live reconcile records the version of a probe that merely CONFIRMS the mirror (which is what a migrated row needs), and a probe older than the marks does not overwrite newer state.

**Proved by table**, `tests/modules/chatwoot-state-order.test.ts`: 27 cases against the pure function, one rule each, no database. Every rule was mutation-checked rather than only red-proved: inverting the comparison, dropping the condition, collapsing the two marks into one, and removing the recency fence each take at least one case red. Reverting a fix proves a test reaches the code; killing the rule proves the test is about the rule.

Each guard in the DB-backed suite was also red-proved by reverting it and confirming the specific case fails.

The existing suites that exercise the mirror (`chatwoot-receiver`, `followup-resolved-guardrails`, `attribute-context`) pass unchanged, including their own degraded-payload pins.

**Not validated live**: reproducing this needs a real handoff burst whose events happen to share a wall-clock second, which is the coin flip the issue is about. The replayed burst uses the timestamps recorded in the report, and the `updated_at` semantics were read off the Chatwoot source (`Conversations::EventDataPresenter#push_timestamps`, `AgentBotListener`, `Message#execute_after_create_commit_callbacks`).

`bun check` green (2094 tests).


gabrieljablonski:
Finding accepted, and it points past the fix rather than at a hole in it.

The scenario is real: with message snapshots distrusted outright, a handoff whose own event is delayed past the first message the human sends leaves the mirror bot-owned for good. The message advances `lastEventAt`, the delayed conversation event then loses to the monotonic guard, and nothing else ever carries the new assignee. `conversation.handoff` never fires either. Measured against the previous commit: `status: "pending"`, `assigneeId: null`, zero `conversation.handoff` deliveries.

The suggested remedy is the second half of the sentence, "or state needs a separate ordering mechanism", and Chatwoot already ships one. `Conversations::EventDataPresenter#push_timestamps` sends the conversation row's `updated_at` as `to_f`: sub-second, moving on exactly the writes `last_activity_at` ignores, serialized together with the state it describes, present in every payload shape including the message events. Conversation state is now ordered by it (`Conversation.chatwootUpdatedAt`), which drops the distrust rule to a fallback for a Chatwoot older than 4.0.2.

Ordering also removes the hand-written carve-out the distrust rule needed for the reopen case, since Chatwoot reopens before it dispatches the message event. Both scenarios are pinned by tests.

gabrieljablonski:
Round 3, both findings.

**Human-message handoffs without a version stamp** — accepted and fixed. On a Chatwoot too old to send `updated_at` the fence had no version to compare, so it fell back to distrusting message snapshots, and that loses the takeover exactly as described. The rule now is narrower: without a version, a message snapshot may still PROMOTE the conversation to a human (`assigneeType` → `User`), because a frozen handoff tail carries the pre-handoff state and can never invent a human assignee, while the first message that human sends is the only witness we get. It never clears an assignee, never moves the status, and does not apply where versions exist — there an older event may genuinely be reporting an assignment that was already undone. Two tests: promotion carried, and the frozen tail still fenced out on the same legacy path.

**Rolling deploys** — real, and waived with the reasoning recorded in `.codex-review-waived`. It needs an operator running multiple web replicas (the documented model is one, `docs/deploy.md` "Single replica (or one leader)"), a rolling deploy, and a webhook delivered late inside the overlap — which Chatwoot bounds to roughly nine seconds (`AgentBots::WebhookJob` retries 3× at 3s). The result is a stale status/assignee the next conversation event repairs, which is the same self-healing any out-of-order delivery already relies on. A drain or two-phase rollout would put deploy ceremony into every future release to close a seconds-wide window.

gabrieljablonski:
Round 4, both findings, and the first one retires the previous round's fix.

**Do not promote from unordered message snapshots** — correct, and it is the mirror image of the round-3 finding rather than a variant of it. A message the human sent *before* the conversation went back to the bot arrives with the same `last_activity_at` and re-takes it, firing a false `conversation.handoff` at every subscriber. My justification ("an unordered message cannot invent a human assignee") only held for the pre-handoff tail, exactly as you say.

So the promotion is withdrawn. Without a version, neither direction can be ordered, and between losing a handoff and inventing one, the mirror keeps what it has: a missed update self-heals on the next conversation event, a false handoff fires downstream automation that nobody can un-fire. The residual — on a Chatwoot older than 4.0.2, a handoff whose events are ALL delayed past the human's first message stays bot-owned until the next conversation event — is recorded in `.codex-review-waived` with that reasoning. Supported deployments ship our fork (4.11.x), where the version exists and decides.

**Do not advance the state watermark from partial payloads** — accepted and fixed, and the fix is one step further than requested. A payload with no `meta` now neither advances the watermark nor applies state at all: a payload that cannot describe the assignee is not a description of the conversation. Advancing was indeed the worse half (the degraded payload outranked a complete one arriving late with the assignment), but applying its `status` while fencing its assignee was never coherent either. Pinned by a test where the degraded payload carries the NEWER version and the real assignment, delivered after it, still wins.

Both red before green: with the promotion restored, the re-take test fails; with the watermark advancing on the degraded payload, the assignment is dropped.

PR #76, opened 2026-08-14
feat(tools): send an image to the conversation from its URL

### LINKED ISSUE #65: Sending an image to a conversation requires the Google Drive integration
## Summary

`ChatwootClient.sendFileAttachment` is the only way to deliver a file to a conversation, and its
only caller in the codebase is `drive_send_file`:

```
src/modules/chatwoot/client.ts:235       async sendFileAttachment(...)   ← definition
src/modules/integrations/toolpacks/google-drive.ts:303                    ← the only call
```

An agent that already knows an image's URL has no way to send it. The delivery path exists and
works; the only entry point into it is a Google Drive download.

## Why it matters

The URL is usually already in hand. An HTTP tool definition, an MCP tool or a catalog integration
returns a product image URL, the agent reads it, and then has to answer with a link to the product
page instead of the picture. For a commerce agent that is the common case, not an edge case —
customers ask to see the item.

Working around it means either uploading the shop's catalogue into Google Drive (where it drifts
from the shop the moment a product changes) or giving up on images.

## What would close it

A way to turn an https URL into an attachment, using the machinery that is already there:
`assertSafeOutboundUrl` for the fetch and `sendFileAttachment` for the delivery.

The part worth designing carefully is which hosts an agent may fetch from. Binding the allowed
origins to the integration instance config — never to a tool argument — keeps it an operator
control, so a prompt injection cannot redirect the fetch. That mirrors how `folderId` is handled in
the Drive toolpack today.

## Happy to contribute

We implemented this locally as a small toolpack against 1.4.1 and it is running in production:

- new file, plus one `import` line in `toolpacks/index.ts` and one `catalog.ts` entry;
- allowed origins from the instance config, SSRF-guarded, https only, no redirects, bounded
  timeout, byte cap enforced against the body rather than a claimed `content-length`, and a
  content type that must really be an image;
- delivery requires the live conversation handle, so the playground degrades to a clear message
  instead of pretending the image was sent;
- 13 tests covering the origin fence, the caps and the failure paths.

If the shape is acceptable we would be glad to open a PR — or to adjust it to whatever design you
prefer.

## Environment

- agents 1.4.1

gabrieljablonski:
Closes #65.

The report is right about the gap and about the part worth designing carefully. `sendFileAttachment` had exactly one caller, so "send the customer a picture" meant "put your catalogue in Google Drive first", and a commerce agent holding a product image URL had to answer with a link instead.

## The tool

`send_image`, a **native** tool rather than a toolpack. A toolpack is the activation of a *catalog integration*: it carries a credential, a fixed origin and the inbound correlation ref. Sending an image to the current conversation has none of those — it is the same kind of act as `private_note` or `react_to_message`, over the same bot token, and the native `ToolCtx` already carries the live client and conversation id. Making it an integration would mean asking the operator to create an instance of a service that does not exist, just to enable a tool.

That also puts the config where every other per-agent tool behavior already lives (`agent.settings`), so it projects over REST, MCP and the editor with no new plumbing, and the playground keeps simulating it like every other conversation tool.

## The fence

The URL is model-supplied, so `agent.settings.sendImage.allowedHosts` decides where it may point — operator-set, never a tool argument. Empty is the default and refuses everything: granting the tool cannot silently turn the agent into a fetcher for arbitrary URLs, which would also be an exfiltration channel (the model writes the URL, and a query string carries whatever it encodes).

Around that:

- the allowlist is checked **before** the SSRF assertion, so an unlisted host is not even resolved;
- `assertSafeOutboundUrl` then keeps an allowed name from pointing at a private address;
- `redirect: "error"` — a redirect is a second URL nobody allowlisted;
- the byte cap is counted against the bytes that **arrive**, not a claimed `content-length` (a chunked response carries none anyway);
- the type comes from the file's own signature, not from a header the same server wrote: PNG, JPEG, GIF, WebP. "It says `image/png`" and "it is a PNG" are different claims;
- the customer-visible file name is the URL's basename stripped to safe characters, with the extension from the **detected** type.

A wildcard entry (`*.loja.com.br`) covers the apex and its subdomains, and the `.` in the suffix check is what stops `evil-loja.com.br` from matching.

Every refusal returns a model-facing message that says what to do instead ("responda com o link em texto"), so a blocked image degrades into a useful answer rather than an apology loop. Only the transport failures (host unreachable, HTTP error, Chatwoot rejecting the upload) are marked as integration failures, so the Logs page and alert channels fire on those and not on an operator's deliberate allowlist.

The hosts ride in the tool description as an XML block, the same way the kanban card and the known attributes do, so the model knows where it may point the tool before it burns a turn guessing.

## Delivery is deferred, like the resolve intent

The tool does not post. It **queues** into the turn's state and the runtime delivers after the same gates the reply passes: the ownership recheck, the supersede gate and the output guardrail. A tool that posts from inside the graph invocation can put a picture in front of a customer whose turn is then discarded, which is exactly why `resolve_conversation` is already deferred.

Consequences that are easy to get wrong, and are pinned by tests:

- captions are **model-written text the customer reads**, so they are screened with the reply — including when the model wrote no reply at all (`send_image` then `skip_reply`), which would otherwise be a way around output moderation. A trip drops the images along with the words;
- an image that reached the customer **is an answer**, so an image-only turn reports `posted`; reporting `empty` would leave a stale turn error on a conversation that was just answered;
- a proactive nudge has no turn to queue into, and its own 24h service-window gate decides what may be sent, so the tool declines — **before** the download, not after;
- one model response can carry a whole batch of tool calls, and LangGraph's `ToolNode` runs them with `Promise.all`, so the per-turn ceiling reserves its slot **before** the fetch. A ceiling checked across the download is read by every call in the batch while the queue is still empty.

## Validation scope

**Proved by test**, `tests/modules/send-image.test.ts`, 21 tests, hermetic (a fake host that serves the body as a **stream**, so the cap is tested against a server that lies in `content-length`; a stub Chatwoot client that records what was delivered):

- the host list: what an operator actually pastes (a full URL, a port, mixed case) reduces to a hostname; a wildcard covers the apex and subdomains and not a look-alike domain; junk is dropped rather than stored;
- the fetch: an allowed host delivers with the type read from the file; an unlisted host is refused **without resolving DNS**; no configured host refuses everything; an allowed host on a private address is still refused; a body past the cap is cut off mid-stream; a page claiming `image/png` is rejected; a GIF served as `application/octet-stream` goes through as a GIF;
- the tool: the image is queued with the right bytes, name, mime and caption; a URL outside the list sends nothing and tells the agent what to do; a Chatwoot failure degrades instead of losing the turn; the description carries the configured hosts; the per-turn ceiling holds against twelve concurrent calls; the byte budget counts the image it is deciding on; the ticket preserves the model's order when the downloads finish out of it; a caption past 500 chars is refused before anything is queued; and a resolver that never answers is cut off by the same deadline as the download, which it used to run before.

Plus the delivery half, DB-backed in `tests/graph/runtime.test.ts`: the image lands before the reply in the same turn; a turn taken over mid-flight delivers nothing at all; a guardrail trip drops the queued image along with the reply; a caption is screened even when the model wrote no reply; an image with no final text still counts as an answered turn.

Nothing the model wrote reaches the customer without passing a gate. That took four passes, because the same value kept finding a different door: the tool's arguments, the string it returns, the refusal message, and the attachment's own file name — which `fileNameFor` used to take from the model-supplied URL path, so an image from an allowed host could deliver `marca-proibida.png` past an output guardrail configured to catch exactly that. The name is now ours, and its extension comes from the detected signature, so it cannot lie about the file either.

And the log half, DB-backed in `tests/graph/tool-flowlog.test.ts`, driving the real graph and reading the persisted row: a URL argument is stored as its origin and nothing past it (a presigned signature, an identifying path, embedded credentials all gone); a string that announces itself as http(s) and then fails to parse is replaced rather than passed through; a caption never reaches storage at all; a URL is replaced outright rather than trimmed to something still readable, because every part past the scheme is model-written — the host included, since a wildcard allowlist hands the model the subdomain; and a string that announces itself as http(s) and then fails to parse is replaced too, since it is unparseable precisely because host and payload cannot be told apart. `ExecutionLog.detail` is documented to hold allowlisted ids/counts/enums and to be exportable, and both of the argument shapes this change introduces would have landed there verbatim.

Red before green on the parts where a regression would be dangerous: with the host check disabled, the two fence tests fail; with the signature check replaced by the declared `content-type`, the "claims to be a PNG" test fails.

**Not validated live, deliberately.** What is new here is the fetch and its fence, and both are fully exercised offline against a fake host — a live run would exercise the same code against one real CDN and prove less. The delivery half is `ChatwootClient.sendFileAttachment` unchanged, which the Drive toolpack has been running in production since it shipped.

`bun check` green (2085 tests).

## What this does not close

Output guardrails read text and never the picture itself. Submitting bytes to them means a multimodal guardrail — the guardrails agent is text-only by construction — and suppressing images whenever they cannot be screened would disable the tool precisely for the operator careful enough to turn guardrails on. An image has its own mandatory, default-closed control instead: `sendImage.allowedHosts` is an operator decision, empty by default, and empty refuses every call. What was missing is that the operator could not know the split, so the hint on that field now says it. Adjudicated in `.codex-review-waived`.

Arbitrary tool arguments reaching `ExecutionLog.detail` is a general problem that predates this change — an HTTP tool whose schema takes a document number writes it there today — and the fix is a policy decision that applies to every tool at once. Filed as #78.

## Note on the offered PR

@YosefAdPro — thank you for the detailed write-up, and for saying what your implementation does rather than just that it exists. The fence here ended up the same shape as yours (origins from config, https, no redirects, bounded timeout, cap against the body rather than the header, a content type that has to really be an image); what differs is where it hangs, native rather than a toolpack, for the reasons above. If you have cases your production traffic hit that these tests do not cover, that is exactly what would be worth a PR on top.


gabrieljablonski:
Accepted, and it is the right shape rather than a hardening detail: the tool was posting to the customer from inside the graph invocation, so a turn later found superseded, taken over or blocked could already have delivered an image, and a caption never reached the output guardrail.

It now queues into `TurnState` and the runtime delivers after the gates, exactly like the deferred resolve intent that exists for the same reason. Concretely:

- the queue is drained after the ownership recheck, the supersede gate and the output guardrail, and dropped on every outcome that discards the reply;
- the image lands before the reply text and before the TTS branch, so an audio reply does not swallow the attachment;
- the captions ride into the guardrail's screening together with the reply, and a trip drops the queue: replacing the words while the picture goes out would moderate half the message;
- a proactive nudge has no turn to queue into, and its own gate (`proactiveSendMode`, the 24h window) decides what may be sent there, so the tool declines instead of posting behind that gate. Attaching an image to a proactive send is a separate change, not a silent one.

Both halves are pinned end to end in `tests/graph/runtime.test.ts`, DB-backed, with the download and the SSRF assertion injected so nothing resolves a real host:

- a queued image is delivered before the reply in the same turn (`sendFileAttachment` then `sendMessage`);
- a turn taken over mid-flight delivers **nothing** — red when the tool sends inline, which is the finding reproduced;
- an output-guardrail trip drops the queued image along with the reply — red without the drop.

PR #91, opened 2026-08-16
ci(docs): publish the API reference from the release tag

gabrieljablonski:
## What was wrong

The published API reference (Swagger UI on GitHub Pages) shows `0.0.0` as the API version, and it documents `main` rather than any released version. Both come from the same place.

`info.version` is wired to `config.packageInfo.version` (`src/api/index.ts`), which reads `package.json`. This repo keeps `package.json` at `0.0.0` by design: the real number is injected from the release tag at image-build time (`npm version "$TAG_VERSION" --no-git-tag-version`, `publish_github_package.yml`) and is never committed. `scripts/build-openapi.ts` boots the app to generate the committed `openapi.json`, so the placeholder is baked into the file, and this workflow copied that file verbatim. Measured on the committed spec:

```
$ jq -r .info.version openapi.json
0.0.0
```

The version was only the visible half. The workflow triggered on every push to `main` that touched `openapi.json`, so the published reference described endpoints nobody could call yet: the newest image an operator can run is the last release, so the page ran permanently ahead of every deployed instance.

Correcting `info.version` in the committed file is not an option: `bun openapi:check` compares it byte for byte against the freshly generated document (wired into `bun check` and CI), so a real version stored there is permanent drift.

## What changed

Publish from the release tag instead of from `main`, and stamp the version at publish time, which is what `publish_github_package.yml` already does to `package.json` when it builds the image.

- **Trigger**: `on: push` becomes `on: release: types: [released]`, the same event `publish_github_package.yml` uses to build the image this documents. `released` fires when a release is published as stable and when a prerelease is promoted in place, and never for a prerelease on its own, so the reference and the image always move together. (`published` would have missed the promotion: GitHub emits no second `published` event for it.)
- **Checkout**: the job resolves the tag and checks that ref out, so the spec published is the released spec.
- **Stamp**: `cp openapi.json ./public/openapi.json` becomes `jq '.info.version = $v'`, with the leading `v` stripped from the tag.
- **Manual repair**: `workflow_dispatch` stays and takes an optional `tag` input. Without it, the job falls back to `repos/<repo>/releases/latest`, documented as the most recent non-prerelease, non-draft release. The input exists because that endpoint sorts by the *tagged commit's* date, not by publication time, so a release cut from an older commit is not the one it returns.

`src/api/index.ts` is untouched. The `0.0.0` in the committed spec stays a placeholder, consistent with `package.json`, and the `openapi:check` drift guard is unaffected.

## Validation scope

Proven locally by extracting the `run:` blocks from the workflow itself (never retyped) and executing them against a fake `gh` and the real committed `openapi.json`:

- the trigger is `release: [released]`, and it is the same trigger `publish_github_package.yml` declares;
- release payload `v1.6.0` resolves `tag=v1.6.0`, `version=1.6.0`;
- a dispatch naming `v1.4.1` resolves `1.4.1`, winning over the latest-release lookup;
- a dispatch naming nothing falls back to `repos/<repo>/releases/latest` and resolves `1.6.0`;
- the stamped artifact carries `info.version: 1.6.0` while the committed file still reads `0.0.0`;
- the published document differs from the committed one in `info.version` and in nothing else;
- red: the previous `cp` path reproduces `0.0.0` on the artifact.

Every check was mutation-tested, one mutation at a time, and each mutation fails the checks it should: reverting the trigger to `published`, re-adding a prerelease guard, dropping the `v` strip, removing the dispatch fallback, pointing the fallback at `releases` instead of `releases/latest`, ignoring `inputs.tag`, making the input required, restoring `cp`, and making `jq` overwrite the rest of `info`.

The stamp was additionally run against this repo's own `openapi.json` (155 paths): `info.version` changes, the rest of the document is byte-identical. `jq 1.7` and `GitHub CLI 2.97.0` are both in the `ubuntu-24.04` runner image (checked against `actions/runner-images`).

**Not validated:** the workflow has not run on GitHub. The trigger change, the `ref:` checkout and the dispatch input are only exercised by a real `release: released` event or a manual dispatch, so the first proof in production is the next release cut.

**Review:** three codex rounds (`gpt-5.6-sol`, effort `high`), converging to `0 findings` on the current head. The reviewer bot's token was unreadable throughout, so the rounds are recorded as comments on this PR rather than as reviews. One finding is waived rather than fixed (gating the docs deploy on the image build), with the reasoning in `.codex-review-waived`.

### DIFF
--- a/.codex-review-waived
+++ b/.codex-review-waived
@@ -103,3 +103,17 @@ src/modules/conversations/service.ts: Persist abandonment after the scheduler co
 # our own checkpointer. That is a mechanism justified by a hypothesis; if a measurable quality gap
 # shows up on multi-step tool loops, it earns its own change.
 src/graph/models.ts: Include encrypted reasoning in stateless tool loops
+
+# PR #91. The window is real and can be long (the image job is a multi-arch matrix and its runner
+# queue has held a job for hours), but the conclusion does not follow. At the instant `released`
+# fires, the version is already public: the GitHub Release exists, with its notes and its source
+# tarball. The reference tracks the RELEASE, which is the announcement of the version, not the
+# container artifact, so a GHCR image lagging its own release is a gap between the image and the
+# release, and putting the docs inside that gap does not close it. The remedy costs real mechanism:
+# `workflow_run` runs in the default branch's context and carries none of the release payload (the
+# tag would have to be re-resolved from the triggering run), `publish_github_package.yml` also has a
+# `workflow_dispatch` trigger so the docs would fire on builds that are not releases, and it adds a
+# failure mode that does not exist today (image published, docs silently not). A release whose image
+# never builds is a broken release: repair the build or cut another tag. `workflow_dispatch` already
+# covers repairing the page by hand.
+.github/workflows/deploy-swagger.yml: Wait for image publication before deploying the docs
--- a/.github/workflows/deploy-swagger.yml
+++ b/.github/workflows/deploy-swagger.yml
@@ -5,17 +5,21 @@ name: Deploy Swagger UI to GitHub Pages
 # tooling drops THIS WORKFLOW from the Pro tree (openapi.json itself ships in both), and the job is
 # additionally guarded to the public Free repo so it no-ops if it ever reaches the master or a fork.
 #
+# It publishes from the RELEASE TAG, not from main. Publishing on every push to main documented
+# endpoints nobody could call yet: the newest image an operator can run is the last release, so the
+# page ran ahead of every deployed instance. Publishing from the tag also makes the version real:
+# the committed spec carries the 0.0.0 placeholder (info.version reads package.json, which this repo
+# keeps at 0.0.0 by design), so the number is stamped here from the tag, exactly like
+# publish_github_package.yml does to package.json when it builds the image.
+#
 # NOTE: Pages must be enabled on the repo with "GitHub Actions" as the source, or `configure-pages`
 # fails with "Get Pages site failed". It cannot self-enable: the action's `enablement` input needs a
 # token other than GITHUB_TOKEN. Pages on a private repo also requires a paid plan, which is why
 # this job only ever succeeded after fazer-ai/agents went public.
 on:
-  push:
-    branches:
-      - main
-    paths:
-      - "openapi.json"
-      - ".github/workflows/deploy-swagger.yml"
+  release:
+    types: [published]
+  # Republishes the latest release: use it to repair the page without cutting a release.
   workflow_dispatch:
 
 # Serialize Pages deploys; never cancel an in-flight publish.
@@ -26,7 +30,9 @@ concurrency:
 jobs:
   deploy:
     runs-on: ubuntu-latest
-    if: github.repository == 'fazer-ai/agents'
+    # A prerelease (rc/beta) is not what operators run, so it must not become the public docs.
+    # On workflow_dispatch there is no release payload and the expression reads null, i.e. falsy.
+    if: github.repository == 'fazer-ai/agents' && !github.event.release.prerelease
 
     permissions:
       contents: read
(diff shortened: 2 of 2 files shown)


gabrieljablonski:
### Review round 1 (codex `gpt-5.6-sol`, effort `high`)

Posted as a plain comment, not as a review: the reviewer bot's token was unreadable at publish time (1Password mount down), and publishing under the author's account would attribute a verdict to a human who did not write it.

**1 finding, P2, confidence 0.90 — Handle promotions from prerelease to stable**

> When an RC/beta is promoted in place to a stable release, GitHub emits `release: released`, not another `published` event. The original `published` event was skipped while `prerelease` was true, so the documentation remains on the previous tag even though `.github/workflows/publish_github_package.yml` publishes the promoted image on `released`; subscribe to `released` as well.

Accepted, and it points at a bigger mistake than the one it describes: `publish_github_package.yml` triggers on `release: types: [released]`, so this workflow should simply use the same event as the image it documents, rather than `published` plus a hand-rolled prerelease guard.

`released` fires when a release is published as stable **and** when a prerelease is promoted, and never for a prerelease itself. So switching to it fixes the promotion gap, makes the explicit `!prerelease` guard redundant, and ties the docs to the image by construction: the same event that ships the image publishes its reference. Fix incoming.

gabrieljablonski:
### Review round 2 (codex `gpt-5.6-sol`, effort `high`)

Again posted as a plain comment: the reviewer bot's token is still unreadable.

**2 findings, both P2.**

---

**1. Wait for image publication before deploying the docs** (conf 0.96)

> When the package workflow is delayed or fails, this release-triggered job still deploys independently; the much faster Pages job can therefore advertise the new version while `ghcr.io/fazer-ai/agents:latest` still points to the previous release, permanently so after a failed build. Trigger the deployment from a successful `publish_github_package.yml` completion or verify the released image before updating Pages.

**Rejected**, with the mechanism accepted. The window is real and can be long (the image job is a multi-arch matrix and its runner queue has held a job for hours), but the conclusion does not follow.

At the instant `released` fires, the version is already public: the GitHub Release exists, with its notes and its source tarball. The reference tracks *the release*, which is the announcement of the version, not the container artifact. A GHCR image lagging its own release is a gap between the image and the release, and inserting the docs into the middle of it does not close it.

The remedy costs real mechanism for that: `workflow_run` executes in the context of the default branch and carries none of the release payload, so the tag would have to be re-resolved from the triggering run; `publish_github_package.yml` also has a `workflow_dispatch` trigger, so the docs would fire on image builds that are not releases; and it adds a failure mode that does not exist today (image published, docs silently not). The permanent case the finding describes, a release whose image never builds, is a broken release: the fix there is to repair the build or cut another tag, not to hold the reference hostage. `workflow_dispatch` already covers repairing the page by hand.

---

**2. Resolve manual repairs by publication order** (conf 0.94)

> When a prerelease is promoted after a newer maintenance release, or an older draft is published, `releases/latest` may return the wrong stable release because GitHub selects it by `created_at` (the tagged commit's date), not by the most recent `released` event. A subsequent repair dispatch can therefore roll the documentation back to an older tag; resolve by publication time or accept the intended tag as a dispatch input.

**Accepted.** The premise is documented in the same paragraph I used to justify `releases/latest`: "sorted by the `created_at` attribute", where `created_at` "is the date of the commit used for the release, and not the date when the release was drafted or published". So a release cut from an older commit is not "latest", and a repair dispatch could quietly move the page backwards.

This only affects the manual path (`released` carries an exact tag), and the second remedy is the cheap one: `workflow_dispatch` now takes an optional `tag` input. Whoever repairs the page names the release, and the `releases/latest` lookup stays as the default for the common case. Fix incoming.

gabrieljablonski:
### Review round 3 (codex `gpt-5.6-sol`, effort `high`) — clean

`0 findings`, verdict `patch is correct`, on `54ac17ca9fa8f6876a5232da35deeb99eabf49ee` (the current head).

One finding stands waived rather than fixed, with its reasoning in `.codex-review-waived`: gating the docs deploy on a successful image build. Summary of the rebuttal: the reference tracks the GitHub Release, which is what announces the version, not the container artifact, and `workflow_run` would add a failure mode (image published, docs silently not) that does not exist today.

Still posted as a plain comment: the reviewer bot's token was unreadable for all three rounds, so none of them could be published as a review. No review state is stuck as a result (nothing ever posted a `CHANGES_REQUESTED`), but this PR carries no bot review either. The clean round's `findings.json` is saved locally and can be published once the mount is back.

PR #147, opened 2026-08-20
Remove the per-tool risk tier, which no surface ever read

### LINKED ISSUE #137: Nothing reads the per-tool risk tier, but it still ships in the REST API, the MCP tools and the agent export
Every tool in the system carries a risk tier of `low | medium | high`, and nothing in the codebase ever reads it.

## What exists today

Three sources declare it:

- Built-in native and RAG tools, in `NATIVE_TOOL_RISK` / `RAG_TOOL_RISK` (`src/graph/tools/catalog.ts`).
- Toolpack tools, declared per pack (`ToolRisk` in `src/modules/integrations/toolpacks/types.ts`, projected as `ToolView.riskTier`).
- Custom HTTP tools, persisted in the `ToolDefinition.riskTier` column, defaulting to `medium` for a tool of unknown impact.

No code branches on the value. There is no gate, no approval step, no filter, no ordering, no warning, and no log line that consults it. The tier is inert metadata that is validated on the way in, stored, projected and handed back out.

The UI that once gave it a purpose is gone too: the `RiskBadge` component no longer exists under `src/client/components/`, so the tier is not rendered anywhere. The playground endpoint still projects `risk` on every tool (`PlaygroundToolInfo`), the console receives it, and the tool row draws only the category badge.

## Why remove it instead of leaving it there

A field named "risk" reads like a control. It is the first thing someone reaches for when the goal is to contain what an agent is allowed to do, and anything built on it inherits a guarantee that was never enforced. Keeping a dormant safety-shaped field is worse than not having one, because it invites exactly the wrong assumption at exactly the wrong moment.

It also is not free. The value is validated, defaulted, persisted, projected and round-tripped through three public surfaces, so every change to the tool shape pays a small tax for a field with no reader.

## Where it lives

REST:
- `src/api/v1/tools.controller.ts:107` accepts `riskTier` on the tool body.
- `src/modules/tool-definitions/service.ts` validates it (`:130`), defaults it on create (`:228`), patches it (`:309`) and selects/projects it (`:56`, `:101`).

MCP:
- `src/modules/mcp/server.ts:1362`, `:1421` accept `risk_tier` on the tool write schemas, `:1389` / `:1449` in the arg types, and `:582` advertises it in the list-tools description.
- `src/modules/mcp/write-agents.ts:454`, `:479` map it onto the patch.

Agent export/import:
- `src/modules/agents/transfer.ts:99` pins it in the schema, `:588` writes it out, `:1107` reads it back.

Read projections:
- `src/modules/agents/service.ts` carries it through the tool-listing projections.
- `src/modules/playground/service.ts` projects `risk` per tool for the console.

Persistence:
- `ToolDefinition.riskTier`, `prisma/schema.prisma`.

## Scope

The code surfaces above, in one change: stop accepting the field on input, stop emitting it on output, and delete the declarations behind it.

The `tool_definitions.risk_tier` **column is not part of this issue** — it comes out one release later, in #149. The two are separable because they fail differently: removing the code is reversible under an image rollback, dropping the column is not (the previous build still selects it). Splitting them keeps each one closable by a single pull request.

## What this is not

If a per-tool containment mechanism is wanted later, it should be designed as one, with an enforcement point and tests, rather than by reviving this field.

gabrieljablonski:
Every tool in the system declared a `low | medium | high` risk tier, and nothing in the codebase ever read it. No gate, no approval step, no filter, no ordering, no log line consulted the value. The `RiskBadge` that once rendered it in the tool-selection UI is gone, so the tier survived purely as metadata: validated on the way in, stored, projected, and handed back out through REST, MCP and the agent export format.

A dormant field named "risk" is worse than not having one. It is the first thing someone reaches for when the goal is to bound what an agent is allowed to do, and anything built on it would inherit a guarantee that was never enforced.

## Why this is one removal and not a deprecation window

The issue assumed a staged rollout, because dropping the field looks like a breaking change to three public surfaces and to export bundles produced by older instances. Measured rather than assumed, every input surface **discards** an undeclared key instead of refusing it:

| surface | old client still sends the field |
|---|---|
| REST `POST/PATCH /v1/tools` | `200`; Elysia's `normalize` strips it before the handler |
| MCP `tool_create` / `tool_update` | accepted; the SDK's zod object strips it from the args |
| agent import | imports; `exportedHttpToolSchema` is a plain `z.object`, which strips |

The Elysia behavior is the same one already documented in `tests/api/v1/tools-controller.test.ts` (a field missing from the route body schema is silently dropped — that is how `label` once regressed). It also explains why the service's `.strict()` create schema never sees the retired key: the controller strips it one layer earlier.

So the only observable change is on the **output** side. Responses, projections and newly written export bundles stop carrying the field. Nothing can act on the value, so nothing can break on its absence beyond a consumer reading it for display.

## What changed

- `RiskTier`, `NATIVE_TOOL_RISK` and `RAG_TOOL_RISK` leave `src/graph/tools/catalog.ts`. `ToolRisk` and `ToolSpec.risk` leave the toolpack contract, so a pack now declares a name and a schema.
- REST drops `riskTier` from the tool write body and from every read projection (tool list, agent tool-selection catalog, playground tool list).
- MCP drops `risk_tier` from the two tool write schemas and stops advertising it in the `tool_list` description.
- The agent import stops writing it.

## What deliberately did not change

Two things outlive this release on purpose, and #149 removes both.

**The `tool_definitions.risk_tier` column stays.** The two halves fail differently. Removing the code is reversible: an operator who rolls the image back gets a build that reads a column which is still there. Dropping the column is not: the previous build selects it in the tool-definition read, the agent tool-selection catalog and the export, so a rollback would answer `undefined_column` on every tool screen with no way out but rolling forward. The same asymmetry applies to a deploy that overlaps the two processes, which `docker-compose.coolify.yml` allows by binding no host port while `prisma migrate deploy` runs from the new container's start command.

Customer message handling is unaffected in either direction — the turn loader (`src/graph/prepare.ts`) never selected the column — so the exposure is the operator console, not the agent. Waiting one release costs a dead column with its default intact; since nothing reads the value, dropping it sooner buys nothing.

**The agent export still echoes the stored value.** The compatibility above only covers old bundle into new importer. The reverse matters just as much: the bundle format is versioned as a whole (`version: z.literal(1)`), so an instance one release behind parses our bundle with a schema where `riskTier` is *required*. Dropping the key from the export would make every bundle this build writes unimportable there, and bumping the version would trade that for a cleaner refusal while also making this build reject every v1 bundle. So the key is emitted while the column lives, and optional on the way in, in both directions.

## Validation scope

**Proven by test.** Three tests pin the compatibility this removal rests on, and all three are new:

- `tests/api/v1/tools-controller.test.ts` drives a real request through the route's own exported `writeBody` and asserts a body carrying `riskTier` returns `200` with the field stripped. Red against the previous schema: restoring `riskTier` to `writeBody` makes it fail with one extra key received.
- `tests/modules/agent-transfer.test.ts` imports a bundle whose HTTP tool still carries `riskTier` and asserts the tool row is created. This one is a forward guard rather than a red-first test (the field used to be part of the schema, so it passed before too). Its mutation is `.strict()` on `exportedHttpToolSchema`: applied, exactly one test fails, and it is this one.
- `tests/modules/agent-transfer.test.ts` also exports a bundle and checks every HTTP tool against the older release's required-field shape. Red without the export echo: removing that one line fails exactly this test.

Full suite green on both trees: 2932 tests on the source tree, 2891 on the derived one. The derived tree reports one unrelated failure locally, `tests/client/pages/KnowledgeApprovals.test.tsx`, a 5s timeout that reproduces on pristine `main` in the same checkout and passes in CI — environmental, not from this change.

**Not validated live.** Nothing here touches Chatwoot, a provider, or any outbound request, so there was no live check to run. The MCP strip is asserted from the SDK's documented zod behavior and reproduced against an equivalent schema, not driven through a real MCP client session.

Fixes #137


gabrieljablonski:
### Review record (3 rounds, local `gpt-5.6-sol` at effort `high`)

The reviewer bot's token comes from a 1Password mount that is not serving on this machine right now, so none of these rounds could be posted as a review object under the bot account. Posting them under a human account would attribute a verdict to a person who did not write it, so the findings are recorded here instead, verbatim. No round was published, so there is no pending `CHANGES_REQUESTED` state on this PR — it simply carries no bot review. The three `findings.json` files are parked and will be published if the mount comes back.

**Round 1** — `overall_correctness: patch is incorrect`, 1 finding.

> **[P1] Stage the column drop after the code-only release** (confidence 0.96)
> When this image is rolled out while the previous build is still serving, or if a rollback is needed, this migration runs before the new server and removes a column the previous build explicitly selects and writes; `/v1/tools`, agent tool selection, and exports can then fail with PostgreSQL `undefined_column`, and rolling back the application does not restore them. The deploy contract explicitly orders migration before serving and calls out maintenance for destructive migrations (`docs/deploy.md:26-40`), so defer this `DROP COLUMN` until all old processes are gone or enforce a maintenance rollout.

Adopted. Verified first: the turn loader (`src/graph/prepare.ts`) never selected the column, so the exposure is the operator console rather than message handling; there is no prior `DROP COLUMN` in the migration history to lean on either way; and `docker-compose.coolify.yml` binds no host port, so the two processes can overlap on the platform this project treats as first-class. The migration moved to #149 and the column now carries a NOTE saying why it survives. #137 was rescoped to the code surfaces so a single PR can close it.

**Round 2** — `overall_correctness: patch is incorrect`, 1 finding.

> **[P2] Keep `riskTier` in version-1 exports** (confidence 0.97)
> When a component bundle containing an HTTP tool is exported by this release and imported into a deployment still on the previous release—such as during a staggered upgrade or rollback—this map now omits `riskTier` while `AGENT_EXPORT_VERSION` remains 1. The previous version's `exportedHttpToolSchema` requires that field, so it rejects the entire bundle as invalid; accepting old bundles in the new importer only covers the reverse direction. Continue emitting the retained column for v1 exports, or version the incompatible format change.

Adopted, taking the first option. Bumping the version is strictly worse here: `version` is a `z.literal(1)`, so raising it would trade a confusing refusal on the old instance for a clean one **and** make this build reject every existing v1 bundle. The export echoes the stored value while the column lives, the key is optional on the way in in both directions, and a test checks each exported tool against the previous release's required-field shape (red without the echo).

**Round 3** — `overall_correctness: patch is correct`, 0 findings, against `0d66753`, worktree clean.

Issue #144, opened 2026-08-20
The agent is never told whether it is open, so it recites business hours from the prompt instead of the schedule


gabrieljablonski:
Of the whole business-hours profile, only the timezone reaches the model. The agent is never told whether it is open, when it opens next, or what the schedule is, so it answers those questions from whatever the operator typed into the system prompt, which drifts from what the system actually enforces the moment the schedule changes.

`buildPromptVars` could expose what the resolver already computes:

- `{{esta_aberto}}` / `{{is_open}}` → yes / no
- `{{proximo_atendimento}}` / `{{next_open_at}}` → "Monday, 08:00"
- `{{horario_atendimento}}` / `{{business_hours}}` → the summary `formatWindowsSummary` already renders

The agent then stops reciting hours from memory and starts describing the schedule the gate enforces.

## The one caveat

These belong in the prompt's trailing block, the way `{{data_hora_atual}}` already does, so prefix caching survives. `is_open` only changes at a window boundary, but a variable that lands in the prefix invalidates the cache on every turn that crosses one.

## Origin

Filed originally as a section of #129 marked "optional", which is what makes it a separate issue rather than a part of that one: a PR could close #129 without ever writing these, and then the part the customer actually notices would be the part that quietly did not ship.

Ordering: this reads whatever the resolver returns, so #129 landing first makes it correct about holidays and half-days too. It is not blocked by it, and shipping it before only means it is correct about the weekly grid.

Related: #129.

PR #148, opened 2026-08-20
Model holidays, shutdowns and half-days as date exceptions on a schedule

### LINKED ISSUE #129: BusinessHours only models the week, so holidays, shutdowns and half-days do not exist for the reactive gate, follow-ups or bookable slots
> Line references checked against `v1.8.0` (`74aa261`), which is what `main` currently points at.

## What exists today

A `BusinessHours` profile is a name, a timezone and an array of weekly windows
(`prisma/schema.prisma:572-586`). The unit of a window is the weekday:

```ts
// src/modules/business-hours/hours.ts:8-13
export const windowSpecSchema = z.object({
  day: z.number().int().min(0).max(6),   // 0=Sun … 6=Sat
  start: z.string().regex(/^([01]\d|2[0-3]):[0-5]\d$/),
  end: z.string().regex(/^([01]\d|2[0-3]):[0-5]\d$/),
});
```

Every time decision derives from that — `isOpenAt` (`hours.ts:72`), `isOutOfHoursNow`
(`hours.ts:85`), `fitsWithinWindows` (`hours.ts:98`) and `nextOpenAt` (`hours.ts:180`) — and a
single profile feeds **four** independent consumers:

| # | Consumer | Where | What it decides |
|---|---|---|---|
| 1 | Reactive availability gate | `src/modules/chatwoot/webhook.ts:562-575` (`outOfHoursGate`) | Whether the agent replies at all, or stays silent with a one-shot private note |
| 2 | Follow-up | `src/modules/followups/handlers.ts:366-376` | Reschedules the nudge to `nextOpenAt` instead of messaging out of hours |
| 3 | Console indicators | `src/modules/agents/service.ts:203`, `src/modules/conversations/service.ts:208, 812, 845, 946` | The "out of hours" badge and the next-follow-up ETA |
| 4 | Bookable slots | `src/modules/integrations/toolpacks/calendar-slots.ts:66-74` | Which times `calendar_check_availability` may offer |

There is one editor component (`BusinessHoursForm.tsx` + `SchedulePicker.tsx`), reused by the agent
editor (Availability and Follow-up) and by the Calendar integration. A new profile defaults to
Mon–Fri 08:00–18:00 (`BusinessHoursForm.tsx:27-31`).

## The gap

**A holiday is a date, and the model only knows weekdays.** There is no way to express "closed on
September 7", "December 24 until noon", "shut down from Dec 23 to Jan 2". The only representation
available is editing the weekly grid the day before and undoing it afterwards — by hand, per
profile, with no record of why.

This is not something a prompt can work around, for three independent reasons:

1. **Consumer #1 is a runtime gate, not a prompt.** `outOfHoursGate` decides before the turn runs.
   The model is never consulted and cannot decline to answer.
2. **#2 and #3 compute a date.** `nextOpenAt` walks the calendar; no instruction changes the instant
   it returns.
3. **The agent is never told whether it is open.** Of the whole profile, only the timezone reaches
   the prompt, and only to drive the clock (`src/graph/prepare.ts:420-430`). The interpolable
   variable list (`src/graph/prompt.ts:79-94`) has `nome_contato`, `hora_atual`, `data_atual`,
   `nome_empresa`… and nothing about being open or closed. So the opening hours have to be retyped
   in prose inside the system prompt, where they become a second source of truth that nothing keeps
   in sync with the profile.

## What actually breaks

A metals distributor, quoting triage over WhatsApp, open Mon–Fri 08:00–11:30 and 13:00–18:00.
Scenario: **September 7 (Brazilian Independence Day), falling on a Monday.**

- Gate #1 sees `day=1, 09:14` inside a window and **lets the turn through**. The agent works the
  conversation as if it were an ordinary Monday.
- The prompt tells it to promise a human callback "within 30 minutes during business hours" — and
  the model believes it *is* business hours, because it is Monday morning. The promise reaches the
  customer and no one keeps it.
- Follow-up #2 fires: `isOpenAt` is true, so the nudge goes out on the holiday.
- Badge #3 shows the agent as open to the operator.
- With booking enabled, #4 would offer slots on the holiday.

None of those four symptoms is visible when reading the conversation: every component did exactly
what it was configured to do. The failure only exists against a calendar the product does not have.

The workaround left is hedging in the copy — that agent's knowledge base literally carries the line
*"opening hours may change on commemorative dates"*. That sentence exists only because the correct
information is not representable.

## The concept already exists — on one of the four ends

Worth recording, because it reframes this from "new feature" to "missing symmetry": issue #1
(*calendar_list_events is scoped to the current contact, so the agent cannot read a blocking
calendar*) was closed by adding `blockingCalendarIds` to the Google Calendar integration —
calendars the agent **respects but never operates on**, whose events block availability
(`toolpacks/google-calendar.ts:91, 872-873`; `mcp/server.ts:1850`).

That covers consumer #4 and only #4, at two costs:

- It requires a Google account and a configured Calendar integration. An agent that only does
  WhatsApp triage — no booking at all — has nowhere to declare that it is closed on December 25.
- The very `BusinessHours` profile that **bounds** those slots (`businessHoursId` in the
  integration config) still has no notion of a holiday. The holiday comes in from the side, through
  a different calendar.

So "this date is blocked" has already been accepted and shipped as a product concept. It just does
not exist in the structure the other three consumers read.

## Proposal

A second dimension on the same profile: **date exceptions**, which replace the weekly grid for that
day.

### Data

A new `exceptions Json @default("[]")` column on `business_hours` (additive; existing profiles keep
`[]` and behave exactly as today). Each entry:

```ts
{
  date: "2026-09-07",        // YYYY-MM-DD; or "09-07" when recurring=true
  recurring?: boolean,       // fixed date that repeats every year
  label?: string,            // "Independence Day" — for the editor and the operator
  windows: WindowSpec[]      // [] = closed all day; otherwise replaces that day's grid
}
```

`windows: []` covers holidays and shutdowns; `windows: [{start:"08:00",end:"12:00"}]` covers
half-days (Christmas Eve, Ash Wednesday) without inventing a second concept. A range can be N
entries or, if preferred, an optional `dateEnd`.

### Resolution

One function in `hours.ts`:

```ts
effectiveWindowsForDate(schedule, tz, at): WindowSpec[]
// an exception matching the local date → its windows (possibly empty)
// otherwise → the weekly windows for that weekday
```

`isOpenAt`, `isOutOfHoursNow`, `fitsWithinWindows` and `nextOpenAt` call it instead of filtering on
`w.day` directly. The change is contained, and all four consumers inherit holidays **without any of
them being touched** — which also preserves what `hours.ts` is today: pure, no I/O, unit-testable
against fixed instants.

### Surfaces

- **UI**: a second section in `BusinessHoursForm` ("Exceptions" / "Holidays and closures") — date,
  label, closed-or-windows. `SchedulePicker`'s `ScheduleSummary` could append "+3 exceptions".
  Copy in pt-BR and en (`docs/i18n.md`).
- **MCP/REST**: `business_hours_create` / `business_hours_update` take an optional `exceptions`,
  with the same `assertValidWindows` validation plus a server-side date-format check.
- **Seeding holidays**: a "add <year>'s national holidays" button would help a lot in practice, but
  it is separate scope — the movable ones (Carnival, Good Friday, Corpus Christi) derive from
  Easter and municipal ones vary by city. I would ship the model first and discuss the generator
  after.

## Things worth deciding before implementing

1. **`nextOpenAt` scans 14 days** (`hours.ts:194`) and returns `null` when it finds nothing. That is
   harmless today. With exceptions, an end-of-year collective shutdown easily exceeds 14 days, and
   in the follow-up handler `next === null` falls into `return { outcome: "done" }`
   (`handlers.ts:373-374`): the sequence is **silently ended**, not deferred. If exceptions land,
   the horizon has to grow (60 days?) and `null` needs a better destination.
2. **Half-days are the common case**, not an exception-of-the-exception — hence `windows` on the
   entry rather than a `closed` boolean.
3. **Annual recurrence only applies to fixed dates.** Movable holidays need a dated entry per year;
   the editor should make that obvious so nobody marks Carnival as "repeats yearly".
4. **Mirror.** The `source` column already reserves `CHATWOOT_MIRROR`
   (`business-hours/service.ts:13-15`), and Chatwoot's own working hours have no holiday concept
   either. A re-sync must not wipe locally added exceptions.
5. **Timezone and day boundaries**: an exception matches on the profile's **local** date, via the
   `localYmd` helper `hours.ts:160` already has. Worth a test right on a DST boundary.
6. **Compatibility**: a profile with no exceptions must produce exactly today's behavior — the most
   important regression test in the PR.

## Minimum scope I would propose for a first PR

Column + `effectiveWindowsForDate` + the four consumers inheriting it + the exceptions editor +
MCP/REST + fixing the `nextOpenAt` horizon. Out of scope: the national-holiday generator, the
prompt variables, .ics import.

## Environment

- Observed on agents **1.8.0** (current release), WhatsApp triage agent over Chatwoot, availability
  profile Mon–Fri with two windows, timezone `America/Sao_Paulo`.
- Code references checked against tag `v1.8.0` (`74aa261`); `main` is at the same commit as of
  2026-08-19, and the cited files are identical between the two.

## Scope

One deliverable: the exception model, its resolution in `hours.ts`, and the surfaces that carry it
(editor, MCP/REST). Two things named in this body that are not part of it:

- the prompt variables that let the agent describe the schedule (#144). Marked optional here, which
  is exactly why they are their own issue: a PR could close this one without them and the part the
  customer notices would be the part that quietly did not ship.
- seeding a year's national holidays, already called separate scope above.

gabrieljablonski:
Closes the gap between what a `BusinessHours` profile can express and what a business actually keeps: a holiday, a collective shutdown, a half-day.

## What was broken

A profile modelled the week and nothing else — `{day, start, end}`, with `day` a weekday index. So there was no way to say "closed on September 7", and the four independent consumers of a schedule each did exactly what they were configured to do on a date the product could not represent:

| Consumer | On a holiday, before this |
|---|---|
| reactive availability gate (`outOfHoursGate`) | the turn ran; the agent worked the day as an ordinary Monday |
| follow-up scheduler | `isOpenAt` was true, so the nudge went out |
| console badge + follow-up ETA | showed the agent open, and estimated a send inside the closure |
| bookable slots (`calendar_check_availability`) | offered times on the holiday |

None of it is visible when reading the conversation, which is what makes it expensive.

## The shape

A schedule gains a second dimension: **date exceptions that replace the weekly grid on the dates they match**.

```jsonc
{
  "date": "2026-09-07",      // YYYY-MM-DD
  "dateEnd": "2027-01-02",   // optional, inclusive
  "recurring": true,         // optional: compare month-day only
  "label": "Independence Day",
  "ranges": [{ "start": "08:00", "end": "12:00" }]  // [] = closed all day
}
```

`ranges: []` covers holidays and shutdowns; a non-empty `ranges` is the half-day, with no second concept invented for it. Resolution lives in one function, and a **dated** exception outranks a **recurring** one so a standing yearly rule can be overridden for a single year without deleting it.

Additive by construction: the column defaults to `[]`, which resolves to "the weekly grid decides alone" — the pre-change behavior, byte for byte. There is no backfill.

### Every decision takes a `Schedule`, not a `WindowSpec[]`

This is the load-bearing choice. Exceptions cannot travel inside an array of weekly windows, and **nine** reads across the codebase did `select: { windows: true, timezone: true }` and passed the array on. An optional second argument would have compiled at all nine and silently kept ignoring holidays — the exact failure being fixed. Making the parameter a schedule made the compiler name every site that had to carry the new dimension.

### `nextOpenAt` now scans a year, and says so when it finds nothing

A weekly grid always repeats inside 14 days, so the old horizon was always enough and `next === null` was **unreachable** — which means the `return { outcome: "done" }` behind it in the follow-up handler was dead code. A date span is not bounded by the week, so exceptions are what make that branch live, and the first time it ran it would have ended a follow-up sequence in silence. The horizon is now a year, and the handler logs when a schedule never opens instead of dropping the nudge quietly.

## Three things in the issue that do not hold

- **"All four consumers inherit holidays without any of them being touched" is not true.** See above: nine reads and five signatures change. Not a reason to reject the design, but it is the real size of the change.
- **A fifth surface the issue does not name: agent export/import.** The bundle carries a schedule as raw JSON, so nothing in the type system would have complained — an exported agent would simply have arrived at the destination with every holiday gone. It is covered here.
- **The Chatwoot-mirror concern costs nothing, because there is no mirror.** `CHATWOOT_MIRROR` appears twice in the tree (a comment and a UI badge); `businessHours.create` always writes `source: "LOCAL"`, and no working-hours sync exists. "A re-sync must not wipe locally added exceptions" is a decision about a path that has never been built.

## Choices that differ from the issue's proposal

- **One date format.** `date` is always `YYYY-MM-DD`, and `recurring` changes the comparison to month-day. The proposed `"09-07"` form would have put two formats in one field while both validation and matching branch on the flag anyway.
- **An exception's hours are `ranges`, not `windows`.** They carry no `day` — the date already fixes it — and reusing the weekly shape would have meant a field that is silently ignored, plus two different `windows` schemas at two nesting levels in the same request body.
- **`dateEnd` ships now.** Without it a Dec 23 → Jan 2 shutdown is eleven hand-typed entries, and shutdowns are precisely the case that motivated the horizon fix.

## Validation scope

**Proven by test.** Decision table in `tests/modules/business-hours.test.ts` at fixed instants (34 cases): closure, half-day, an exception that opens a day the grid leaves closed, inclusive spans, recurring, a recurring span wrapping the year end, a recurring Feb 29 only matching leap years, dated-beats-recurring, array order within a kind, local-date matching across a DST boundary, and the horizon clearing a 19-day shutdown. Every pre-existing case in that file is preserved unchanged and is the compatibility guarantee.

End to end, DB-backed, in `tests/modules/business-hours-exceptions.test.ts`, against a Chatwoot double that authenticates the way Chatwoot does: the customer's message during a closure is answered by nothing, the operator gets the one-shot private note and the watermark is stamped; the follow-up nudge is deferred past the closure with no message sent; the header reports the closure and names it; the ETA the console shows skips it; a schedule that never reopens ends the sequence and shows no next step instead of a time inside the closure; the schedule the availability tool resolves carries the exceptions. Slots are covered in `toolpacks-google-calendar.test.ts` against a fetch stub standing in for Google (a holiday empties the day, a half-day bounds it), the export/import round trip in `agent-transfer.test.ts`, and the MCP write path plus its validation in `mcp-write-resources.test.ts`.

**Red before.** With the column present but unread — `parseSchedule` returning no exceptions, i.e. the world before this change — three of the end-to-end tests fail: the nudge fires during the closure, the customer's message is answered, and the header reports the agent open.

**Mutation.** 30 mutations against the final tree, one clause at a time, run over the whole file set the CI runs: 28 killed. Two survive and are declared rather than dressed up:

- `nextOpenAt`'s `start > at` (relaxing it to `>=` changes nothing). The early `isOpenAt` return already answers the case where `at` sits at a range start, so no input reaches that comparison with `start === at`. A strictness guard with no constructible input, not an uncovered rule; it now carries a comment saying so, because the next reader's instinct will be to simplify it.
- the severity of the log line the handler writes when a schedule never reopens (`warn` → `info` survives). The behavior — the sequence ends, nothing is sent — is covered; the severity is not, and this repository has no pattern for asserting logger output. Worth stating plainly: a stdout warn is weaker observability than a flowlog row, and there is no flow context in that handler to write one without introducing a mechanism this change does not otherwise need.

**Not validated.** No live run against a real Chatwoot instance or a real Google Calendar — both are exercised against doubles here. The end-to-end assertions deliberately test only the CLOSED direction, because the open direction depends on the wall clock at run time; the open direction is pinned in the unit table at instants that cannot drift. Nothing was measured for performance: the year-long scan only runs on the closed path and does no timezone work on a day with no candidate range, but that is reasoning, not a measurement.

## Review

Four rounds. Round 1 returned three findings, all real and all the same root — a date exception makes reachable a branch a weekly grid never could:

- the console's follow-up estimate fell back to the ungated time when nothing reopens, promising a follow-up the handler ends;
- `parseExceptions` validated the array as one unit, so a single malformed entry discarded every valid holiday and silently restored the weekly hours;
- a **dated** span running backwards was read as wrapping the year end, closing almost the whole year instead of nothing.

The last two are reachable because agent import writes the column as raw JSON, bypassing `assertValidExceptions`. They are fixed in the readers, where the contract that a hand-edited row cannot widen availability already lives, rather than by adding a third validation mechanism on the import path.

Round 3 (after rebasing onto #147, which invalidated the earlier clean round) returned two more:

- **P1** — the never-reopens path returned `done` without stamping `lastFollowUpAt`, so the sweep matched the same episode on the next pass and every eligible conversation re-entered the year-long scan once a minute, indefinitely. It now abandons the episode with a watermark, exactly like the retry-exhaustion path beside it.
- a third caller of `BusinessHoursForm` — the agent editor's reuse-review modal — built its `initial` without `exceptions`, so the form hydrated `[]` and its unconditional PATCH deleted every holiday the operator had, from a screen opened to change something else. The field is now **required** on the props type, which is what stops a fourth caller from repeating it; the form's own round trip is covered by a DOM test.

Round 4 came back clean.

## Out of scope

The prompt variables that would let the agent describe its own schedule (#144), seeding a year of national holidays, and `.ics` import — all three named as separate in the issue.

Fixes #129


JotaSXBR:
Well done!

PR #159, opened 2026-08-20
fix(chatwoot): make handoff customer message terminal

### LINKED ISSUE #158: handoff_to_human can send the same closing reply twice when its mirror event lags
## Problem

`handoff_to_human` sends `customerMessage` immediately from inside the tool, then sets the conversation to `open`. The graph may still return a final assistant reply with the same closing text.

Before posting that final reply, `runLoadedTurn` rechecks status/assignee in the local mirror. If Chatwoot's `open`/assignment event has not reached the mirror yet, the row still looks bot-owned and the runtime posts the final reply too.

The proactive `runAgentNudge` path has the same two delivery paths.

## Reproduction

1. Start a turn on a pending, bot-owned conversation.
2. Have the model call `handoff_to_human({ customerMessage: \"A human will continue.\" })`.
3. Do not mirror the tool's `toggleStatus(\"open\")` before graph completion (normal webhook lag).
4. Have the model return a final reply such as `A human will continue!`.

Observed sends: `customerMessage`, then the final reply seconds later.

## Expected

Once a handoff `customerMessage` is successfully delivered, it is terminal for customer-facing text in that turn. Reactive turns and nudges must not post the model's final reply or nudge post-actions afterward. If the customer-message send fails, the final reply remains available as fallback.

## Scope

This is not inbound webhook duplication and not split retry. It is one turn with two distinct send owners: the handoff tool side effect and the runtime/nudge final post.

debaotrader:
Fixes #158

## What changed

- add an isolated per-turn handoff delivery signal;
- set it only after `handoff_to_human.customerMessage` is successfully posted;
- make both `runLoadedTurn` and `runAgentNudge` treat that delivery as terminal;
- drop pending reactive post-actions after the handoff;
- keep the final model reply available when the customer-message send fails or is omitted.

## Why

The handoff tool posted `customerMessage` before Chatwoot's `open`/assignee event reached the mirror. The graph then returned a similar final reply; the stale mirror recheck still looked bot-owned, so runtime/nudge posted it too.

This is one turn with two send owners, not webhook retry or split duplication.

## Tests

- regression: reactive turn with delayed mirror event posts only `customerMessage`;
- regression: nudge with delayed mirror event posts only `customerMessage` and skips post-actions;
- unit: successful handoff send marks terminal state;
- unit: failed handoff send does not mark terminal state.

Local evidence:

- `bun test tests/graph/tools.test.ts tests/modules/send-image.test.ts` — 65 pass, 0 fail;
- `bun run build-check` — pass;
- focused Biome check — pass;
- full pre-commit `bun check` — 2127 pass, 0 fail, 951 DB-gated skips;
- independent re-review — `APPROVED_CLEAN`.


gabrieljablonski (commented):
Reviewed the diff merged onto current `main`, because the branch is 11 commits behind and conflicts in `src/graph/nudge.ts` and `tests/graph/nudge.test.ts` after #134. Everything below was measured against a live test database, comparing the call sequence with and without the early return.

**What holds up, first.** The diagnosis is right, and it is the hard part: one turn with two send owners. The three regression tests are real ones. I removed each guard in turn (the flag assignment in the tool, then each early return) and every removal takes a test down with it. The `onSideEffectError` case asserting the flag stays `false` when the send throws is a good catch.

Four things below. The first is the one that matters.

## 1. The flag says "the closing line went out", but it is read as "the handoff happened"

`sendPrivateNote` and `toggleStatus("open")` are not inside a `try`. When either throws, the exception leaves the tool, LangGraph turns it into a tool error, and the model gets another step in which it can write a recovery reply. The flag is already `true` by then, so both callers suppress that reply.

Measured, with a `toggleStatus` that throws:

```
before this PR   sendMessage("Um humano já te atende")
                 toggleStatus  ->  502
                 sendMessage("Desculpe, não consegui transferir. Vou seguir te ajudando.")

with this PR     sendMessage("Um humano já te atende")
                 toggleStatus  ->  502
                 (nothing)

conversation status in both cases: pending
```

`pending` means the conversation is still the bot's. No human was engaged, the queue received nothing, and the one sentence that could have recovered it was dropped. The customer was told a human is coming and none is.

This is not a corner case: `sendPrivateNote` runs on every handoff where the model supplies a `reason` and `transferWithSummary` is on, which is the default.

The gate wants to be "the transfer completed", set after `toggleStatus` succeeds, rather than "the customer message was sent".

## 2. The proactive path jumps over the 24h service window and the post-actions

`runAgentNudge` makes a decision `runLoadedTurn` never has to: outside WhatsApp's 24h service window, a free-form message is exactly the send the provider refuses, so the code picks an approved template or leaves an explanatory private note. The early return sits before that decision.

A 48h follow-up that hands off:

```
before this PR   outcome "noted-window"
                 private note: "⏳ Fora da janela de 24h do WhatsApp: a mensagem abaixo NÃO foi enviada..."
                 labels ["follow-up"]

with this PR     outcome "messaged", no note, no labels
```

Two losses there. The note is how an operator finds out a follow-up was fenced. And the labels: dropping `resolve` after a handoff is right, but the branch immediately below keeps labels deliberately, with the reason written into the code, "labels still apply so the operator can triage the fenced conversations".

`messaged` is also not true outside the window, since free-form is precisely the send that does not arrive.

## 3. An image queued earlier in the same turn is dropped

`send_image` queues and the runtime delivers, after the same gates the reply passes (#65). A turn that queues a picture and then hands off:

```
before this PR   sendMessage(closing)  toggleStatus(open)  sendFileAttachment(imagem.png)  sendMessage(final text)
with this PR     sendMessage(closing)  toggleStatus(open)
```

Dropping the final text is the fix. Dropping the attachment is collateral: a photo is not a duplicate of the closing line, and the tool already told the model it was queued.

## 4. Two of the four lines in the reactive early return are dead

`turnState` is local to `runLoadedTurn`, and nothing reads `pendingImages` or `resolveRequested` after the return. Removing either line breaks no test, and not because they are untested: they do nothing. The drop the description promises happens because of the `return`, not because of them.

## On the evidence in the description

The `bun check` you report shows 951 DB-gated skips, and both new regression tests live inside `describe.skipIf(!dbUp)`, so they were in that set. They do pass with a database up; I ran them.

## What happens next

The branch allows maintainer edits, so rather than handing you a list I am pushing onto it: `main` merged in with the #134 conflict resolved, then the four fixes above. The PR stays yours and your commit stays first. If you would rather make these changes yourself, say so and I will stop.

The deeper shape, where the tool records the intent and the runtime owns delivery, is what #160 asks for and I am deliberately not doing it here. It carries an open design question (what a guardrail suppression should mean for a handoff line) that does not belong inside a bugfix. Your own sentence about two send owners is the right diagnosis for that one.

gabrieljablonski:
Second review round on the merged branch, three more findings, all measured the same way. Both are pushed.

**The deferred resolve was keyed on the wrong event.** It fell with the suppression of the final text, but those two come apart exactly when the closing line fails to send: the customer never heard it, so the model's reply still has to go out, while the conversation is already a human's. A `resolve_conversation` from earlier in the same turn would then close an open request out from under whoever just received it. It is keyed on the transfer now, on both paths.

**The ownership recheck called our own handoff a takeover.** It reads the row the tool has just set to `open`, concludes the conversation is no longer ours, and returns `taken-over` before anything is delivered. So the image restored in the first round only arrived while Chatwoot's event happened to lose a race. A handoff this turn completed is not a takeover, and delivery should not depend on which side of that race wins.

**And a correction to my first review, which matters for anyone reading it later.** I labelled the 24h-window reproduction "a 48h follow-up". It is not one: an inactivity follow-up runs with `requireLiveBotOwnership` (`src/modules/followups/handlers.ts`), so its check before delivery is a live GET rather than the mirror, and that GET reports the conversation as no longer the bot's. What I measured was an event-driven nudge, which does not set that flag. The finding stands for those; for inactivity follow-ups the mechanism is a different one, and worse:

```
before this PR   live probe -> not-owned -> "stale"     episode ends, no watermark, no next step
original commit  shortcut answers first  -> "messaged"  watermark stamped, NEXT STEP SCHEDULED
```

Which means the ladder would have gone on chasing a conversation a human had just been handed. `stale` is the right word there, and keeping the shortcut inside the freeform branch preserves it. There is now a test for each of the two triggers, and the one for inactivity follow-ups fails if the shortcut moves back above the probe.

Eight mutations across the two rounds, each killed by a test: the gate, the flag's position relative to `toggleStatus`, the shortcut's placement, the resolve on both paths, the recheck, the image branch and its throw. `bun check` is green locally at 3232 tests, and CI is now running on the PR, which it had never done: fork workflows were sitting on `action_required` since the day you opened it. That was on us, not on you.

One thing I am deliberately leaving alone, because it is older than this PR and wants its own issue: the tool sends `customerMessage` with `sendMessage` and never consults the 24h service window, so on a proactive turn outside it that closing line is a free-form send the provider refuses. The nudge path has the whole `proactiveSendMode` machinery for exactly this and the tool does not use it.

gabrieljablonski:
Third round, one finding, and it was against the previous round's fix rather than against yours: letting a completed handoff skip the ownership recheck also skipped it for a human who grabbed the conversation in that same window. The mirror stores no reason for a status change, so our own transition and a genuine takeover look identical there.

Pushed the removal rather than a fourth condition. The gate exists for the takeover, so it fails closed for both again, and the bound that leaves is now a test instead of a surprise: once the mirror reflects the status the tool set, the turn ends at that gate and an image queued before the handoff does not go out. Telling the two apart needs the mirror to record who caused the change, which is its own piece of work.

Worth saying plainly, since it cuts against a claim in my first review: the queued-image delivery only holds while Chatwoot's event is still in flight. That is a bounded improvement over dropping it unconditionally, not a complete one.

Where the branch stands: three review rounds, eight findings, nine mutations each killed by a test, `bun check` green at 3232 locally, CI green on the PR.

gabrieljablonski:
Fourth round, one finding, and this one I am waiving rather than fixing, with the reason on the record so the next round does not re-open it.

The finding is that on an audio turn (`mirror` mode after a voice note, `preference` mode for a contact who asked for audio) suppressing the final text means the handoff produces text only, ignoring the configured modality.

True, but it is not what this PR changed. `handoff_to_human` has always written its closing line with `client.sendMessage`, which is text, and it still does. What an audio-preferring customer used to receive was the closing line as text **and then the model's paraphrase of it as audio** — the duplicate this PR exists to remove. The modality was never honoured for the handoff line itself; it was honoured for the second copy. Removing the duplicate makes a pre-existing gap visible, it does not create it.

Fixing it properly means the tool recording the intent and the runtime owning delivery, so the closing line passes the output guardrail, the split/typing pacing and the TTS branch like any other reply. That is #160, and the same change closes it. Doing it inside a bugfix would also force the design question I flagged earlier, which is what a guardrail suppression should mean for a handoff line.

So: three rounds of findings and one waiver, nine mutations each killed by a test, `bun check` green at 3232 locally, and CI green on the PR.

PR #162, opened 2026-08-20
Give the credential endpoints a bucket of their own, and refuse a config that deletes one

gabrieljablonski:
`/api/auth/login`, `/signup`, `/setup` and `/accept-invite` were bounded only by the global 600/min. There is no lockout, no failed-attempt counter and no backoff anywhere else in the auth path, so that number **was** the brute-force protection: 600 password guesses a minute against one account. `/accept-invite` consumes a token from an unauthenticated request body, so it is a guessing target too.

`strictRateLimitMiddleware` was written for exactly this, and mounted nowhere.

## Mounting it as written would have been worse than leaving it out

The plugin seeds itself with `max:duration:scoping` and Elysia deduplicates matching plugins. It shipped as `(10, 60000, scoped)`, which is the DCR limiter **exactly**. Measured on the real app with that collision in place:

```
POST /api/auth/login              -> RateLimit-Limit: 600   ← the GLOBAL bucket
POST /api/v1/mcp/oauth/register   -> RateLimit-Limit: 10
```

The credential limiter never mounted. Nothing was logged. The endpoints were no better off than before, while every reader of `src/app.ts` would have said they were protected.

That is also what the tests here pin: reintroducing the collision fails them with `Expected: "20", Received: "600"`.

## Why the window is minutes

A per-minute ceiling resets 60 times an hour, so a 10/min credential limit still allows 600 guesses an hour. The window is what bounds brute force, not the number.

**20 per 5 minutes** is 240 an hour and still absorbs a burst, which is what a whole office signing in at once looks like through a single NAT. Having a different `duration` also puts the collision out of reach *by construction*, rather than by a `max` that happens not to clash.

## Boot refuses the configuration that deletes a bucket

`config.ts` throws, naming both limiters, when any two of the five share a budget **and** a window. It also refuses a credential budget that is not tighter than the global one, compared as rates since the windows differ.

Both checks are exported as plain functions so a test can drive them. Without them the trap above is reachable again the next time anyone adds a limiter, with a green suite. The failure is not symmetric, which is why it throws rather than warns: whichever limiter loses is the only thing counting its traffic, so a collision never tightens anything, it deletes a bucket.

## What is in the bucket, and what is not

| route | in | why |
| --- | --- | --- |
| `POST /api/auth/login` | yes | password guessing |
| `POST /api/auth/signup` | yes | account creation |
| `POST /api/auth/setup` | yes | the one-time setup token |
| `POST /api/auth/accept-invite` | yes | consumes an invite token, unauthenticated |
| `GET /api/auth/invite` | yes | **a token oracle**: `security: []`, 200 with the invited email and role for a valid token, 404 otherwise |
| `HEAD /api/auth/invite` | yes | the **same** oracle: Elysia dispatches HEAD to the GET handler, the lookup runs, and the 200/404 is the whole answer |
| `GET /api/auth/me` | no | polled on every page load; would lock the app out of itself |
| `PATCH /api/auth/password` | no | behind a session; including it lets a stolen session lock the owner out of changing their password |
| `POST /api/auth/google` | no | needs a token Google signed, which is not a thing to guess |
| `POST /api/auth/logout` | no | nothing to guess |

Matching is **METHOD + path**, not a path set gated on POST. The first draft said POST-only "because these paths only exist as POST", and review round 2 showed that is false: the credential travels in the query string on `GET /auth/invite`, which is a cheaper oracle than the POST that consumes the same token.

Round 3 found the second half of the same gap: Elysia dispatches `HEAD` to the `GET` handler, so `HEAD /auth/invite` runs the same lookup and answers with the same 200/404, which is everything an oracle is worth. It is folded into `GET` **inside** the matcher rather than listed as another literal, so it widens exactly the routes `GET` already covers and nothing else.

Pairing the method with the path lets that one GET in while keeping every other GET out, which matters because a 404 spends whatever budget covers it. Measured on the real app:

```
GET  /auth/invite?token=      -> 404 limit=20    ← the credential bucket
HEAD /auth/invite?token=      -> 404 limit=20    ← the same oracle, the same bucket
GET  /auth/invite/ (slashed)  -> 404 limit=20
POST /auth/invite (404)       -> 404 limit=600   ← a method the route does not serve
GET  /auth/me                 -> 200 limit=600
HEAD /auth/me                 -> 200 limit=600
GET  /auth/login (404)        -> 404 limit=600   ← no crawler burn on the login budget
HEAD /auth/login (404)        -> 404 limit=600   ← folding HEAD into GET widens nothing else
```

One trailing slash is canonicalized, sharing the helper the DCR limiter already needed since Elysia answers both spellings from the same handler.

## Compose files

They now pass all four rate-limit variables. They listed **none** of them, including the two that already existed, so every ceiling was unreachable in every shipped deployment. That is tolerable for a runaway guard at 600 and not for this one: when a shared office address trips the credential bucket, raising the max is the operator's only remedy. Adding just the new pair beside two that stay unreachable would have left the file incoherent.

This reaches **new** installations. Coolify freezes a `${VAR:-default}` at creation, so an existing installation keeps the compose it was created from and the operator has to add the variable in the UI to raise a ceiling there.

## A budget the limiter cannot honour (review round 1)

The budgets were parsed with `RAW && Number(RAW) > 0`, which accepts `Infinity`, and `1e309` parses to exactly that. Measured:

```
200 reset=NaN rem=1
200 reset=NaN rem=0
429 reset=NaN rem=0
429 reset=NaN rem=0     ← and forever after
```

With an infinite window the limiter advertises `RateLimit-Reset: NaN` and the bucket **never resets**, so the credential endpoints answer 429 permanently once the budget is spent. A config typo would have locked login for good, recoverable only by an edit and a restart.

That shape was on all four budgets, including the two that predate this change, so all four now go through one parser requiring a positive **whole** number. It throws by name rather than falling back, because falling back is how a typo becomes a budget nobody chose, the same reasoning `TRUST_PROXY` already carries in this file.

Round 2 found the other half of the same defect: `Number.isInteger(1e12)` is true, but 1e12 minutes in milliseconds is past `Date`'s range, so a large **finite** window reaches the same invalid reset. The answer is a bound rather than another finiteness check. A window caps at a day (`MAX_WINDOW_MINUTES = 1440`), well past any real throttle and far below anything `Date` cannot represent; the budgets cap at `MAX_BUDGET = 1_000_000` for the mirror-image reason, since an absurd ceiling is a bucket that never trips while reading as a configured one.

## Validation scope

**Proven by test.** `tests/api/middlewares/credentialRateLimit.test.ts` goes through the real app and reads which bucket answered off the ceiling it advertises, so it checks that **both** buckets exist rather than that one of them answers. Reintroducing the collision fails it at `Expected: "20", Received: "600"`.

Those numbers are the shipped defaults, and `tests/setup.ts` now **pins** the four `RATE_LIMIT_*` variables at preload so they stay the numbers under test. Round 5 caught the tests asserting them while this change is what made them configurable: a developer tuning `RATE_LIMIT_CREDENTIAL_MAX` in their `.env` would have watched a correct app fail, and fail with the collision's own signature.

Reading the expectations from `config` instead was the obvious answer and the wrong one, which round 6 rejected: the boot check keeps the (budget, window) **pairs** distinct, not the budgets, so setting the credential budget **to** the global one is supported and boots fine, and then both buckets advertise the same ceiling with nothing able to tell them apart. Pinning is what keeps the signal. Measured with the pin, `RATE_LIMIT_CREDENTIAL_MAX=50`, `=600`, `RATE_LIMIT_USER_PER_MIN=900` and `RATE_LIMIT_CREDENTIAL_WINDOW_MINUTES=1` all run green; measured without it, `RATE_LIMIT_USER_PER_MIN=5` fails five of the eight tests in `rateLimitMetering.test.ts`, the same dependency sitting in a file this change never touched.

Round 7 asked whether the probe route in `rateLimitMetering.test.ts` still routes, since it is registered on the shared app singleton in a `beforeAll` that now runs after this file has already called `listen()` on it. Measured, it does: 404 with the thrown body, alone and under the full suite, in one process shared with the new file. The blind spot behind the question was real, so that test now asserts the status and the body before the cost: an unregistered `/__metering/*` path falls through to the SPA catch-all, answers **200**, and costs exactly the same 1, so the old cost-only assertion would have stayed green while testing nothing. With the path mutated to one that was never registered it now fails at `Expected: 404, Received: 200`.

Full suite green on both trees at this commit: 3152 tests on the source tree and 3112 on the derived public tree, 0 failures on either.

**Not validated live.** No deployment was exercised; every number comes from the app served on a local port.

**Not covered.** This bounds guessing per IP. It is not account lockout, and it does not help against a distributed attempt from many addresses; everyone behind one NAT still shares the bucket, which is the trade `.env.example` states.

---

**No `Fixes #N`**, same deviation as #155, #156 and #157.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -151,6 +151,19 @@ FLOWLOG_RETENTION_DAYS=30
 # through a single address. Raise if a shared-NAT deployment trips them. Defaults: user 600, MCP 1200.
 RATE_LIMIT_USER_PER_MIN=600
 RATE_LIMIT_MCP_PER_MIN=1200
+# NOTE: The credential endpoints (login, signup, the one-time /setup token, accept-invite) get a
+# SECOND, tighter bucket on top of the global one. The window is minutes rather than one minute on
+# purpose: a per-minute ceiling resets 60 times an hour, so 10/min would still allow 600 password
+# guesses an hour against one account. 20 per 5 minutes is 240 an hour and still absorbs a burst,
+# which is what a whole office signing in at once looks like through a single NAT. Raise the max if
+# yours trips it. Boot refuses a value that collides with another limiter's budget+window (they are
+# deduplicated by the plugin, and the loser silently stops limiting) or that is not tighter than
+# RATE_LIMIT_USER_PER_MIN. All four of these must be positive WHOLE numbers, and boot refuses
+# anything else by name rather than falling back: a window the limiter cannot turn into a date (an
+# `Infinity`, which is also what `1e309` parses to) makes it advertise `RateLimit-Reset: NaN` and
+# never reset, so the endpoints it guards answer 429 for good once the budget is spent.
+RATE_LIMIT_CREDENTIAL_MAX=20
+RATE_LIMIT_CREDENTIAL_WINDOW_MINUTES=5
 
 # NOTE: Who the limits above apply to. The app sees the reverse proxy as the client of every request,
 # so until TRUST_PROXY is on, the whole deployment shares one bucket.
--- a/docker-compose.coolify.yml
+++ b/docker-compose.coolify.yml
@@ -27,6 +27,17 @@ services:
       # proxy to trust the CDN before raising it.
       - TRUST_PROXY=true
       - TRUSTED_PROXY_HOPS=${TRUSTED_PROXY_HOPS:-1}
+      # NOTE: rate-limit ceilings, per client address. The first two are runaway guards, generous on
+      # purpose. The credential pair is the tight one: it covers login, signup, the one-time /setup
+      # token and accept-invite, and its window is minutes rather than one minute because a
+      # per-minute ceiling resets 60 times an hour. Raise RATE_LIMIT_CREDENTIAL_MAX if a shared
+      # office address trips it; the bucket is keyed by IP, so everyone behind one NAT shares it.
+      # Boot refuses a credential budget that collides with another limiter's budget+window, or that
+      # is not tighter than RATE_LIMIT_USER_PER_MIN. See .env.example.
+      - RATE_LIMIT_USER_PER_MIN=${RATE_LIMIT_USER_PER_MIN:-600}
+      - RATE_LIMIT_MCP_PER_MIN=${RATE_LIMIT_MCP_PER_MIN:-1200}
+      - RATE_LIMIT_CREDENTIAL_MAX=${RATE_LIMIT_CREDENTIAL_MAX:-20}
+      - RATE_LIMIT_CREDENTIAL_WINDOW_MINUTES=${RATE_LIMIT_CREDENTIAL_WINDOW_MINUTES:-5}
       # First-run: disable the /setup token in this trusted, TLS-fronted onboarding window (the
       # advisory lock still caps /setup at one admin). Set to `true` to require it.
       - SETUP_TOKEN_REQUIRED=${SETUP_TOKEN_REQUIRED:-false}
--- a/docker-compose.portainer.yml
+++ b/docker-compose.portainer.yml
@@ -90,6 +90,17 @@ services:
       # ranges first, then raise this to 2.
       - TRUST_PROXY=true
       - TRUSTED_PROXY_HOPS=${TRUSTED_PROXY_HOPS:-1}
+      # NOTE: rate-limit ceilings, per client address. The first two are runaway guards, generous on
+      # purpose. The credential pair is the tight one: it covers login, signup, the one-time /setup
+      # token and accept-invite, and its window is minutes rather than one minute because a
+      # per-minute ceiling resets 60 times an hour. Raise RATE_LIMIT_CREDENTIAL_MAX if a shared
+      # office address trips it; the bucket is keyed by IP, so everyone behind one NAT shares it.
+      # Boot refuses a credential budget that collides with another limiter's budget+window, or that
+      # is not tighter than RATE_LIMIT_USER_PER_MIN. See .env.example.
+      - RATE_LIMIT_USER_PER_MIN=${RATE_LIMIT_USER_PER_MIN:-600}
+      - RATE_LIMIT_MCP_PER_MIN=${RATE_LIMIT_MCP_PER_MIN:-1200}
+      - RATE_LIMIT_CREDENTIAL_MAX=${RATE_LIMIT_CREDENTIAL_MAX:-20}
+      - RATE_LIMIT_CREDENTIAL_WINDOW_MINUTES=${RATE_LIMIT_CREDENTIAL_WINDOW_MINUTES:-5}
       # First-run: disable the /setup token in this trusted, TLS-fronted onboarding window (the
       # advisory lock still caps /setup at one admin). Set to `true` to require it.
       - SETUP_TOKEN_REQUIRED=${SETUP_TOKEN_REQUIRED:-false}
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -43,6 +43,17 @@ services:
       # configured to forward the chain it received; see .env.example.
       - TRUST_PROXY=${TRUST_PROXY:-false}
       - TRUSTED_PROXY_HOPS=${TRUSTED_PROXY_HOPS:-1}
+      # NOTE: rate-limit ceilings, per client address. The first two are runaway guards, generous on
+      # purpose. The credential pair is the tight one: it covers login, signup, the one-time /setup
+      # token and accept-invite, and its window is minutes rather than one minute because a
+      # per-minute ceiling resets 60 times an hour. Raise RATE_LIMIT_CREDENTIAL_MAX if a shared
+      # office address trips it; the bucket is keyed by IP, so everyone behind one NAT shares it.
+      # Boot refuses a credential budget that collides with another limiter's budget+window, or that
+      # is not tighter than RATE_LIMIT_USER_PER_MIN. See .env.example.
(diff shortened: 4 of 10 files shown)


gabrieljablonski:
### Review round 2 — recorded, not published as a review

Round 2 ran clean against `3819e5a` (the PR head at the time, `dirty: false`) and returned two findings. The review bot's token is a 1Password mount that is not serving right now, and I am not passing `ALLOW_HUMAN_FALLBACK`, so this is the record instead. Both findings are reproduced verbatim rather than summarised.

> **[P2] Cover invite validation with the credential limiter** — confidence 0.96
>
> When guessing invitation tokens, a caller can use `GET /api/auth/invite?token=...`, which performs the token lookup and distinguishes a valid token with a 200 response. This predicate excludes every GET and only covers the later accept endpoint, so validation attempts retain the global 600/minute budget and bypass the new credential bucket; include the invite-validation route while continuing to exclude unrelated GETs.

> **[P2] Reject credential windows outside Date's supported range** — confidence 0.99
>
> For a large finite integer such as `RATE_LIMIT_CREDENTIAL_WINDOW_MINUTES=1000000000000`, `Number.isInteger` passes, but converting it to milliseconds produces a duration beyond JavaScript Date's ±8.64e15 ms range. The limiter then stores an invalid reset date, advertises `RateLimit-Reset: NaN`, and never resets after the credential budget is exhausted; validate the converted window against Date's usable range.

**Both accepted.**

The first one contradicts a claim I made in the PR body, so to be explicit about it: I wrote that matching POST-only was right "because these paths only EXIST as POST". That is false for `GET /api/auth/invite`, which is declared `security: []`, looks the token up, and answers 200 with the invited email and role when it exists, 404 when it does not. The credential travels in the query string there, and it is a cheaper oracle than the POST that consumes the token.

Matching is now `METHOD + path` rather than a path set gated on POST, which lets that one GET in while keeping every other GET out. That distinction is load-bearing in both directions and now measured on the real app:

```
GET  /auth/invite?token=      -> 404 limit=20    ← the credential bucket
GET  /auth/invite/ (slashed)  -> 404 limit=20
POST /auth/invite (404)       -> 404 limit=600   ← a method the route does not serve
GET  /auth/me                 -> 200 limit=600
GET  /auth/login (404)        -> 404 limit=600   ← no crawler burn on the login budget
```

The second is the other half of the defect round 1 found, so the fix is a bound rather than another finiteness check: `MAX_WINDOW_MINUTES = 1440`. A day is well past any real throttle, and every window whose millisecond duration `Date` cannot represent is orders of magnitude beyond it. The budgets get `MAX_BUDGET = 1_000_000` for the mirror-image reason — an absurd ceiling is a bucket that never trips while reading as a configured one.

Round 3 follows on the new head.

gabrieljablonski:
### Review round 3 — recorded, not published as a review

Bot token still unavailable (1Password mount), still not passing `ALLOW_HUMAN_FALLBACK`. Round 3 ran against `3146358`, `dirty: false`, and returned one finding, verbatim:

> **[P1] Cover HEAD requests to the invite oracle** — confidence 0.99
>
> Elysia automatically dispatches `HEAD` to `GET` handlers, so `HEAD /api/auth/invite?token=...` performs the same token lookup and exposes the same 200/404 oracle, but this set recognizes only `GET`. These probes therefore count against the global 600/minute bucket instead of the new 20-per-5-minute credential bucket, bypassing the protection added for invite-token guessing; include the `HEAD` alias in this matcher.

**Accepted.** Measured before the fix:

```
GET  /api/auth/invite?token=chute   -> 404 limit=20
HEAD /api/auth/invite?token=chute   -> 404 limit=600   ← the oracle, at 30x the budget
HEAD /api/auth/me                   -> 200            ← proving HEAD reaches GET handlers
```

HEAD leaks the status, and on that route the status *is* the answer, so the whole oracle was reachable at the global budget.

This is the third alias-shaped finding in a row on this family — a trailing slash on the DCR limiter, a query-string credential on `GET /auth/invite`, and now a method alias — so the fix folds HEAD into GET **in the matcher** rather than adding one more literal to the set. Any GET route added later covers its HEAD alias automatically, instead of leaving the same omission to be rediscovered.

It does not widen anything else, which is pinned by test and measured: `HEAD /auth/login` resolves to `GET /auth/login`, which is not in the set, so a crawler still cannot burn the login budget with it.

```
GET  /api/auth/invite?token=chute    -> 404 limit=20
HEAD /api/auth/invite?token=chute    -> 404 limit=20
HEAD /api/auth/invite/?token=chute   -> 404 limit=20
HEAD /api/auth/login                 -> 404 limit=600
HEAD /api/auth/me                    -> 200 limit=600
```

gabrieljablonski:
### Review round 4 — clean, recorded, and NOT published as an approval

Round 4 returned **0 findings** against `fb2476e`, `dirty: false`, matching the PR head:

> `patch is correct` (confidence 0.91) — "The credential limiter is mounted correctly, covers the intended route aliases and methods, and its configuration validation prevents Elysia plugin deduplication. No actionable defects were identified."

CI is green on the same commit (lint, test, type-check).

**This does not clear the gate**, and I want that on the record rather than implied. The bot token is a 1Password mount that has been unavailable since round 2, so rounds 2, 3 and 4 all live in comments under my own account instead of as reviews from `fazer-ai-bot`. I am not passing `ALLOW_HUMAN_FALLBACK`.

The consequence is concrete: the `CHANGES_REQUESTED` review the bot posted on round 1 is still standing against a commit that no longer exists, and nothing has dismissed it. When the mount is serving again, rounds 2 through 4 get republished properly and the approval lands on the current head.

Issue #106, opened 2026-08-17
Follow-up never starts for pre-existing agents when the migration role is the DB owner but not a superuser


gabrieljablonski:
The `follow_up_armed_at` backfill runs an `UPDATE` over a table with FORCE ROW LEVEL SECURITY, without the super-admin escape hatch. The last statement of `prisma/migrations/20260807032257_agent_follow_up_armed_at/migration.sql` is:

```sql
UPDATE "agents" SET "follow_up_armed_at" = NOW();
```

`agents` carries `FORCE ROW LEVEL SECURITY`, which subjects even the table OWNER to the `tenant_isolation` policy; only a superuser (or a `BYPASSRLS` role) escapes it. `docs/deploy.md` documents `MIGRATION_DATABASE_URL` as "superuser **or** DB owner", and on managed Postgres (RDS, Neon, Supabase, and several panel-managed setups) the administrative role is typically the owner **without** `rolsuper`.

On those installs the statement matches **zero rows and reports success**: no error, no warning, `migrate deploy` green.

Why that is not cosmetic: `conversations/service.ts` treats a null `followUpArmedAt` as "not armed", and the column only gates sequence STARTS. So on an affected install, **no agent that existed before that deploy ever starts a follow-up sequence**, indefinitely. It fails safe and silently, which is why nobody would report it as a bug: the operator sees follow-ups configured, enabled, and simply never firing, and there is nothing in the logs.

**Repro** (any install, no data needed):

```sql
-- as the runtime app role, with no tenant context
UPDATE "agents" SET "follow_up_armed_at" = NOW();   -- UPDATE 0

SET app.is_super_admin = 'on';
UPDATE "agents" SET "follow_up_armed_at" = NOW();   -- UPDATE <n>
RESET app.is_super_admin;
```

**Fix:** a follow-up migration that arms the agents the original backfill missed, wrapped in `SET app.is_super_admin = 'on'` / `RESET`, and idempotent (`WHERE "follow_up_armed_at" IS NULL`) so an install where the original backfill DID work is untouched.

**Status confirmed:** the mechanism, by executing the statement as a non-superuser role against a live database. **Not confirmed:** how many installs are affected. A self-hosted Postgres where the migration role is a real superuser is unaffected.

Same class of problem, already documented as part of #105: any DATA migration over a tenant-scoped table needs that GUC, and `docs/deploy.md` now says so.

PR #204, opened 2026-08-22
Let the agent hand the customer the quote it already rendered

### LINKED ISSUE #21: Quotes render a PDF but there is no way to deliver it to the customer
`POST /v1/quotes` generates a quote and renders its PDF, but nothing in the product can put that PDF in front of the customer. The generation half is built and the delivery half is missing, even though the primitive for delivery already exists elsewhere in the codebase.

### What exists

`src/modules/quotes/service.ts` generates idempotently — same `idempotencyKey` produces one quote and one PDF, "never N links to the lead" — and `src/lib/pdf.tsx` renders it with `@react-pdf/renderer`. Both are solid: the render is pure, never fetches a remote URL, and the storage path is derived from the row id so a retry overwrites instead of orphaning.

### What is missing

I read `src/api/v1/quotes.controller.ts` in full (174 lines). It exposes four routes:

```
GET   /v1/quotes/
POST  /v1/quotes/
POST  /v1/quotes/:id/revoke
GET   /v1/quotes/:id/pdf     <- requireAuth: true
```

There is no public or signed route — no occurrence of `signed`, `public`, `noAuth` or an access token anywhere in the file, and the whole controller is behind `tenancyPlugin`. That is the right default for a tenant-scoped file (the comment says it plainly: the filesystem has no RLS). But it means the only way to read the PDF is with a tenant credential, and the customer does not have one.

`getQuotePdf` has no consumer other than that controller and the tests. So a quote is generated, rendered and stored, and then stops there.

**This is not an authorization problem.** An HTTP tool could fetch the bytes server-side with a vault credential — but an HTTP tool's return value goes to the *model*, not to the customer. No credential creates the last step.

### The primitive is already there

`ctx.chatwoot.client.sendFileAttachment(conversationId, …)` in `src/modules/chatwoot/client.ts` sends a file to the customer as an attachment in the live conversation. Today it has exactly one consumer: the Google Drive toolpack (`drive_send_file`).

Drive cannot be used as a bridge, because that toolpack is read-only — `drive_find_file` and `drive_send_file`, with every Google call being a read (`/files/:id`, `?alt=media`, `/export`). There is no upload, so a generated quote cannot be placed in Drive and sent from there.

The Drive path also already settles the questions a quote-sending tool would raise: `MAX_DOWNLOAD_BYTES` is 15 MB, and a file above the ceiling is refused and offered as a link instead. A quote-delivery tool could inherit that contract as-is.

### Why it matters

For a sales agent, the quote is the step between "the customer asked for a price" and "the customer decides". Being able to generate the document but not hand it over means the conversation has to leave the product exactly at its most valuable moment.

A native tool that sends an already-generated quote as an attachment in the current conversation — taking the quote id, reusing `sendFileAttachment` and the existing byte ceiling. That keeps the PDF route authenticated and never exposes a public link.

If a delivery path already exists and I missed it, that is the more likely answer and I am happy to be pointed at it.

### Environment

gabrieljablonski:
## What was missing

`POST /v1/quotes` generates a quote, renders its PDF and stores it. The only way to read that file back is `GET /v1/quotes/:id/pdf`, which sits behind `tenancyPlugin` and `requireAuth` — the right default for a file the filesystem cannot scope, and a credential the customer will never hold. `getQuotePdf` had no consumer other than that route and its tests, so a quote was generated, rendered, stored, and stopped there.

The primitive for the last step already existed: `client.sendFileAttachment`, whose only caller was the Google Drive toolpack. What was missing was something to call it with a quote.

## What this adds

A native tool, `send_quote`, granted like any other. It **queues** the PDF on the turn's attachment queue and the runtime delivers it — after the ownership recheck, the supersede gate and the output guardrail, ahead of the reply text. Posting from inside the graph invocation would put a priced document in front of a customer whose turn is then discarded, which is the same reason `resolve_conversation` defers and `send_image` queues.

The queue was `pendingImages`; it is now `pendingAttachments`, carrying the tool that queued each file. One queue, because the gates a file passes to reach a customer are a property of the **turn**, not of what the file is — a second queue would be a second place to remember them. The flow line and the "nothing was delivered" error now name the queueing tool, so an operator debugging a PDF read off our own disk is not sent to the image host allowlist.

## The part of the issue this does not follow

The issue proposes a tool "taking the quote id". This does not, and the reason is a fence rather than ergonomics.

A quote is addressed to one lead. An id in a tool argument is **model-supplied**, and inside a tenant every other lead's quote is a valid id — so one hallucinated or injected number is somebody else's name and prices delivered to the wrong person. `send_quote` takes no arguments at all: it looks the row up by the conversation it is bound to, which the model has no way to influence.

That made an existing under-specification load-bearing, and settling it is most of this diff:

- `Quote.conversationId` holds a **Chatwoot** conversation id (that is what `{{conversation_id}}` interpolates into the HTTP tool an operator's backend is called from). A Chatwoot conversation id is unique per **account**, not per tenant. Every other conversation lookup in the tree keys on `(tenant, instance, conversation)` for that reason; this table was the outlier that could afford not to, because nothing read the column.
- So the row now also carries the resolved **thread key** (`tenant:instance:conversation`), in the `Quote.threadId` column that already existed and was never filled, resolved at generation from the caller's conversation id. Fail-closed: no mirrored conversation, or two of them, leaves it null and the tool declines rather than picking one. `conversationId` is untouched and stays as the caller wrote it.
- The `threadId` column already existed, so binding needs no schema change. Two migrations do come with this: an index on `(tenant_id, thread_id)`, because `send_quote` filters by thread on a customer-facing path that had only the tenant index behind it, and a **backfill** that binds the quotes generated before this existed — without it the agent answers "there is no quote for this conversation" about conversations that have one. The backfill applies the same fail-closed rule, and runs under `SET app.is_super_admin`: `quotes` and `conversations` both carry `FORCE ROW LEVEL SECURITY`, so a migration role that owns the tables without being a superuser would match zero rows and report success, which is how issue #106 lost a backfill in silence.
- **Retrying the same idempotency key heals an unbound quote.** `skipDuplicates` means a second call reaches the existing row and the resolution lands nowhere, so a quote generated before this path existed — or one whose first attempt ran while the conversation was not yet mirrored — would have stayed undeliverable forever. It binds on the retry instead, guarded on the conversation matching so reusing one key for a *different* conversation still returns the original quote without re-pointing it.

Two smaller decisions:

- **The thread's most recent quote, never "the most recent one that can go out".** A revised quote replaces the one before it; a fallback would attach the price the operator just replaced, silently, at the moment the customer is deciding on it. A newest quote that is revoked or expired is a refusal **with its reason** (`none` / `not_rendered` / `revoked` / `expired`), so the model has something true to say instead of promising a document.
- **No byte ceiling, and the image ceilings stay about images.** The issue suggests inheriting Drive's 15 MB cap. That cap exists because Drive fetches a file a model named from a host we do not control; a quote is our own render, read off our own disk. The turn's rule is one quote, and `SEND_IMAGE_MAX_PER_TURN` / `SEND_IMAGE_MAX_TURN_BYTES` keep counting only images — a turn that happens to carry a quote must not silently become "2 images per message".

**One review finding is waived, not fixed:** re-validating the quote in the database immediately before `sendFileAttachment`. The tool call is the moment the turn commits to the document, and the model writes the rest of its answer around it ("segue o orçamento"); a revocation landing in the seconds that follow would, with a re-check, drop the file and leave a message promising an attachment that never arrives. The one place the queue *is* dropped is the output guardrail, and there the reply text is replaced in the same breath, so the pair stays consistent. Same shape as the deferred resolve: the intent is decided when the tool runs.

The deliverability rule itself moved into a pure predicate (`src/modules/quotes/deliverable.ts`) shared with the HTTP route, which still collapses every reason into 404 on purpose. It returns a verdict carrying the storage key rather than a nullable answer, so neither caller has a re-check branch below it that no input can reach.

## Validation scope

**Proven by test** (`bun check` green: 3714 pass, 0 fail):

- End to end, through `runAgentTurn` against a real database: a model calls `send_quote`, the customer receives `sendFileAttachment` with `quote-<id>.pdf`, before the reply text, and the trail carries `{ tool: "send_quote", outcome: "sent" }`.
- The binding heals: a quote first generated unbound is bound by a retry of the same key, and reusing one key for another conversation does not re-point it.
- The backfill, by **reading the migration's own SQL** and running it as the RLS-bound app role: an unbound quote whose conversation is mirrored gets its thread, a conversation number two accounts share is left unbound, and re-running it rewrites nothing. Removing the `SET app.is_super_admin` line turns that suite red rather than turning the backfill into a silent no-op.
- The binding: a mirrored conversation gives the quote its thread; a conversation number two accounts of one tenant share leaves it unbound **and undeliverable to either**; an unmirrored conversation leaves it unbound; an explicit thread from the caller wins.
- The selection: the thread's own quote comes back with its bytes; a revoked newest quote refuses instead of falling back to the one it replaced; expired, still-rendering and missing-file are each named; another tenant reading the same thread key finds nothing.
- The tool: queues instead of sending mid-turn; declines with no turn (a proactive nudge) and with no thread; a batch that asks twice attaches one copy; a refusal reaches the model with what to do instead.
- The decision table for the predicate, including the three precedence pairs.
- That a quote riding along does not take an image's slot or its bytes.

**Mutation-tested.** Twelve mutations, all killed against the final HEAD: dropping the tool registration (6 tests, the e2e among them — this is the red-first proof), killing the thread resolution (9), widening the ambiguity guard to pick the first row (1), removing the `revoked` condition (5), reading the oldest quote instead of the newest (1), removing the post-await duplicate check (1), letting the quote count against the image ceilings (1), hardcoding `send_image` in the delivery flow line (1), removing the retry's healing update (1), dropping the conversation guard on that update (1), removing the backfill's RLS escape hatch (1), and dropping the backfill's ambiguity guard (1).

**Not validated:**

- No live run. There is no rendered quote in a real Chatwoot conversation behind this: the attachment path is the same `sendFileAttachment` the Drive toolpack and `send_image` already use in production, and the test asserts the arguments, not the wire.
- No model was asked to decide *when* to call it. The tool description is written and unmeasured; whether an agent reaches for it at the right moment in a real sales conversation is a prompt question this does not answer.
- The multi-account ambiguity is derived from the schema (`chatwoot_conversation_id` is unique per instance, and `@@unique([tenantId, accountId])` allows several per tenant), not observed in an install. What is measured is the fail-closed behaviour on a fixture built that way.
- Nothing here changes who may generate a quote. `POST /v1/quotes` is still `TENANT_ADMIN`, and the agent has no tool to create one.

Fixes #21


gabrieljablonski:
Superseded by the work for #207.

This PR did the last mile — a `send_quote` native tool that attaches the stored PDF to the reply — and reviewing the subsystem around it showed the first two miles are missing: no UI for quotes, no way for an agent to generate one (`generateQuote`'s only caller is the REST controller), a template compiled into `src/lib/pdf.tsx`, no custom fields, and three columns nothing writes. The tool would have answered "there is no quote for this conversation" in practically every installation. Round 3 also found an ambiguity this PR could not fix from inside it: a stored quote was matched on a Chatwoot conversation id, which repeats across a tenant's accounts.

None of the work is lost. The attachment queue this PR introduced — tools queue, the runtime delivers after the turn's ownership recheck, supersede gate and output guardrail — comes back generalized in the replacement, and the conversation-id ambiguity dissolves there because a document is bound to the conversation's thread key at issue time.

PR #220, opened 2026-08-23
Ask whether the agent is live before the redirect ladder chases a lead

### LINKED ISSUE #219: The channel-redirect follow-up ladder never asks whether the agent is enabled, so a switched-off agent keeps messaging the lead
The `REDIRECT_FOLLOWUP` job is careful about one kind of staleness and blind to the other. It reloads the redirect config fresh — "an operator may have changed or disabled it since this job was armed, and a scheduled delay can span that change. Never trust the arm-time snapshot" — but it loads the agent itself as `select: { settings: true }` (`src/modules/channel-redirect/followup.ts`), so `enabled` and `mode` never come back from the database. Nothing downstream asks for them either.

The ladder has three stages, and only the first is covered by accident:

- **chat** — goes through `runAgentNudge`, which loads `select: { mode: true }` and returns `"silenced"` for an unactivated test conversation (`src/graph/nudge.ts`). Test mode is handled here; `enabled` is not, so this stage still runs for a switched-off agent.
- **whatsapp** — `sendWhatsAppFollowUp` sends `cfg.waFollowupMessage` on the WhatsApp sibling as fixed text, or an **HSM template** when `proactiveSendMode` says the 24h window is closed. No agent gate at all.
- **closing** — `deliverRedirectClosing` posts the closing message on both channels and calls `toggleStatus(..., "resolved")`. No agent gate at all.

The generic follow-up cannot cover this path by construction: `isFollowUpLive` carries a `managedByRedirect` term (`src/modules/followups/eligibility.ts`) whose whole purpose is to keep the generic sweep out of redirect-owned conversations. And nothing cancels the armed jobs when an agent is disabled — I could not find a cancellation path keyed on `Agent.enabled` anywhere.

**What an operator sees.** They switch the agent off — the one control that is supposed to mean "stop" — and the redirect ladder keeps chasing leads: a WhatsApp message they did not authorise, possibly a paid template, and then both conversations resolved out from under them. Same for an agent still in `test` whose widget conversation was never activated with `/teste`: the mode's promise is silence, and two of the three stages break it.

**Suggested fix**, in the shape the module already uses for this question: a pure predicate next to the handler, `agentEnabled && !isTestSilenced(agentMode, testActivatedAt)`, evaluated once before the stage switch so all three stages are covered, returning `{ outcome: "done" }` — the same arm as `!cfg.enabled` directly above, which is right, because a fresh customer message re-arms the ladder from stage `"chat"` through the `dedupeKey` upsert.

PR opened alongside this issue.

One thing deliberately left out of that PR, since it is a second logical change: `runAgentNudge` gates on `isTestSilenced` but never on `Agent.enabled`, so *every* nudge caller — not just this one — will act for a disabled agent. Happy to open that separately if you would like it fixed at the nudge level instead.

JotaSXBR:
Closes #219.

The `REDIRECT_FOLLOWUP` handler reloaded the redirect config fresh but read the agent as `select: { settings: true }`, so `enabled` and `mode` never reached it. Stage `"chat"` is covered by accident (it goes through `runAgentNudge`, where the test-mode gate lives); the WhatsApp link re-send and the closing send fixed text and an HSM template with no agent gate at all, and `isFollowUpLive` cannot cover this path — its `managedByRedirect` term exists precisely to exclude these conversations.

**The change**

- `isRedirectFollowUpLive`, a pure predicate in the shape of `isFollowUpLive`: `agentEnabled && !isTestSilenced(agentMode, testActivatedAt)`.
- The handler's scoped read now returns the agent's `enabled`/`mode` and the widget conversation's `testActivatedAt` in one transaction, and the gate runs once before the stage switch, so all three stages are covered by a single check.
- Failing the gate returns `{ outcome: "done" }`, matching the `!cfg.enabled` arm right above it: a fresh customer message re-arms the ladder from stage `"chat"` through the `dedupeKey` upsert, so this drops a stale ladder without stranding a live lead.

**Tests**

Four cases on the predicate in `tests/modules/channel-redirect-followup.test.ts` — enabled production agent runs; disabled agent never runs, in either mode; a test agent is silent until the widget conversation has a `/teste` stamp; an activation stamp never revives a disabled agent. Pinned as a predicate for the same reason `isFollowUpLive` is (`tests/modules/followup-eligibility.test.ts`): the rule is the thing worth freezing, and the handler's collaborators (`loadChatwootClient`, `sendTemplate`, the scheduler) are not mockable from a unit test here.

**On `bun check`**

Lint, `tsc`, the worker type-check, the OpenAPI check and `i18n:extract` are clean. The test run is green for this change — 4080 pass against 4076 on a clean `main`, exactly the four added — but both runs report the same 17 pre-existing failures, all in `tests/api/v1/carveout.test.ts` and the MCP OAuth suites, all of them `TypeError: Expected Request object` raised from `elysia-rate-limit` at `src/api/middlewares/rateLimit.ts:46`. They only appear in a full-suite run; each file passes in isolation. Untouched by this branch, and I did not want to fold an unrelated fix in here — tell me if you would rather have it as its own issue.

I have read and agree to the CLA for this contribution and future ones.

### DIFF
--- a/src/modules/channel-redirect/followup.ts
+++ b/src/modules/channel-redirect/followup.ts
@@ -439,35 +439,6 @@ export async function redirectFollowUpHandler(
   if (!cfg.enabled) return { outcome: "done" };
   // Dropping the ladder (not rescheduling) matches the !cfg.enabled arm above: a fresh customer
   // message re-arms it from stage "chat" via the dedupeKey upsert.
-  const liveness = async (): Promise<boolean> => {
-    const now = await runScopedOn(base, sysCtx(tenantId), async (db) => {
-      const a = await db.agent.findUnique({
-        where: { id: agentId },
-        select: { enabled: true, mode: true },
-      });
-      if (!a) return null;
-      const c = await db.conversation.findUnique({
-        where: {
-          tenantId_chatwootInstanceId_chatwootConversationId: {
-            tenantId,
-            chatwootInstanceId: parsed.instanceId,
-            chatwootConversationId: parsed.conversationId,
-          },
-        },
-        select: { testActivatedAt: true },
-      });
-      return { a, testActivatedAt: c?.testActivatedAt ?? null };
-    });
-    // A read that fails is not a refusal — same reasoning as `retired`: an unknown answer must not
-    // silently drop work that was legitimately armed. A DELETED agent is an answer, and it is no.
-    return now === null
-      ? false
-      : isRedirectFollowUpLive({
-          agentEnabled: now.a.enabled,
-          agentMode: now.a.mode,
-          testActivatedAt: now.testActivatedAt,
-        });
-  };
   if (
     !isRedirectFollowUpLive({
       agentEnabled: agent.enabled,
@@ -539,12 +510,7 @@ export async function redirectFollowUpHandler(
     if (await retired()) return { outcome: "done" };
     if (cfg.waFollowupEnabled && entryInboxId !== null) {
       const outcome = await sendWhatsAppFollowUp({
-        // The switch is asked again HERE, not only at the top: the link mint is an HTTP round trip
-        // to Chatwoot, so the answer above is older than the send, and an operator switching the
-        // agent off is most likely to do it while it is chasing somebody. The chat stage needs no
-        // such re-ask — `runAgentNudge` reloads the agent config, which is fail-closed on `enabled`.
-        stillWanted: async () =>
-          (await retired()) === false && (await liveness()),
+        stillWanted: async () => !(await retired()),
         tenantId,
         instanceId: parsed.instanceId,
         agentId,
@@ -577,10 +543,7 @@ export async function redirectFollowUpHandler(
   if (await retired()) return { outcome: "done" };
   if (cfg.closingEnabled && entryInboxId !== null) {
     await deliverRedirectClosing({
-      // Same window as the WhatsApp stage above, and this one messages BOTH conversations and
-      // resolves them.
-      stillWanted: async () =>
-        (await retired()) === false && (await liveness()),
+      stillWanted: async () => !(await retired()),
       tenantId,
       instanceId: parsed.instanceId,
       widgetConversationId: parsed.conversationId,
--- a/src/modules/channel-redirect/followup.ts
+++ b/src/modules/channel-redirect/followup.ts
@@ -4,6 +4,7 @@ import basePrisma from "@/api/lib/prisma";
 import { type AgentNudge, parseThreadId, runAgentNudge } from "@/graph/nudge";
 import type { RuntimeDeps } from "@/graph/runtime";
 import { runScopedOn, type ScopedDb, type TenantContext } from "@/lib/tenancy";
+import { isTestSilenced } from "@/modules/agents/test-mode";
 import { loadAgentBot, loadChatwootClient } from "@/modules/chatwoot/instance";
 import {
   type ObservedConversation,
@@ -99,6 +100,22 @@ export async function retireRedirectFollowUp(
     followUpDedupeKey(widgetThreadId),
     base,
   );
+}
+
+export interface RedirectFollowUpLiveness {
+  // Agent.enabled — an operator switched the agent off after the ladder was armed.
+  agentEnabled: boolean;
+  // Agent.mode + the WIDGET conversation's testActivatedAt: a test agent is silent until /teste.
+  agentMode: string;
+  testActivatedAt: Date | null;
+}
+
+// The ladder's own liveness gate. The generic follow-up cannot cover this path by construction —
+// its managedByRedirect term excludes exactly these conversations — and two of the three stages
+// send FIXED text (the WhatsApp link re-send, the closing) without ever passing through
+// runAgentNudge, where the test-mode gate lives. So the ladder asks here, for EVERY stage.
+export function isRedirectFollowUpLive(s: RedirectFollowUpLiveness): boolean {
+  return s.agentEnabled && !isTestSilenced(s.agentMode, s.testActivatedAt);
 }
 
 export type RedirectFollowUpStage = "chat" | "whatsapp" | "closing";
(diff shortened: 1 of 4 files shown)


gabrieljablonski:
Reviewed, and the change is right: the ladder is the correct place for the gate, the predicate covers all three stages because it runs before the stage switch, and dropping the job rather than rescheduling matches the `!cfg.enabled` arm above it. I checked the two ungated stages against the code — `sendWhatsAppFollowUp` and `deliverRedirectClosing` reach `client.sendTemplate` / `client.sendMessage` with no agent read anywhere on the path — so a switched-off agent really was re-sending links and burning HSM templates.

I pushed two things to the branch rather than sending it back for another round trip, since this is a `priority: high` customer-facing send.

**1. Rebased onto `main`.** #202 (`/reset` reaching the conversation it exists to rescue) landed in both files after you opened this. The conflicts were adjacency only — an import line and a block that grew next to yours — and your commit is unchanged apart from that. `tsc`, lint, the worker check, OpenAPI and `i18n:extract` are clean, and the full suite is **4753 pass / 0 fail** here, including the 17 failures you reported: they do not reproduce on this machine, so that looks environmental rather than something on `main`.

**2. Added two tests, at the handler.** This is the one thing the PR was missing, and it is worth stating precisely because the tests you wrote are good ones. I deleted the handler's entire call to `isRedirectFollowUpLive` and **all 34 tests stayed green**. A decision table proves what the predicate decides; it cannot prove that the code being fixed asks it — and the stages that had no gate at all are exactly the ones worth pinning.

The DB-backed harness in the same file already drives `redirectFollowUpHandler` with a stub client and an HTTP double, so the two cases cost very little: a switched-off agent re-sends no WhatsApp link (asserted on the wire being empty — not even the link mint, which is a Chatwoot round trip made before the send) and posts no closing, resolving neither conversation. Both assert `outcome: "done"` for the reason your PR body gives. Each dies under both mutations: dropping the handler's call, and making the predicate return `true`.

**One question, not a change request.** The test-mode half reads `testActivatedAt` from the **widget** conversation, while stages 2 and 3 message the **entry** (WhatsApp) one. Keying it to the widget is defensible — the ladder itself is keyed by the widget thread — but it means a test agent activated in the chat still sends fixed text on WhatsApp, where the lead never typed `/teste`. The `enabled` half, which is what #219 is about, is unaffected either way. Worth your view: if the two-conversation episode should carry one activation, that is its own issue and I would rather file it than widen this PR.

Thanks for this one — the write-up did the diagnosis work, and the correction on #230's framing came from it too.

JotaSXBR:
Thanks — the rebase and the two handler tests are both better than what I had. The mutation you ran is the honest test of my suite and it failed it: a decision table proves the predicate, never that the ladder consults it. Point taken well beyond this PR.

**On your question — I went and measured it, because I did not want to answer it from reading.** It is not a keying preference, and the answer is not "read the other row": the widget key is wrong in *both* directions. Three throwaway probes on your branch (`e39aab8`), on the DB-backed harness already in that file, then removed — the file is back to 36/36.

`shouldPropagateTestMode` runs in **one** direction, **once**, **only at link time**:

```ts
mode === "test" && siblingTestActivatedAt !== null && widgetTestActivatedAt === null
```

WhatsApp → widget, and `linkRedirect` stamps `redirectLinkedAt` unconditionally in the same update, so it is a one-shot. `/teste` itself writes only to `ctx.conv.id`, the conversation it was typed in. There is no other writer of `testActivatedAt`.

**Probe 1 — false positive.** Agent `mode: "test"`, `/teste` in the website chat only (widget stamped, entry `null`), stage 2:

```
POST https://.../accounts/11/conversations/7172/messages
```

7172 is the WhatsApp conversation. A real send, on a lead who never typed `/teste`.

**Probe 2 — the closing, same setup.** Stage 3 with that same chat-only activation:

```
sent:     [[7171,"Vamos encerrar por aqui."],[7172,"Vamos encerrar por aqui."]]
resolved: [7171,7172]
```

Both conversations messaged, both resolved. This is the part that rules out any single-row key: whichever conversation you read, one of the two destinations is being judged by state that is not its own.

**Probe 3 — false negative.** `/teste` on WhatsApp *after* the link (entry stamped, widget `null`) — the one-shot has already run, so nothing propagates. Stage 2 sends **zero** messages. The ladder goes mute on an episode that *is* activated, on the very channel it would have messaged.

So the correct shape is an **episode-level activation** — one unit spanning the linked pair — **evaluated per destination at send time**, not a row read once at handler entry. Stage 2 adds a wrinkle on top: `resolveWhatsAppSibling` takes the most recent contact conversation in the entry inbox, which need not be this episode's thread at all.

That is schema-shaped and touches `/teste`, `linkRedirect` and both stages — well outside #219, and I would not want it riding on a `priority: high` send fix. Yours to file, and the three probes above are yours to reuse as the reproduction; say the word if you would rather I open it with this as the body. Either way this PR is done from my side.

gabrieljablonski:
Follow-up to the review rounds. Four in total, and the summary of what changed since your commit:

**Round 1 — two findings, both real.**

1. The resolve-triggered closing is the *other* way this goodbye reaches a customer. `webhook.ts` turns a resolve on the widget conversation into a closing on the WhatsApp sibling — fixed text, no nudge behind it — so nothing on that path asked the switch. The runtime it already loads (`inboxAgentRuntime`) carries `enabled` and `mode`; they were simply never read. Your predicate is now called there too, with the widget conversation's stamp. The ladder cancel above it stays ungated on purpose: standing the chase down is not a send, and a switched-off agent wants it stopped either way.
2. The gate going stale across the ladder's own I/O — the link mint is an HTTP round trip, which is exactly the window your file already calls "the last free place to stop" for the retirement question.

**Rounds 2 and 3 — why (2) was taken back out.** I first composed the liveness re-ask into the existing `stillWanted` fences. The next round returned three findings, all consequences of that one change, and both mechanisms it collided with turned out to be load-bearing:

- `sendWhatsAppFollowUp` reads a false `stillWanted` as **retired**, and the handler still reschedules the closing on it (`rescheduleTo` re-asks `jobRetired` and nothing else). Under a retirement that is right; under a liveness refusal the ladder *advances*, and if the agent comes back before the delay expires the old episode closes both conversations with no fresh inbound behind it.
- `jobRetired` swallows a failed read deliberately — an unknown answer must not drop work that was legitimately armed. A liveness read composed into the same callback did not, so a read that throws after `deliverRedirectClosing` claimed the at-most-once watermark skips the release branch and the goodbye is lost permanently.

One bit, two questions with different consequences. What that window actually needs is a refusal **distinct** from retirement — one the ladder ends on rather than advances past, and one that fails open on an unreadable answer — which is a change to the ladder's advance semantics rather than a term added to an existing fence. So it is out of this PR and filed as #246, with the two failure modes written down.

Round 3 then caught the gate's own cost: the activation lookup I added ran for every agent, on a path whose failure is permanent (the best-effort catch sits after the ladder cancel, the delivery is marked `PROCESSED`, and a conversation resolves once). Only a test agent's liveness depends on that stamp, so only a test agent pays the read now.

**Round 4 came back clean.** CI is green on `e1a27f7` — lint, type-check and the full test suite.

**What is in the PR now:** your gate and predicate, unchanged apart from the rebase; the same question asked at the resolve-triggered closing; and five tests that pin the call sites rather than only the predicate — a switched-off agent re-sending no WhatsApp link and posting no closing, the resolve path with the agent off and with a test agent nobody activated, plus the control that shows the same path does deliver when the agent is on. Each was measured against a mutation that removes what it pins.

Your `testActivatedAt` question from my earlier comment still stands as a question, and #246 is not it — that one is about the answer going stale, not about which conversation's stamp is the right one to ask.

gabrieljablonski:
Merged, and thank you — this closes #219.

Your measurement is filed as #249, with the three probes as the reproduction and the `shouldPropagateTestMode` one-shot written down as the reason a single row cannot answer for both destinations. I credited it to you there rather than having you re-open it, since you had already done the work; if you would rather own the fix, say so on the issue and it is yours. I also carried over the `resolveWhatsAppSibling` wrinkle you flagged, which turns out to be the same pairing problem as #222.

One thing recorded in #249 for whoever picks it up: the false negative (`/teste` on WhatsApp after the link → the ladder goes mute) is *introduced* by this PR, and it fails toward silence on a test-mode agent, which is why it did not hold up a `priority: high` fix for a switched-off agent still messaging real leads. The false positive predates us both.

The freshness window is #246, and it is still unclaimed.

Issue #203, opened 2026-08-22
The in-flight turn fence is process-local, so a scaled web tier can lose a memory append


gabrieljablonski:
Raised by review on #201, and left out of it because it is not that PR's mechanism: it belongs to `src/graph/inflight.ts` and to all three of its consumers.

## What happens

`inflight.ts` is an in-memory `Map` of the turns currently executing. Three places ask it whether a turn owns a memory thread before writing to that thread:

- **memory compaction** (`src/modules/memory/compact.ts`), which rewrites the message channel;
- **the proactive follow-up nudge** (`src/graph/nudge.ts`), which must not fire in the middle of a long turn;
- **continuous ingestion** (`src/graph/ingest.ts`), since #194, which defers instead of appending.

A LangGraph invoke is a read-modify-write of the whole message channel: it saves the state it loaded plus its own messages, erasing anything that landed meanwhile. The fence is what makes the three writers exclusive with a turn, and the module says which invariant it holds under:

> Safe under the single-replica / one-leader invariant: the webhook turn and the scheduler worker share this process, so they share this Set.

`docs/deploy.md` §4 sanctions a topology where that is not true. To scale the web tier you run extra replicas with the workers off and keep one leader with them on. Turns then run on whichever replica the Chatwoot webhook lands on, while the scheduler runs on the leader — different processes, different `Map`s. Every consumer above reads `false` on a thread that is in fact busy.

## What it costs, per consumer

- **Compaction**: the rewrite is undone by the turn that saves after it. Recoverable, and the module already says so — the next attendance boundary re-arms it and the summary row already exists.
- **The nudge**: it races one reply. Visible, not durable.
- **Ingestion**: the append is undone *and* the thread's row records the message as ingested (`recentSyncedMessageIds`). The message is gone and marked handled, which is exactly the loss #194 exists to close. It is the first consumer whose cross-process failure is irreversible.

Two things narrow it. Ingestion only arms for messages no turn covers (silenced out of hours, a human owns the conversation, or an attendant's own reply), so the collision needs a *different* concurrent turn on the same contact inbox. And the residual is not new: before #194 the inline append consulted no fence at all and lost the message on every topology, single replica included. #194 closes it for the documented invariant and leaves this.

## Why it is its own issue

Making the fence durable is a design decision for the module, not a guard to add at one call site:

[...]

PR #227, opened 2026-08-24
Stop the thread's critical section from holding a transaction across a second pool

### LINKED ISSUE #225: Produção: turnos morrem (timeout transação/memória) + resolve-on-failure + nota pública de erro
## Resumo
Na produção (Ponto do Celular), conversas novas ficam sem resposta do bot por um encadeamento de 3 problemas: (1) 404 do fork do Chatwoot por divergência `id` vs `display_id` ao postar a resposta; (2) lifecycle do turno (resolve-on-failure + gate que só age em `pending`) → turno silencioso com nota pública de erro; (3) timeouts de transação Prisma (5 s) sob pressão de memória/swap da VPS.

## 1. Postagem de resposta usa id incompatível com a API do fork
O espelho guarda `chatwoot_conversation_id` = id da tabela (vem do webhook) e o POST de resposta vai para `/conversations/{id}/messages` com esse mesmo valor. A API do fork faz deploy que resolve por `display_id` (sequência por conta, defasada do id) → **404** para conversas novas; funciona por coincidência quando `id == display_id`. Impacto real: clientes sem resposta; log: `chatwoot: failed-turn note not posted (conv=1320): Chatwoot API 404 for POST /conversations/1320/messages`. Ver issue no fork (fazer-ai/chatwoot). Sugestão: mapear `display_id` a partir do payload/listas da API (ou esperar o fix no fork e apontar o POST por display_id).

## 2. Lifecycle do turno — resolve-on-failure + gate pending-only + nota pública
- Conversa 1320 foi setada `status=pending, bot=3` às 11:38Z e voltou sozinha para `status=resolved` (re-resolve após falha) — o bot (gate `shouldBotHandle`, webhook.ts ~L1406) só age com `pending`, então o turno seguinte fica mudo ("bot silent by gate").
- A cada falha o fork injeta nota **pública** (message_type=2): "A conversa foi marcada como aberta pelo sistema devido a um erro com o robô" — o cliente enxerga a falha.
- Perguntas: resolve-on-failure é intencional? Em falha de infra (timeout/transação) a política não deveria ser manter `pending` + retry com backoff? A nota de erro deveria ser privada?

## 3. Timeouts de transação Prisma sob pressão de memória
Erros reais (logs do container agents, 24/08):
```
Unable to start a transaction in the given time
query cannot be executed on an expired transaction (5000ms, passaram 9613ms)
```
- VPS: swap 86% cheia (1,7/2 GB), ~1,9 GB livres → swap-thrash → Postgres demora a abrir transação → turno morre antes de gravar (conversa do incidente: ZERO linhas em `execution_logs`).
- Banco individual rápido (SELECT 1 em 0,31 s) — gargalo é pool/transação sob thrash, não query.
- Suspeita: eventos de trace gigantes (17–37 MB com imagem/áudio) sobrecarregam ClickHouse/minio e inflam a memória (build novo trunca eventos oversized, mas a fonte do problema continua).

## Sugestões
- Retry no ingest/execução com backoff quando o erro for timeout/transação (P2028 / expired transaction).
- Truncar payloads de trace COM mídia na fonte, antes do ClickHouse.
- Reavaliar política de resolve-on-failure (manter pending em falha de infra) e nota pública de erro.

## Contexto
- Ambiente: agents-pro (deploy via compose/Coolify), Postgres pg17 com RLS (MIGRATION_DATABASE_URL p/ operação), Langfuse self-hosted, VPS 12 GB RAM.
- Mitigação operacional: `drop_caches` deu alívio temporário; workaround de resposta por `display_id` aplicado manualmente hoje (não escala).

gabrieljablonski:
Fixes #225

## What broke

Two Postgres pools, both sized `dbPoolMax`: Prisma's and the LangGraph checkpointer's. In five places the `ingest:<threadId>` critical section ran **inside** a Prisma transaction, so a connection from pool A sat `idle in transaction` for the length of two or three round trips to pool B, one of them rewriting the whole thread channel, while also holding a `pg_advisory_xact_lock`. Under load pool A drained and every unrelated query in the process waited out `maxWait` (2s) and failed. The Chatwoot webhook ack is one of those queries, and a late ack costs the conversation its bot.

`multi-tenant.ts` already warned "no network/LLM await inside fn", and the invariant was respected to the letter: no transaction in the repo calls Chatwoot, an LLM, or `fetch`. What escaped is that a checkpointer call **is** that await, to a second Postgres.

## What changed

The five sites take a process-local queue (`withKeyedQueue`) instead of a transaction-scoped advisory lock, with short transactions inside. `runScopedOn`/`asSuperAdminOn` also pass explicit transaction options rather than inheriting Prisma's defaults, which were the two numbers in the original report.

Removing the enclosing transaction moved a question that used to be free. The retirement probe inside the critical section borrowed that transaction's live connection; on its own it can now exhaust `maxWait`, and `jobRetired` swallows an unreadable row as "still wanted". That guess is the expensive one **there**: the run recreates the graph state `/reset` just cleared, the operator is told the conversation was wiped, and the agent keeps answering from it, with no later fence to catch it. `jobRetiredStrict` propagates the read failure instead, sending the job back through the scheduler's own bounded retry.

And `stillWanted` turned out to answer **two** questions. Only the ask inside the critical section runs before anything is written; the rest guard a send, where throwing abandons the bookkeeping of a message the customer already has (the follow-up skips its watermark and job completion and the scheduler resends; the redirect closing leaves `redirectClosedAt` set so the retry loses its claim and the terminal messages are dropped for good). The callback now takes which question is being asked.

## Risk that has to stay written down

The old lock was `pg_advisory_xact_lock`, **cross-process**. `withKeyedQueue` is process-local. Under the deploy contract (`docs/deploy.md` §4, one replica or one leader with the workers) this regresses nothing: `isTurnInFlight` is already an in-memory `Map`. But there was margin **beyond** the contract, and this spends it: in a scaled tier the advisory lock still serialized the `AgentThread` read-and-write and the channel append across replicas. Issue #203 tracks the process-local claim failing there; this widens its scope from the claim to the whole critical section, and #203 has been commented to say so. The path back to cross-process safety is the CAS watermark, deliberately not in this change.

## Validation scope

**Proven by test.** A new `pool-inversion.test.ts` counts Prisma transactions open at the moment the checkpointer is touched and asserts the count is zero; on the original code all four samples report one. The existing dedupe and attendance-boundary suites are the gate on behaviour being unchanged. `scheduler.test.ts` pins that the strict probe propagates an unreadable read where the lenient one reports "not retired", and that both agree when the read succeeds. `contact-auth-nudge.test.ts` had asserted the OLD invariant (the probe arrives WITH a connection); it now asserts the one that replaces it, that no ask finds a transaction open, and that exactly one ask is strict and it is not the entry one.

The strictness rule is per **call site**, so a source walk polices it: one strict ask per file that owns a critical section, none in the file whose every ask surrounds a send, and the strict one inside `withKeyedQueue` rather than merely somewhere in the file. A table-driven test of the callback would prove the callback and say nothing about which of the eleven asks passes `strict: true`, and the one that matters is reached only through `runLoadedTurn`, which the webhook path never uses (`stillWanted: null`, having no job to retire). Full suite 4817 pass / 0 fail.

**Not validated.** No production measurement. Nobody has confirmed which call site dominated the original incident, or that the ack was in fact what exhausted the pool; `pg_stat_activity` filtered on `idle in transaction` during an incident would answer it and has not been run. The concurrency the queue provides is asserted structurally, not under real parallel load: a timing test here would pass on a fast machine for the wrong reason. And the cross-process regression above is reasoned from the deploy contract, not measured against a scaled deployment.

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -45,6 +45,10 @@ Destructive migrations (enum changes, `ACCESS EXCLUSIVE` on `users`) want a main
 
 The realtime pub/sub, the scheduler tick, the debounce worker, the outbound-webhook worker, and the alert-delivery worker are single-process by construction (in-process state + reentrancy guards). **Run one replica.** To scale the web tier: run the extra replicas with `SCHEDULER_WORKER_ENABLED=false`, `WEBHOOK_WORKER_ENABLED=false`, `DEBOUNCE_WORKER_ENABLED=false`, and `ALERT_WORKER_ENABLED=false` (workers off), and keep exactly one "leader" replica with them on. Scaling with workers on every replica causes double-fires and lost cross-replica realtime events (a durable claim / leader election / Redis pub/sub bridge is the post-MVP path).
 
+**One more thing now depends on the single process.** The `ingest:<threadId>` critical section (held by continuous ingestion, a reactive turn, the proactive nudge, memory compaction and `/reset`, all on the same key) is serialized by an in-process queue (`withKeyedQueue`), not by the Postgres advisory lock it used before. That lock had to go: those sections span the LangGraph checkpointer, which is a **separate** connection pool, and holding a Prisma transaction open across it drained the main pool and made unrelated queries fail on `maxWait` (issue #225). The in-flight turn registry (`src/graph/inflight.ts`) was already in-process, so the invariant is not new; its scope is wider. On a genuinely scaled web tier two replicas can now interleave a thread's channel read-modify-write, which is issue #203.
+
+`/reset` is the one member that still holds a transaction across the checkpointer, deliberately: there the rollback is what keeps a failed checkpoint delete from leaving an operator told the memory was cleared while the thread still answers from it (`src/modules/memory/reset.ts`).
+
 ### 5. CSP / build
 
 Production serves `dist/index.html`; its inline-script hashes are baked into the CSP at build time. Rebuild (`bun run build`) after editing any inline script. Per-tenant theming uses `setProperty` (DOM API, no inline `<style>`) precisely so it does not break the CSP hash.
--- a/src/graph/ingest-job.ts
+++ b/src/graph/ingest-job.ts
@@ -2,10 +2,15 @@ import type { BaseCheckpointSaver } from "@langchain/langgraph-checkpoint";
 import type { PrismaClient } from "@/../generated/prisma/client";
 import { decryptJson, encryptJson } from "@/api/lib/crypto";
 import logger from "@/api/lib/logger";
+import { runScopedOn, type TenantContext } from "@/lib/tenancy";
 import { armCompaction } from "@/modules/memory/compact";
 import { type ClaimedJob, enqueueJob } from "@/modules/scheduler/service";
 import { type JobResult, registerJobHandler } from "@/modules/scheduler/worker";
 import { type IngestRole, ingestMessageIntoThread } from "./ingest";
+
+function sysCtx(tenantId: bigint): TenantContext {
+  return { tenantId, userId: null, role: "TENANT_ADMIN" };
+}
 
 // Continuous ingestion as a scheduler job, instead of an append made inline while the webhook is
 // being acked (issue #194).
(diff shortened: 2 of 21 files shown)


matheusvitorino28:
@release-ping-227.md

PR #266, opened 2026-08-25
Check an error's translation key against the catalog that has to answer it

### LINKED ISSUE #256: Two tool-grant refusals are never translated: errors.invalidToolGrant names a locale entry that does not exist
## What happens

Two refusals on the tool-grant path name a locale entry that does not exist:

```ts
throw new AppError(`${field} is required`, 400, "errors.invalidToolGrant", undefined, field);
throw new AppError(`${field} must be a numeric id`, 400, "errors.invalidToolGrant", undefined, field);
```

`errors.invalidToolGrant` is in neither `src/api/locales/en.json` nor `src/api/locales/pt-BR.json`.
`translateWithLocale` falls back to `message`, so the operator gets the untranslated English string
whatever they asked for. Measured through the real app:

```
accept-language: en     400 {"error":"agentId is required","field":"agentId"}
accept-language: pt-BR  400 {"error":"agentId is required","field":"agentId"}
```

Its neighbour on the same path answers correctly, which is what makes the gap visible rather than
merely theoretical:

```
accept-language: pt-BR  400 {"error":"O texto em guardrails.output.templateMessage é longo demais: 5000 caracteres (limite 2000).","field":"guardrails.output.templateMessage"}
```

## Why the entry is missing

API locale entries are registered by a magic comment beside the route and materialised by
`bun i18n:extract` (`docs/i18n.md`): `src/api/v1/agents.controller.ts` carries a
`// translate('errors.settingsTextTooLong', …)` line and roughly thirty siblings, and
`errors.invalidToolGrant` never got one. The extractor prunes what nothing references, so adding the
entry to the JSON by hand does not survive the next run.

## Worth deciding before writing the entry

The key is used at **twelve** sites in `src/modules/agents/service.ts`, for different problems: a
missing id, a non-numeric id, an unknown tool name, a grant naming something the tenant does not own.
One entry for twelve refusals can only be a generic sentence, which throws away what each of them
says in English today. So the question is whether this is one key or several — and if several, which
of them the operator can actually act on.

Found while working on #231.

gabrieljablonski:
`errors.invalidToolGrant` names a locale entry that does not exist, so a tool-grant refusal answers in English whatever the caller asked for.

```
accept-language: en     400 {"error":"agentId is required","field":"agentId"}
accept-language: pt-BR  400 {"error":"agentId is required","field":"agentId"}
```

## The issue undercounts, twice

It reports two refusals. That key is thrown at **twelve** sites in `src/modules/agents/service.ts`, and sweeping the family found the same state everywhere:

| | |
| --- | --- |
| `errors.*` keys named in `src/` | 152 |
| registered in `src/api/locales/en.json` | 104 |
| **missing** | **48 keys, 112 call sites, 21 files** |

About a third of the localized refusals in this API were never localized. `src/modules/chatwoot/management.ts` was missing 9 keys, `agents/service.ts` 8, `mcp-connections/service.ts` 7.

## Why nothing caught it

i18next answers an unresolvable key with `defaultValue ?? key`, and `onError` passes `message` as the default. A missing entry does not throw, does not log, and renders a plausible English sentence: it is invisible at runtime by construction.

The `// translate(...)` ledgers that feed `bun i18n:extract` are maintained by hand, and nothing compares them to what the code throws. `agents.controller.ts` **had** a ledger of 13 entries while `agents/service.ts` threw 8 keys that were not in it, so this is not a controller that was forgotten. It is a ledger with no check.

## The key is typed, not checked later

```ts
type ApiLocale = typeof import("@/api/locales/en.json");
export type ErrorTranslationKey = `errors.${keyof ApiLocale["errors"] & string}`;
```

`typeof import(...)` is a type query, erased at compile time, so `src/lib/errors.ts` keeps the "no imports, no cycles" property its header claims. `AppError`, its subclasses, `translate` and `translateWithLocale` take that type instead of `string`.

Applying it turned the class into **113 compile errors**, which is the red proof this change is measured against. It is now 0.

The type also reached two sites a source sweep cannot see: a key travelling inside a `Refusal` struct (`documents/templates.ts`), and a key built by interpolation, where it named all three strings the producer can emit.

A Biome plugin was considered and cannot do this: GritQL has no cross-file knowledge, so a plugin can see the literal but never the catalog. `docs/i18n.md` already records that limit.

## What the type cannot see, and the defect that proves it

`tests/api/error-catalog.test.ts` covers the three blind spots: an `as` cast, a key used as a **comparison token** rather than an argument, and a catalog that holds the key but answers pt-BR with the English sentence.

The middle one was live. `resolveEmbeddingConfig` threw `errors.embedding.<snake_case>`; the RAG ingest catch stores that `translationKey` verbatim in `KnowledgeDocument.error` (deliberately, so the console can localize it); the console matched `errors.embeddingCamelCase`:

```ts
// producer, src/modules/rag/documents.ts
`errors.embedding.${embeddingBlock(status).reason}`   // errors.embedding.credential_empty
// consumer, useKnowledgeManager.tsx
if (error === "errors.embeddingEmpty")                // never equal
```

Two spellings, no call site in common, so neither branch had ever fired and the operator read the raw token off a tooltip. `credential_pending` had no branch on either side. Both ends now read one map (`src/lib/embedding-block.ts`) and a test asserts the console covers it **entry for entry**, so a reason added later fails a test instead of shipping a token to a tooltip.

## One key for twelve refusals could only be a generic sentence

Split by what the caller has to change, each carrying that value as an interpolation param:

| Key | The question it answers | Sites |
| --- | --- | --- |
| `toolGrantIdRequired` | an id that was not sent | 1 |
| `toolGrantIdInvalid` | an id that is not a number | 1 |
| `toolGrantDuplicate` | a source granted twice | 6 |
| `toolGrantUnknownTool` | a tool name that does not exist | 2 |
| `toolGrantUnknownSource` | a source that does not exist | 1 |
| `toolGrantToolNotInIntegration` | a tool that exists, but not for that integration | 1 |

`tests/modules/tool-grant-refusal.test.ts` asserts each **through the translation**, not on the error object: the masking happened in that step, so an assertion on `message` passes with or without the fix. Its negative case is the design decision, that the six must not answer with the same sentence.

## Red proof

| | |
| --- | --- |
| `bun run type-check` before | 113 errors |
| `bun run type-check` after | 0 |
| suite, this repo | 5305 tests, 0 fail |
| suite, derived Free tree | 5206 tests, 0 fail |
| suite, derived Pro tree | 5204 tests, 0 fail |
| Mutations | 57, all break a test (0 survivors) |

The battery is one script, so the number above is re-derivable in one run rather than assembled from
notes. Six mutations survived a first pass and each named something real. Four were guards with no offender in live
data — the cast detector, the plural relaxation, the stale-waiver check and the wire sweep's status
check all asserted over catalogs that do not offend, so blinding the predicate changed no answer.
Each is now a function proven against synthetic input that DOES offend. The fifth was a fixture whose
expectation was built from the same variable as its input, so blanking that variable moved both sides
together; it is asserted on a character code now. The sixth was mine: I ran the mutation against the
wrong test file. A seventh appeared when the battery was consolidated and re-run against the final
code: the per-token render assertions look up the sentence they expect IN the catalog, so swapping a
pt-BR entry for its English text moved the expectation with it. The control that does not move — the
two languages must differ on screen — now covers all three tokens instead of one.

## The English catalog was holding Portuguese

Found by the same sweep, on the other side of the line. Eight client entries read identically in both
languages because `en.json` held the pt-BR text, so an English-speaking operator read the knowledge
screen in Portuguese:

```
knowledge.addContent        'Adicionar conteúdo'
knowledge.addContentTitle   'Adicionar conteúdo em {{name}}'
knowledge.addContentAction  'Adicionar'
knowledge.clearFile         'Limpar arquivo'
knowledge.dropHint          'Arraste e solte arquivos aqui, ou clique para escolher'
knowledge.docStatus.READY   '{{n}} trechos'
knowledge.tabArquivo        'Arquivo'
knowledge.tabTexto          'Texto'
```

The defaults in `t(...)` were Portuguese and the extractor copied them, so fixing the JSON alone
would not have held: it preserves an existing value, which is why those entries had to be deleted and
regenerated rather than edited. Four of the eight were **pinned by existing tests** querying the
knowledge screen by its Portuguese labels (`getByRole("button", { name: /adicionar/i })`).

An earlier version of this PR caught only the first four, with a rule that required three-plus prose
words in the entry — on the argument that a shorter exception list would be a ledger nobody keeps.
The other four are one and two words long, and they shipped past it. The threshold is gone: the rule
is now every client entry that reads the same in both languages, minus **an enumerated list of 102**,
each grouped with the reason it is legitimately identical (brand names, acronyms, OAuth-console
vocabulary, words spelled the same in both). A hundred lines of data anyone can check, against four
user-visible defects a threshold hid. A second test refuses a waiver whose key was fixed or deleted,
so the list cannot become a place where an argument outlives what it argued about.

## A registered key can still say nothing

Round 1 of the review found three keys registered with `{{placeholders}}` whose throw sites passed no `translationParams`, so the translated sentence rendered the placeholder literally while the English fallback stayed correct. That is the **inverse** of the defect this PR is about: there a missing entry fell back to a correct English sentence, here the entry exists and interpolates nothing, so registering it made the answer worse. Both are invisible at runtime.

The type cannot close this one, and that was probed rather than assumed: `keyof` survives JSON module widening, so the KEY is checkable, but TypeScript widens JSON string VALUES to `string`, so no template-literal type can extract `{{x}}` from a catalog entry.

So the class is closed at runtime instead. `translateWithLocale` falls back to the already-interpolated `message` when the catalog template names a placeholder the caller did not supply. Information beats language: a pt-BR caller reading an English sentence that names the value can act on it, and one reading `{{timezone}}` cannot.

That guard caused a regression in its first form, caught by `bun check`. It tested the RENDERED string for `{{`, and an interpolated value can legitimately hold braces — a document-template refusal quotes the token it rejected (`token "{{cliente}}" names no field...`), so a correct pt-BR sentence read as broken and fell back to English. It now reads the catalog TEMPLATE and compares against the params supplied, and that negative case is pinned by a test.

## A key that cannot say what its call site says

Round 3 of the review found the sharpest instance of the same inverse defect, on a key **this PR
registered**. `refusalBody` prefers the catalog sentence over `AppError.message`, so registering a key
REPLACES the message — and where the message was the more informative of the two, registering it made
the answer worse, in English as well as in pt-BR:

```
Google Drive denied the request. Reconnect the credential granting the 'Drive (read-only)'
or 'Drive (full access)' scope (the 'Drive (app files)' scope cannot list existing folders).
                                    ↓ errors.upstream
The integration provider refused or failed the request.
```

Sweeping the family found sixteen call sites behind five keys this PR had registered. `errors.badRequest`
covered two different facts ("not a connected Google account" / "invalid credential reference"),
`errors.upstream` covered three, `errors.chatwootConfirmMismatch` could not say WHICH of two
confirmations the operator mistyped, and `sttFailed`, `invalidModelConfigDetail` and the three
business-hours refusals dropped an interpolated value on the floor. Each fact now has its own key, and
what varies rides as a param.

The rule that generalizes it is mechanical, and it is now a test. A key whose catalog entry has no
`{{placeholder}}` cannot carry a value its message interpolates, nor a second, different literal
message. Twenty-five keys already in `main` are in that shape; they are listed as **grandfathered**,
with a test asserting the list may only shrink and that nothing on it was introduced here.

**And it has a limit, which round 5 found by finding two more.** Where the message is simply the more
specific prose — one call site, one literal, nothing interpolated — no mechanical rule separates it
from a paraphrase. `errors.invalidModelConfig` answered "The model configuration is not valid" where
the message said `modelConfig must be an object`, and `errors.chatwootDifferentDeployment` dropped
"disconnect it first to switch servers", which was the only recovery step in a 409. Reading the whole
family afterwards turned up a third, `errors.chatwootAccountTaken`, which lost the sentence explaining
why one account cannot be shared. The three are fixed and **pinned by wording** in a test, because
what makes them right is not derivable from the code.

## A subclass is a throw site with no arguments

Found by pointing `curl` at a running server, which is the one thing none of the sweeps above does.

```
POST /api/v1/integrations/inbound/<bad-token>   accept-language: pt-BR
{"error":"Unauthorized"}
```

`UnauthorizedError` and `TenantTargetRequiredError` hard-code their own message and status and pass no
key, so there is no `"errors.X"` literal for a source sweep to find and nothing for the type to check
— while `errors.unauthorized` and `errors.tenantTargetRequired` sat in both catalogs, translated, with
nothing pointing at them. **Twenty-eight call sites** answered English to a pt-BR caller. Both classes
now carry their key, and a test enumerates every `AppError` subclass whose `super(...)` passes none.
One is waived, with the argument already written at the class: a 503 whose body the client never shows,
because it retries.

## The column keeps the old spelling

Round 6, and the last of the family. `KnowledgeDocument.error` is a STORED column and nothing rewrites
it when the producer changes, so every row that failed on an older release still carries
`errors.embedding.<snake_case>`. Those rows show the raw token today — that IS the defect this PR
opens with, and it is why they were never readable — but the fix as written only helped rows failing
from now on.

The console now maps the three old spellings onto today's tokens, derived from `EMBEDDING_BLOCK_KEY`
so a reason added later cannot get a modern token and be forgotten. No data migration: the column is
one the app rebuilds on re-index anyway, and a migration over it would be a write with nothing to gain.

The waiver this needed in the literal sweep is itself checked: every exempt token must still be
reached, and reached ONLY from the console's alias map — a legacy spelling reappearing in the producer
would mean the rename was undone, and the waiver would then be hiding the bug it was written around.

## The gate now depends on a step that ran after it

Round 7, and it is about the workflow this change creates rather than what it answers.
`ErrorTranslationKey` is `keyof` the GENERATED `src/api/locales/en.json`, so adding a key is: write
the `// translate(...)` line, run `bun i18n:extract`, and only then does the throw site type-check.
`bun check` ran `build-check` four steps BEFORE the extractor, so it reported every newly declared key
as unknown at a point where nothing said why. Measured on a probe key: 1 type error before extracting,
0 after.

CI already had the right order; only the local script disagreed, which is the worst version of it —
the two differed about what a clean tree is. The extractor now runs first in `check`, the dependency
is written in `docs/i18n.md` where someone adding a key will read it, and a test pins both orders.

CI's freshness check was also diffing only `src/client/locales/`. The API catalog is the one the type
reads, so it could drift with nothing failing; the diff now covers both.

## Validation scope

**Proven by test:** the type refuses an unregistered key (113→0); every one of the **168** catalog keys
driven through the app's real `onError`, in both locales, over `app.handle` — asserting none answers
the untranslated fallback, the bare key, or an unfilled placeholder, and that no two locales answer
the same sentence; the knowledge screen's failure reason **rendered in the DOM** in both languages,
from the real catalogs, through the real component; the six tool-grant refusals and the business-hours
id refusal, each through `translateWithLocale`; the embedding map covered entry for entry, injectively;
catalog parity in both directions; en/pt placeholder agreement.

**Proven on the derived Free tree**, not only here: the whole suite was run against `derive:free`
output, which is what caught two waivers that name Pro-only keys and would have turned the public CI
red.

**Proven against a running instance**, on a throwaway database, with a real session:

| | |
| --- | --- |
| `errors.unauthorized` (inbound receptor) | English before, `Não autorizado` after — this is how the subclass family was found |
| `errors.tenantTargetRequired` | `A target tenant is required` / `É necessário informar um tenant de destino` |
| `errors.emptyVaultSecret` | `A vault secret must not be empty.` / `Um segredo do cofre não pode ficar vazio.` |
| tool-grant refusals, 5 of 6 | each with its own sentence and its params interpolated, in both languages |
| `errors.invalidRequestValue` (#264) | `{{field}}` interpolated on the wire, both languages |

The sixth tool-grant refusal, `toolGrantUnknownSource`, is unreachable over HTTP: the source is a
closed enum in the route schema, so the request is refused a layer earlier. It is covered by the type
and by the decision table.

**Proven in a real browser** (Playwright, both languages), on the knowledge screen:

| document state | pt-BR | en |
| --- | --- | --- |
| `errors.embeddingEmpty` | `A credencial de embedding está vazia…` | `The embedding credential is empty…` |
| `errors.embedding.credential_pending` (the pre-rename spelling) | `A credencial de embedding ainda não foi preenchida…` | `The embedding credential has not been filled in yet…` |
| a raw provider diagnostic | `connect ECONNREFUSED 10.0.0.4:443` verbatim | same |

The four English labels that were holding Portuguese render as `Add`, `Text`, `File` and `12 chunks`.

One thing the live run established that no test could: a configuration block does **not** stamp the
document row. `resolveEmbeddingConfig` refuses at the reindex gate, before anything is queued, so the
token reaches `KnowledgeDocument.error` only in the window where the credential changes mid-job. The
legacy rows round 6 is about come from that narrow window, and from releases where the gate was not
there.

**Not validated:** the pt-BR wording of the 56 new sentences was written and audited against the
project's own rules (no calques, diacritics, brand spelling) by one person; nobody else read it. The pt-BR wording of the new sentences was written and audited by one person against
the project's own style rules (no calques, correct diacritics, brand spelling); nobody else read it.
`toolGrantToolNotInIntegration` is the one refusal reached only inside the scoped transaction, so its
key is covered by the type and by the catalog test, not by the decision table. Two entries,
`errors.currentPasswordIncorrect` and `errors.passwordChangeUnavailable`, have English text that
predates this PR and is missing the final period its pt-BR side carries; the pt-BR side was aligned to
match rather than editing an auth controller this PR does not otherwise touch.

Fixes #256


gabrieljablonski:
### Review round 1 — `codex` `gpt-5.6-sol` effort=high — 3 findings, all P2

Posted as a plain comment: the reviewing bot account's token comes from a 1Password mount that is not
yielding right now, and `ALLOW_HUMAN_FALLBACK` would attribute a bot review to a human. Recorded here
before fixing, so the round has a record that predates its own fix.

All three are the same defect and all three are correct. I registered catalog entries carrying
`{{placeholders}}` but never passed `translationParams` at the throw site, so the translated sentence
renders the placeholder literally while the untranslated `message` fallback still reads correctly:

| Key | Throw site | Missing param |
| --- | --- | --- |
| `errors.integrationNoInboundWebhook` | `rotateInboundRouteToken()` | `integration` |
| `errors.invalidTimezone` | `validateTimezone()` | `timezone` |
| `errors.mcpLauncherInvalid` | MCP stdio launcher check | `launchers` |

This is the exact inverse of the defect the PR is about. There, a key with no catalog entry fell back
to English and the caller still read a correct sentence; here the entry exists and interpolates
nothing, so registering it made the answer worse than before. Both are invisible at runtime.

Fixing the class rather than the three lines: every catalog entry whose text contains `{{x}}` obliges
every throw site of that key to pass `x`. That is checkable, so it becomes a test rather than three
edits.

gabrieljablonski:
### Review round 2 — `codex` `gpt-5.6-sol` effort=high — 0 findings, `patch is correct`

Head reviewed: `097fa737a6ef923dfb6dc431693c69c378f63beb`. Posted as a plain comment for the same reason as round 1: the bot
account's token comes from a 1Password mount that is not yielding, and `ALLOW_HUMAN_FALLBACK` would
attribute a bot review to a human. That means this PR carries no `APPROVE` review object, so branch
protection still reports "an approving review is required" — the merge uses `--admin`, as usual here.

Round 1's three findings were one class, and fixing the class rather than the three lines turned up
two things worth recording:

**The type could not close this one.** The key is typed against the catalog because `keyof` survives
JSON module widening. The TEXT does not: TypeScript widens JSON string values to `string`, so a
template-literal type cannot extract `{{x}}` from a catalog entry. Probed before assuming.

**My first fail-safe caused a regression, and `bun check` caught it.** It tested the RENDERED string
for `{{`, but an interpolated value can legitimately hold braces: a document-template refusal quotes
the token it rejected (`token "{{cliente}}" names no field...`), so a correct pt-BR sentence read as
broken and fell back to English. It now reads the catalog TEMPLATE and compares against the params
supplied. That negative case is pinned by a test.

One mutation was informative rather than green: with the fail-safe in place, deleting a param no
longer breaks anything visible — the answer becomes correct ENGLISH instead of braces. So the
language of those three sites was untested until the structural ledger went in, which also refuses a
placeholder key thrown through `NotFoundError`/`ForbiddenError`, whose constructors carry no params
argument at all.

Battery: 18 mutations, all break a test, baseline 0.

PR #285, opened 2026-08-25
Date the team's answer, so an inbox the agent never touched still has a service level

### LINKED ISSUE #283: Every quality KPI is computed from `llm_usage`, so an inbox served by humans reads as zero traffic instead of as traffic the agent did not touch
## What breaks

The dashboard answers "how is the agent doing?" and cannot answer "how is this inbox doing?". Those are the same question on the day the agent is off — during a rollout, after a handoff-heavy week, or on an inbox whose bot was never enabled — and the answer it gives then is not "no data", it is **zero**, which reads as failure.

`getKpis` opens with a `Conversation.count`, and then every other number is derived from the set of conversations that produced `llm_usage` rows (`src/modules/analytics/service.ts:251`, `distinct conversationId`, `source: "inbox"`). `involved` is the size of that set, and `resolutionRate` / `automationRate` divide by it. `getTimeseries` is `FROM llm_usage` outright (`:352`). Of the whole analytics module, exactly two queries read `Conversation`: the status roll-up at `:163` and the outcome lookup at `:270`, and the second is already keyed by `involvedIds`.

So the dashboard has one number that survives the agent being off — the conversation count — and it is the one number that says nothing about how the attendance went.

## Measured

Against a real Postgres and the real receiver (`processChatwootDelivery`), on an agent with `enabled: false`: one conversation, one customer question, one human reply typed by an operator (`message_type: "outgoing"`, `sender.type: "user"`), one second apart.

```
totalConversations: 1
involved:           0
involvementRate:    0
resolutionRate:     0
automationRate:     0
```

That is correct as written — no LLM ran — and useless as displayed: an operator answered a customer in one second and the panel reports a flat zero across the board.

The mirror row after that reply, every timestamped field it carries:

```
lastInboundAt        = 14:31:02   ← the customer's question
lastEventAt          = 14:31:04   ← moved by the human reply
lastHandledMessageId = 6001
assigneeId           = 31
assigneeName         = "Ana (atendente)"
```

## The gap is narrower than it looks

The data is not missing from the pipeline — it is **discarded at the mirror**.

`mirrorChatwootEvent` runs unconditionally on every event, before any gate (`webhook.ts:2556`). The event that carries a human reply is already recognised by name: `isNewHumanAgentMessage` (`normalize.ts:299`), added for #187, correctly excluding private notes, reactions, and other bots. And the mirror already computes exactly this shape for the other side — `inboundAt` at `mirror.ts:105`, written to `lastInboundAt`, deliberately advanced only on a brand-new incoming message.

There is no counterpart. `grep` for `humanReply|firstResponse|responseTime|lastOutbound` over `prisma/schema.prisma` returns nothing. The one field that did move, `lastEventAt`, is an **activity** watermark: a status change, an assignment, or the customer writing again all move it, and it carries no author. `14:31:04 − 14:31:02` happens to be the response time in the probe above and is not a response time in general, which is why reading it as one would be worse than not reporting it.

## What I think this is

Two columns on the mirror, written where `lastInboundAt` already is, gated on the predicate that already exists:

- `firstHumanReplyAt` — set once per inbound episode; `firstHumanReplyAt − lastInboundAt` is a first-response time that means what it says.
- `lastHumanReplyAt` — the human counterpart of `lastInboundAt`.

And KPIs derived from those, alongside the LLM ones rather than replacing them: first-response time, conversations answered by a human, volume per assignee. `assigneeName` is already mirrored, so per-person is available — I would default the reported cut to **aggregate**, with per-person opt-in: in the jurisdictions this ships to, a per-operator performance number is a labour-law and LGPD question, not just a chart.

Cost: zero. No model call, no new event subscription, no extra query on the hot path — two columns on a write that already happens.

## Why it is worth doing on its own

I raised the "watch without answering" mode in #209 and this issue does **not** depend on it. It is worth having with the agent fully on: today a conversation the agent hands off leaves the humans' half of it unmeasured, so "handoff" is a terminal state in the funnel rather than the start of the part a supervisor actually wants to see. The columns are the same either way.

The probe used to measure this is offline and deterministic; happy to bring it as the test if the shape is welcome.

JotaSXBR:
Closes #283.

The per-assignee cut that issue also asked for is now #365, on its own, because it is the only one of the three that needs a decision before it needs code (see **Scope**).

> Opened by @JotaSXBR, who did the diagnosis, the measurement and the case for it. The mechanism below is the second one this branch carried: the first was a watermark state machine of our own, and six review rounds took it apart. The history is in the thread.

## The problem

`getKpis` opens with a `Conversation.count`, and every other number is derived from the set of conversations that produced `llm_usage` rows. `getTimeseries` is `FROM llm_usage` outright. So on an inbox served by humans, during a rollout, after a handoff-heavy week, on an inbox whose bot was never enabled, the panel does not read "the agent did not touch this". It reads **zero**, which reads as failure.

Measured through the real receiver, agent `enabled: false`, one customer question answered by an operator one second later:

```
totalConversations: 1
involved / involvementRate / resolutionRate / automationRate: 0
```

## What changed, and why it is not what this branch started as

The first version of this PR computed a first-response watermark on our side, from the events we receive. Rounds 2 through 6 each found a different ordering it got wrong: a retry delivering the reply before the question, a conversation the mirror first meets mid-dialogue, `/reset`, a business-initiated conversation whose opener was taken as the answer. Five design changes in six rounds is the signal that the model was wrong rather than incomplete.

**Chatwoot already computes this, from its messages table, and ships it on every conversation payload.** `Conversations::EventDataPresenter#push_data` carries `created_at` and `first_reply_created_at`; `Message#valid_first_reply?` is the same predicate this codebase spells `isNewHumanAgentMessage` (outgoing, not private, not a reaction, sender outside `AgentBot` / `Captain::Assistant`). Both are set once at the source and never revised.

That collapses every one of those findings at once. A value computed from a messages table does not depend on the order webhooks arrive in, is already correct for a conversation that predates our mirror, and makes a late retry carry exactly the readings the delivery it duplicates carried, so the write is idempotent whatever order it lands in.

So: **two mirrored columns, no state machine.**

| column | source | written |
| --- | --- | --- |
| `chatwootCreatedAt` | `conversation.created_at` | on every event that names a conversation |
| `chatwootFirstReplyAt` | `conversation.first_reply_created_at` | same |

Both nullable, both never wiped by a payload that says nothing, and both written even by a delivery the ordering decision refuses as stale. Losing the ordering says something about the conversation's *state*, and nothing about two immutable readings the source computed.

And one KPI reading them: `firstResponseSeconds`, the **median** over the conversations carrying both, beside `firstResponseSampled`. Median and not mean because one conversation opened on a Friday night and answered on Monday moves a mean by hours and says nothing about the week; the sample count beside it because a median over four conversations is not a service level.

## The panel

The number reached the API and stopped there, which does not answer an issue about the dashboard. It now has a section of its own beside the funnel, not a fifth card inside it, because the funnel is about the agent and this is about the team.

**Absence renders as absence.** With no sample the card shows an em dash and says why underneath. A first-response time that printed `0 s` for "nobody has answered yet" would be the same lie this issue exists to stop, one layer down.

Which unit a duration is read in is a rule, so it lives in `src/client/lib/duration.ts` as a pure function with a decision table, not inline in the render: seconds until two minutes, minutes until two hours, then hours and days with one decimal. The thresholds are where the number stops being useful in the smaller unit, not powers of sixty.

## The semantics, stated rather than inherited

On a conversation the **business** opened, `first_reply_created_at` marks that opening message, because that is what the field means to Chatwoot's own SLA. This PR keeps that reading deliberately: the number then matches the one the operator already sees on their Chatwoot dashboard, which is worth more than a private definition of ours. Timing the customer's wait instead would mean anchoring on `waiting_since`, which is a different metric and a product decision. It is said in the schema, in the type, and in the panel itself.

## Tests

`tests/modules/chatwoot-first-response-sla.test.ts`: DB-backed, driven through the real receiver (`processChatwootDelivery`), on an agent with `enabled: false` so nothing can reach the agent pipeline. **Every fixture is a `message_created` or a `conversation_updated`, and nothing in the file constructs a `conversation_created`**: `AgentBotListener` never dispatches one to a bot, so a rule needing it would pass under a fixture and report nothing in production. It covers the two spellings, a row born mid-dialogue, a delivery refused as stale, a payload that says nothing, the median with its sample, and a reply dated before its conversation.

`tests/client/lib/duration.test.ts`: the unit rule as a table: every threshold on both sides, the precision the formatter applies, and the four inputs that are not durations.

`tests/client/pages/DashboardFirstResponse.test.tsx`: renders the real page and asserts the **figure**, not the label above it. A card whose heading is right and whose value is a stale zero passes a label-only test and fails the issue.

### What review found here, after the redesign

Round 7, two findings, both real:

- **An epoch a `Date` cannot hold reached the column.** `Number.isFinite` passes for `1e20` and so does `> 0`, so what came out was an Invalid Date, which Prisma refuses: `PrismaClientValidationError` on the conversation write, measured. That fails the WHOLE delivery over an optional field, and fails it identically on every retry because the payload never changes. The guard now sits on the one exit every branch leaves through, since it is one question asked of three spellings. The ISO branch already had it and had no test; it does now, and each of the three falls to its own mutation.
- **A test read the sample the test above it had seeded**, so running it by name reported `sampled: 0` and failed while the service was correct. The fixture moved to setup.

Round 8, one finding, and it is the same defect as the issue one layer down: the card was built so absence would not read as zero, and its caption then claimed *"no conversation in this period has been answered yet"*. An empty sample proves only that no mirrored pair is available. Right after this ships, and indefinitely for a conversation closed before it that receives no further event, both columns stay NULL although a person did answer. The caption now says there is no data yet, and the note beside it says when a conversation joins the count.

Round 9 came back clean.

### Mutation battery

| mutation | tests that fall |
| --- | --- |
| unit threshold 120 s → 60 s | 3 |
| unit threshold 2 h → 1 h | 1 |
| negative-interval guard removed | 1 |
| NaN/Infinity guard removed | 2 |
| hours precision 1 → 0 decimals | 2 |
| formatter ignores `fractionDigits` | 1 |
| card prints the raw number (absence → `0`) | 4 |
| card prints the sample instead of the median | 4 |
| hint always reads "no sample" | 1 |
| caption claims "nobody has answered yet" again | 1 |
| unreadable-timestamp guard becomes the identity | 1 |
| the same guard dropped from the numeric branch | 1 |
| ...from the all-digits branch | 1 |
| ...from the ISO branch | 1 |

The hours-precision row is why this table is here: `fractionDigits` survived its first mutation, because the decision table could not see a field that only acts inside the formatter. The formatter rows exist to close that.

## Validation scope

**Proven by test.** Everything above, plus `bun check` green on the merged base: `6169 pass / 0 fail`.

**Proven against the vendor's source.** The two spellings the fixtures use are read off the fork this ships against, not assumed: `push_timestamps` renders `created_at` as `created_at.to_i` (epoch seconds), while `first_reply_created_at` is a plain ActiveRecord `datetime` attribute and serializes as an ISO-8601 string. `message.webhook_data` nests `conversation.webhook_data`, so both reach a bot through `message_created` as well as through the four conversation events, which is what makes reading them from either shape correct rather than lucky. `Message#valid_first_reply?` and the `update!(first_reply_created_at:, waiting_since: nil)` beside it confirm the field is set once and never revised.

**NOT validated.** No live Chatwoot delivered a webhook to this build; the payload shapes are read from the fork's source and from a real `message_created` captured earlier on this branch, not re-captured against this HEAD. The panel is asserted through a rendered page in tests, not in a browser. And nothing here measures how many existing rows will carry the pair after deploy: a conversation whose first reply already happened and which receives no further event stays NULL until it does, which is exactly what `firstResponseSampled` is on the response to make visible rather than assume away.

## Scope

#283 asked for three things: the measurement, the panel that reads it, and a per-assignee cut. The first two are here. The third is **#365**, split off rather than carried along, because it is not a leftover: the issue's own author flags it as a labour-law and data-protection question in the jurisdictions this ships to, wanting aggregate by default with per-person opt-in, and what it needs first is a decision about who may turn it on and what the cut may carry, not code.

Nothing here has to change for that to arrive later, and nothing here has to be undone if the answer is "we do not ship it": this PR is aggregate by construction.


gabrieljablonski:
@JotaSXBR — I owe you a long comment, because I pushed four commits onto this branch and I now think the design they built is the wrong one. The measurement is worth having; the mechanism should not be ours.

## What I pushed, and why

Six review rounds ran on this branch. Rounds 1 and 2 found two real defects in the original shape:

- a conversation that already carried customer traffic when the columns were added was **not** excluded, contrary to what the migration said — its next message anchored it, putting a mid-conversation interval into the median;
- an out-of-order delivery lost the watermark entirely, because `decideConversationWrites` refuses a stale event for the *state* it carries and the early return took the message with it.

Those became one function with the orderings as a table (`attendance-watermarks.ts`), called from all three mirror branches. Then round 3 found business-initiated conversations (an operator writes first, so the opener was taken as the first response and, the column being set-once, every real answer afterwards was too late to replace it). Round 4 found the same class in a different ordering. Round 5 found a third. Round 6 found two more, and your `attendanceTrackedFromStart` commit is a fair answer to one of them.

Five design changes in six rounds is the signal I should have read three rounds earlier: the model was wrong, not incomplete.

## What I found when I finally looked at the platform instead of the next finding

Chatwoot already computes this, and already ships it in the webhook payload. From a real `message_created` captured off an Agent Bot webhook on a local instance:

```
conversation.created_at             = 1787677147
conversation.first_reply_created_at = 2026-08-25T16:59:29.341Z
conversation.waiting_since          = 0
```

The customer wrote at `16:59:07` and the operator answered at `16:59:29.341` — 22.3 seconds, which is the same number the five columns on this branch produce.

`Message#valid_first_reply?` (`app/models/message.rb`) is the predicate this PR reimplemented as `isNewHumanAgentMessage`: outgoing, not private, not a reaction, `sender_type` outside `['AgentBot', 'Captain::Assistant']`. `waiting_since` is set on an incoming message and cleared on an outgoing response (`update_waiting_since`), and both fields are on every payload via `Conversations::EventDataPresenter`.

The decisive part: Chatwoot computes them **from the messages table**, not from webhook arrival order. Every finding rounds 2 through 6 produced — retry ordering, `/reset`, a conversation that predates the integration — is a problem for a state machine fed by webhooks and is not a problem for a mirrored field.

## The premise under the last commit does not hold

`attendanceTrackedFromStart` is `true` only when the mirror creates the row from a `conversation_created` event. An Agent Bot never receives that event. `AgentBotListener` implements exactly seven handlers, in upstream Chatwoot and in the fork alike:

```
conversation_resolved  conversation_opened  conversation_status_changed
conversation_updated   message_created      message_updated  webwidget_triggered
```

Measured rather than assumed: across 14 webhooks the bot received during a live run on a local instance, **7 `message_created`, 5 `conversation_updated`, 2 `message_updated`, and zero `conversation_created`**.

So in production the bit stays `false` forever, no conversation is ever anchored, and the KPI reports `sampled: 0` permanently. The suite is green because the fixtures construct the event — `{ event: "conversation_created", ...convPayload() }`, five times in `human-reply-watermarks.test.ts`. That is a test passing for the wrong reason, and CI cannot see it.

I want to be clear that this is not a criticism of the commit: it is the right instinct (durable provenance that `/reset` cannot forge) applied to a signal that does not arrive. The four commits I pushed before it are the same mistake at greater length.

## What survives

The half of this PR that was right from the first commit stands, and it is the half that took the real work:

- the diagnosis — `getKpis` derives everything from `llm_usage` (`analytics/service.ts`), so an inbox served by humans reads as zero rather than as traffic the agent did not touch;
- `firstResponseSeconds` beside `firstResponseSampled` on the response, and the argument for the median over the mean;
- the case for measuring the humans' half of an attendance at all, which is what makes a handoff something other than a terminal state in the funnel.

What goes is the mechanism: five columns, the decision module, and the ordering tests that exist only because the state machine had orderings to get wrong.

One thing to decide rather than inherit: on a business-initiated conversation Chatwoot's `first_reply_created_at` marks the opener, because that is its SLA semantics. If the number we want is "how long the customer waited", the pair with `waiting_since` answers it — but that is a product call, not a translation.

## Attribution

I am leaving this PR **open** rather than superseding it with a new one. The rework belongs here, on your branch, so that whatever eventually merges carries your authorship — the diagnosis, the measurement and the case for it are yours, and the merge should say so.

JotaSXBR:
@gabrieljablonski i will work on it after 17:00. Running out of tokens right now. LOL.

JotaSXBR:
@gabrieljablonski — agreed, and reworked. Pushed as `1066544`, a fast-forward on `38e289d`: your four commits are untouched, no force-push.

**−1254 / +524.** Out: `attendance-watermarks.ts`, the five columns, and the ordering suite that existed only because a state machine had orderings to get wrong. In: `conversation.created_at` and `conversation.first_reply_created_at`, read in `normalizeChatwootEvent` and written from all three mirror branches. `firstResponseSeconds` / `firstResponseSampled` keep the shape they already had on the response.

## On `attendanceTrackedFromStart` — confirmed from this side, and it is worse than "the event rarely arrives"

Your 14-webhook measurement is about Chatwoot. The corroboration in *this* repo is exhaustive rather than statistical: `conversation_created` appears **once in all of `src/`** — as a literal inside `CHATWOOT_HANDLED_EVENTS` (`chatwoot/types.ts:25`) — and that constant is **imported by no file**. There is no branch on that event name anywhere in `webhook.ts`. So the bit could not have been set even if `AgentBotListener` did dispatch the event. Nothing in the new suite constructs it.

## The mirrored pair does not need the guarantees the watermarks needed

Both values are computed at the source from the messages table and neither is revised once set, so every delivery mentioning a conversation carries the same two readings. That collapses three of the six rounds at once:

- **out-of-order retry** — the write is idempotent under any order. The stale branch writes them too, deliberately: the ordering that delivery lost is about the conversation's *state*, and these are not state this side maintains. It compares against what is stored first, so the ordinary stale delivery still adds no `UPDATE`.
- **rows older than the columns** — no exclusion and no backfill. Their next event carries their real numbers.
- **`/reset`, an integration installed over live traffic** — nothing local to forge or clear.

## The product call, decided rather than inherited

I took **parity with Chatwoot**: `first_reply_created_at − created_at`, the number the operator already reads on their own dashboard. A conversation the business opened therefore counts its opening message as the reply, and I would rather report that bias than have the console disagree with Chatwoot about the same inbox. It is stated in the column comment, in the KPI comment and in the commit message.

Timing the customer's wait is the better number for a service level, and I did look at anchoring on `waiting_since` — but Chatwoot clears it when the reply goes out, so capturing it means catching the first webhook that carries it, which puts arrival order back into the design we just took it out of. If you want that metric, it is its own issue and its own decision; say the word and I will open it.

## Verification

`bun run lint`, `build-check`, `openapi:check` clean. The new suite is six tests, and I checked it fails for the right reasons rather than trusting it: removing the create-branch write fails 5 of 6, and removing only the stale-branch write fails exactly the stale test.

Full suite here: 5345 pass / 17 fail — the same 17 fail on `origin/main` (5339 / 17) and all 17 pass when their two files are run on their own, on both. Local cross-file pollution in the MCP/OpenAPI/SPA routing tests, not this change; CI has been green on them.

Thank you for leaving this on the branch, and for writing the reasoning out instead of just superseding it. The five-design-changes-in-six-rounds reading is the part I will take to the next one.

JotaSXBR:
@gabrieljablonski i will leave this PR for you to finish for now. Very busy day at the office, be developing and directing a company is hard sometimes.

gabrieljablonski:
> @gabrieljablonski i will leave this PR for you to finish for now. Very busy day at the office, be developing and directing a company is hard sometimes.

👍

Issue #118, opened 2026-08-18
Tool argument descriptions carry operator-only text into every model turn


gabrieljablonski:
A toolpack tool's argument description is written once and consumed by two audiences with opposite needs: the model, which reads it on every turn the tool is bound, and the operator, who reads it in the console's tool list. `CALENDAR_ID_DESC` in `src/modules/integrations/toolpacks/google-calendar.ts` currently carries a sentence that exists only for the second one:

> Which calendar to act on: name or id of one of the calendars in `<allowed_calendars>`. **This arg only appears when the integration allows several calendars; with a single one it is used automatically.**

The bolded half is tautological for the model. `calendarArgSchema` removes `calendarId` from the schema whenever the instance has exactly one calendar, so the only context in which a model ever reads that sentence is the one where it is already false to worry about. It is tokens spent, every turn, to tell the model a condition that is guaranteed by the fact that it can read it at all.

For the operator it is genuinely useful, and the comment above the constant says so. `getToolpackToolViews` builds the console's argument list from the **static** schema, keyed by `catalogType` rather than by instance, so the console shows `calendarId` even for an integration with a single calendar. Without that sentence an operator has no way to know why their agent never receives the argument they can see documented.

**Why it is worth moving rather than deleting.** The frontend already has the right home: `src/client/lib/toolpackTools.ts` (`toolpackToolMeta`) holds per-tool labels and descriptions, keyed by tool name, and it is translated. Operator-facing text belongs there, where it is i18n'd and costs nothing at inference. The zod `.describe()` should carry only what the model needs to decide the call.

**Scope.** `CALENDAR_ID_DESC` is shared by six tools (`calendar_list_events`, `calendar_create_event`, `calendar_update_event`, `calendar_cancel_event`, `calendar_confirm_appointment`, and until recently `calendar_check_availability`). The same question is worth asking of the other toolpack descriptions while in there: anything that explains *the console* rather than *the call* is in the wrong file.

**Note on measurement.** This changes text a model reads on every turn, so it falls under the project's rule for prompt-facing edits. The change is a deletion rather than a rewrite, which lowers the risk, but the tools involved include `calendar_create_event` and `calendar_cancel_event`, where a wrong or missing `calendarId` has a real cost. Worth an A/B on whether the model still resolves the argument correctly with the sentence gone, rather than assuming a deletion is free.

[...]

PR #310, opened 2026-08-25
fix(google-calendar): a pinned slot grid is not the model's to change

rronaldovasconcelos-star:
## The bug

`calendar_check_availability` took `slotDurationMinutes` and `granularityMinutes` as optional args, and the resolvers let them win over the integration's config:

```ts
// src/modules/integrations/toolpacks/google-calendar.ts
return clampMinutes(override ?? config.slotGranularityMinutes, ...)
```

So whatever the model sent overrode what the operator configured. The arg description even suggested the value that causes the damage:

> Spacing between candidate start times, in minutes. Optional; defaults to the integration's setting (e.g. 15 ⇒ 09:00 and 09:15 are both offered).

## How it showed up in production

A school runs 1-hour visits on the half hour — `slotDurationMinutes: 60`, `slotGranularityMinutes: 30`. The agent offered a family:

> **14:00** · **14:15** · **14:30**

The first read was that the model had hallucinated `14:15`. It hadn't: it had sent `granularityMinutes: 15`, and the tool did exactly as told. The slot was real and bookable. A family would have shown up at 14:15 to a calendar built on one-hour blocks.

Across three runs the model sent 30 sometimes and 15 other times. That is the shape of the problem — the business rule lived somewhere the model could redefine per call, so it holds most of the time and quietly breaks the rest.

## The change

The two args are not the same kind of setting, so they do not get the same treatment.

**The spacing is the operator's, always.** `granularityMinutes` leaves the schema outright. The console has no "Let the AI choose" for spacing — only for the length — so a config *without* `slotGranularityMinutes` is an operator who never made the choice, not one who delegated it. Removing the arg only when a value happens to be pinned would leave the reported failure open for every integration created through the API: the model sends `granularityMinutes: 5` and gets a 5-minute grid over a 60-minute service.

**The length stays conditional**, via `slotDurationArgSchema` — pinned ⇒ the arg is removed, "Let the AI choose" ⇒ it stays and the model picks per request. That is the clinic case, where appointment length legitimately varies. It mirrors `calendarArgSchema`, which already does exactly this for `calendarId`, and sits next to it.

Dropping the arg from the schema is what fixes this; a description saying "don't do this" is still a request. No handler guard is needed alongside it: zod strips a residual key from an older tool definition before the body runs, which is why `calendarArgSchema` never needed one either. The regression test is what proves the strip.

The tool description stops suggesting a 15-minute grid, says the length is fixed when `preset="true"`, and tells the model to offer the returned start times exactly as they come back.

Per issue #118, the "when is this argument offered" half of the story lives in the console's operator note rather than in the model-facing `.describe()`: the model can only read that sentence in the case where it is already true, while the console — which lists args by `catalogType`, not by instance — is the audience that otherwise cannot tell why the agent never receives an argument it can see documented.

## Tests

Five regression tests, all failing against the previous behavior:

1. config pinned at 60/30, model sends `granularityMinutes: 15` → slots must be `14:00 · 14:30 · 15:00`. Before, five extra `:15` and `:45` slots came back.
2. **no** spacing configured, model sends `granularityMinutes: 5` → the runtime default (15) wins anyway. This is the case a conditional removal would have missed.
3. `granularityMinutes` is absent from the schema at every configuration; `slotDurationMinutes` is absent only when pinned.
4. every argument a conditional schema can remove carries an operator note.
5. the #118 tautology check now asks its question of **every** argument rather than of `calendarId` — the second conditional schema arrived carrying the same sentence on a different argument, and a test naming one argument would have let it through.

## Compatibility

- An integration that pinned a **spacing** is unaffected in practice: the model's override was never the operator's intent, and the pinned value is what applies now.
- An integration that pinned a **length** stops honoring a model-sent override. Switching the console's "Appointment length" to "Let the AI choose" restores per-request choice. Worth a line in the release notes.

### DIFF
--- a/src/client/lib/toolpackTools.ts
+++ b/src/client/lib/toolpackTools.ts
@@ -56,6 +56,18 @@ export function toolpackArgNote(
       "The agent only receives this argument when the integration allows several calendars; with a single one it is used automatically.",
     );
   }
+  // Same shape as calendarId: `slotDurationArgSchema` removes the argument whenever the appointment
+  // length is fixed, so the note is the only place the operator learns that "Let the AI choose" is
+  // what puts it back.
+  if (
+    argName === "slotDurationMinutes" &&
+    toolName === "calendar_check_availability"
+  ) {
+    return t(
+      "toolpackTools.argNote.slotDurationMinutes",
+      'The agent only receives this argument when the appointment length is set to "Let the AI choose"; a fixed length always applies and cannot be changed per conversation.',
+    );
+  }
   return null;
 }
 
--- a/src/client/locales/en.json
+++ b/src/client/locales/en.json
@@ -2007,7 +2007,8 @@
   },
   "toolpackTools": {
     "argNote": {
-      "calendarId": "The agent only receives this argument when the integration allows several calendars; with a single one it is used automatically."
+      "calendarId": "The agent only receives this argument when the integration allows several calendars; with a single one it is used automatically.",
+      "slotDurationMinutes": "The agent only receives this argument when the appointment length is set to \"Let the AI choose\"; a fixed length always applies and cannot be changed per conversation."
     },
     "asaas_create_pix_charge": {
       "desc": "Open a PIX charge and return the copy-and-paste code plus the payment page.",
--- a/src/client/locales/pt-BR.json
+++ b/src/client/locales/pt-BR.json
@@ -2015,7 +2015,8 @@
   },
   "toolpackTools": {
     "argNote": {
-      "calendarId": "O agente só recebe este argumento quando a integração permite vários calendários; com um só, ele é usado automaticamente."
+      "calendarId": "O agente só recebe este argumento quando a integração permite vários calendários; com um só, ele é usado automaticamente.",
+      "slotDurationMinutes": "O agente só recebe este argumento quando a duração do agendamento está em \"Deixar a IA escolher\"; uma duração fixa vale sempre e não muda por conversa."
     },
     "asaas_create_pix_charge": {
       "desc": "Abre uma cobrança PIX e retorna o código copia-e-cola e a página de pagamento.",
--- a/src/modules/integrations/toolpacks/google-calendar.ts
+++ b/src/modules/integrations/toolpacks/google-calendar.ts
@@ -322,6 +322,17 @@ function resolveSlotDuration(
   );
 }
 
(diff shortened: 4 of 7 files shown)


gabrieljablonski:
Thank you for this — and especially for the part that is easy to skip: you went back and checked whether the model had actually hallucinated `14:15` before writing a prompt fix. It hadn't, and the tool obeying a model-sent `granularityMinutes: 15` is a much worse bug than a hallucination, because it produces a real, bookable slot. We reproduced it against `main`: your pinned-config case returns five extra `:15` and `:45` slots today.

We've taken the PR the rest of the way rather than sending it back. It's rebased onto `main` (which meanwhile added the aggregate-calendar search and `calendarArgSchema`, so the rebase was not mechanical — resolving it naively would have dropped the conditional `calendarId` removal from #98), and pushed to your branch with your commit intact. Three changes on top:

1. **`granularityMinutes` leaves the schema outright, not conditionally.** This is the one that matters. The console has no "Let the AI choose" for spacing — only for the length — so a config *without* `slotGranularityMinutes` is an operator who never made the choice, not one who delegated it. Making the removal conditional left your exact failure open for any integration created through the API: the model sends `granularityMinutes: 5` and gets a 5-minute grid over a 60-minute service. There's now a test with no spacing configured pinning that.

2. **The length keeps its conditional removal**, as `slotDurationArgSchema`, sitting next to `calendarArgSchema` — same problem, same shape, same place. The handler guard went away with it: zod strips a residual key from an older tool definition before the body runs, which is why `calendarArgSchema` never needed one either. Your regression test is what proves the strip, so it earns its keep twice.

3. **The "when is this arg offered" sentence moved out of `.describe()`** into the console's operator note. That's issue #118: the model can only ever read that sentence in the case where it's already true, so it's tokens spent every turn on a tautology — while the console, which lists args by `catalogType` rather than by instance, is the audience that genuinely can't tell why their agent never receives an argument they can see documented. We also generalized the #118 test to ask its question of *every* argument, since your PR is the proof that a test naming one argument doesn't hold the line.

Five regression tests now, all red against `main`. Merging once CI is green; it'll ship in the next release, with the length-override change called out in the notes.

The write-up in the original description — the three runs where the model sent 30 sometimes and 15 other times — is what made the diagnosis obvious. That's exactly the kind of report that's worth more than the patch.

Issue #349, opened 2026-08-26
The agent editor holds a refusal for two of its values; the rest are behind tabs and answer in a toast


gabrieljablonski:
Split out of #320, which wired every other form in the console.

## What is left

`useFieldRefusal` holds two of the agent editor's values: `name` and `systemPrompt`, the two it renders a control for directly on the General tab. Everything else an agent write can be refused about lives in a bag and is edited behind one of eight tabs:

- `settings.tts.normalizeCredentialRef`, `settings.stt.credentialRef`, `settings.vision.credentialRef` and the rest of the eight refs `requireVaultRef` names by their dotted path;
- `guardrails.customPolicy`, `guardrails.input.*`, `guardrails.output.*` — the text caps that raise `SettingsTextTooLongError(field, length, max)`;
- `kanban.instructions`, `handoff.instructions`, `toolGuidance.*`, `vision.extractionPrompt`, `followUp.steps[…]`;
- `modelConfig.credentialRef`;
- the tool-grant id fields `bigOrThrow` names (`toolDefinitionId`, `mcpServerConnectionId`, `integrationInstanceId`, `documentTemplateId`, `knowledgeBaseIds`).

Today each of those answers in a toast. The sentence is the server's — that part landed — but the operator is on one tab and the value is on another, so a mark written to a control nobody is looking at would be silence, which is why the sweep stopped here rather than declaring the names.

## What it takes

The missing half is the deep-link, and the map for it already exists: `TEXT_CAP_TARGETS` in `src/client/lib/configHealth.ts` turns a dotted settings path into `{ tab, sectionId }`, and `goToIssue` in the editor already navigates to that pair and highlights the section. A refusal naming a path this map covers can do exactly what a config-health warning does today: take the operator to the tab, scroll to the section, and mark the input.

Three things have to be decided rather than defaulted:

1. **Whether a failed save navigates on its own.** Config-health navigates on a click. Moving the operator's tab out from under them because a save failed is a different act, and the alternative — a toast that names the tab and offers the jump — may be the honest one.
2. **The paths the map does not cover.** `TEXT_CAP_TARGETS` deliberately has no entry for ten of the thirteen `toolGuidance` notes, because they have no control in the console at all. A refusal about one of those has nowhere to go and must stay a toast.
[...]

Issue #255, opened 2026-08-25
A refusal from the schema layer answers a body with no error in it, so the console shows a generic sentence for every one of them


gabrieljablonski:
## What happens

When a write fails the route's SCHEMA rather than a service check, Elysia's `VALIDATION` code falls
through to the default branch in `src/app.ts`, so the answer is TypeBox's own shape instead of the
app's. Measured through the real app, on a route declaring `t.Object({ name: t.String({minLength:3}), nested: t.Object({ inner: t.Number() }) })`:

```
POST (body: {"name":"ab","nested":{"inner":"x"}})
422 {
  "type": "validation",
  "on": "body",
  "property": "/name",
  "message": "Expected string length greater or equal to 3",
  "summary": "Expected string length greater or equal to 3",
  "expected": { "name": "   ", "nested": { "inner": 0 } },
  "found":    { "name": "ab",  "nested": { "inner": "x" } },
  "errors": [ … one entry per failing path … ]
}
```

Three consequences, in the order they cost something:

1. **There is no `error` key at all.** `apiErrorMessage` (`src/client/lib/apiError.ts`) reads
   `value.error` and returns `null` for anything else, and `null` is its "transport failure, show the
   generic sentence" branch. So every schema refusal in the console shows a fixed sentence. That is
   the defect #224 is about, on a whole family the sweep in #233 cannot reach: there is nothing in the
   body for a call site to surface.
2. **The published contract says otherwise.** `errorSchema` (`src/api/lib/openapi.ts`) declares 4xx as
   `{ error, field? }` and `openapi.json` carries that for every declared status — 422 included, which
   `knowledge.controller.ts` declares today.
3. **It echoes the request back.** `found` is the whole submitted body and `expected` a synthesized
   object of the same shape. The same information already goes to the server log from the `logger.error`
   in that branch.

It is not localized either: `message` is TypeBox's English, while every `AppError` refusal is
translated per `Accept-Language`.

## Why this is not "pass the field like #231 did"

#245 put a `field` on the `AppError` branch, naming the value by the **server's** name for it
(`guardrails.output.templateMessage`, `systemPrompt`, `key`). The validation branch also knows which
value failed, but names it as a JSON pointer into the request body (`/settings/guardrails/output/templateMessage`).
That is a second vocabulary for the same value, decided by which layer refused it. Making one key
carry both spellings is the decision this issue is about; the plumbing is the easy half.

## What has to be decided

[...]

Issue #233, opened 2026-08-24
68 error call sites drop the reason the server sent and show a fixed sentence


gabrieljablonski:
Sub-issue of #224, and the third of the three: the one an operator actually notices.

**Re-measured on `main` after #245 and #313 landed**, because the counts in the original body ("28 files, 106 call sites, exactly one file reads the server's message") are stale and, more importantly, the wrong denominator: a third of those call sites are showing a fixed sentence *correctly*.

## What happens

Every error toast in `src/client`, classified by whether the server's own sentence is reachable where the toast is raised:

| | n | |
| --- | --- | --- |
| the error is destructured and in scope | 42 | actionable |
| a bare `catch {}` that the `try` reached by `throw err` | 24 | actionable |
| a `catch (e)` that ignores `e` | 2 | actionable |
| already reads it (`apiErrorMessage` / `useFieldRefusal`) | 10 | done |
| a bare `catch {}` with no re-throw — nothing to read | 16 | correct as is |
| client-side validation, or a race reported inside a 200 | 18 | out of scope |
| **total** | **112** | |

So the sweep is **68 call sites**, not 105, and the third row is the one worth naming: `catch {}` placed after `if (err || !data) throw err` receives the Eden error object and discards it at the binding. That reads as "there was nothing to show" and is the opposite.

The mechanism is already there. `apiErrorMessage` returns the server's sentence, or `null` for a failure with no server behind it, which is exactly where the generic sentence is the honest thing. It is the call sites that do not use it.

## Why the fixed sentences are worse than they look

Several of them *sound* specific, which sends the operator to fix something that was never the problem. `hours.saveError` says "Could not save (check the timezone)" for every refusal the business-hours write can answer, duplicate name included.

## On the transport half

Measured, because it decides 16 of the rows above: **Eden does not reject on a transport failure.** It resolves with `{ status: 503, value: { message, line, column, sourceURL } }` — a `value` with no `error` key. So `apiErrorMessage` correctly returns `null` there, and a `catch {}` that no `throw` reaches reliably catches only a fault in our own handler. Several comments in the tree say the opposite ("Eden REJECTS on a transport failure") and are wrong.

## What this issue no longer covers

Putting the refusal **at the input** it names, which the original body also asked for. #232 built the mechanism and wired one screen; doing the remaining forms is per-screen design work (each form has to declare the server names it renders) and is split out so this one can close. See the sibling issue.

## Depends on

[...]

Issue #232, opened 2026-08-24
The console has nowhere to render a refusal at the input it is about


gabrieljablonski:
Sub-issue of #224, and the second of the three: it consumes the wire field the first one adds.

## What is missing

Even where a refusal's message does get shown, it lands as prose in a toast, which can be far from the input that caused it. On a long form the operator still has to work out which field to go fix, and a toast that scrolls away takes the only copy of the reason with it.

There is no form-error mechanism in the console today. Nowhere to put "this input was refused, here is why", so the toast is not merely the default channel, it is the only one.

## What this issue covers

Somewhere to hold a per-field refusal and render it at the input. The shape is open, and the choice matters more than the plumbing:

- where the state lives (per-form, per-page, or a shared hook),
- how a refusal is cleared (on edit, on resubmit, or both),
- what happens to a refusal whose field is not on screen, for instance a collapsed section or another tab,
- whether the toast still fires when the message is already rendered at the input, since showing it twice is its own kind of noise.

## Depends on

The wire format that names the field (the first sub-issue of #224). Building this against the current `{ error: string }` body would mean guessing the field from the prose.

PR #353, opened 2026-08-26
Put a refusal at the input the server named, on every form that writes

### LINKED ISSUE #320: Every form but one sends a field refusal to a toast instead of to the input
Sibling of #233, split out of it so each can close.

## What is left

#231 put the refused field on the wire, #232 built the state that renders it at the control and wired one screen (`CompanyProfileCard`) as the reference. Every other form still sends the refusal to a toast, which is far from the input and scrolls away with the only copy of the reason.

The producers that name a field are already several, and one of them covers every write route:

- `schemaRefusal` — any 422 on any route names the request field it refused;
- `SettingsTextTooLongError` — the dotted settings path;
- `requireVaultRef` — the column or bag path a bad `vault:<id>` was written to;
- `updateCompanySettings` — the key of the patch;
- `bigOrThrow` — the tool-grant id field;
- the documents template refusals — name and slug.

So most forms in the console can already receive a `field` today and none of them acts on it.

## What it takes per form

`useFieldRefusal(fields)` needs each form to **declare the server's names for the inputs it renders**, pass `error={refusal.at(name, value)}` to each `FormField`, and route its submit handler through `capture`. The declaration is the part that is not mechanical: the server's name for an input is a column, a patch key or a dotted path, and which one a given control edits has to be read, not guessed.

## Why per screen

The forms do not group by file, and which screens belong together is not knowable before someone reads them — which is the same reason #233 was split off rather than done blind. Candidates, by where a `field` can actually land: the agent editor (whose fields sit behind tabs, so it also needs the deep-link that `TEXT_CAP_TARGETS` already maps), the Building blocks pools, the vault credential form, Channels, and the admin surfaces.

## Not in scope

Anything about the sentence itself — that is #233, and it should land first so this one is not re-editing the same handlers.

### LINKED ISSUE #329: A refusal rendered into local error state keeps the fixed sentence, and it can name the wrong field
Sibling of #233, found by driving that issue's fix against a live server.

## What happens

Not every screen shows a refusal in a toast. Several render it into local state — an error line inside the modal, a banner above the form — and those were never counted, by the original body or by the re-measure: both scanned `showToast`.

Observed live, on a real server, with the console in pt-BR. Create an MCP server named `Servidor Alfa`, then create a second one with the same name:

```
server  POST /api/v1/mcp-connections → 409
        "mcp connection name already in use"

console "Não foi possível salvar (verifique a URL/comando)."
        (Could not save — check the URL/command.)
```

The name is the problem and the operator is sent to the URL. That is the exact failure #224 describes — a fixed sentence that *sounds* specific is worse than one that does not — and it survives the sweep in #233 because it is not a toast.

## How many

Counted on `main`: **22** places set an error state to a fixed sentence after an `await api.` in the same handler, across 15 files. Modals (`McpEditModal`, `ToolEditModal`, `CreateApiKeyModal`, `RegisterMcpClientModal`, the admin tenant/user/invite modals, `WebhookSubscriptionModal`), the auth pages (`LoginPage`, `SignupPage`, `AcceptInvitePage`, `OAuthConsentPage`), `SettingsProfilePage`, `AlertChannelsSection`, `DocumentsPanel`.

A **separate 12** already read the wire, by hand: `(apiError.value as ApiErrorPayload)?.error || …`, twelve copies of what `apiErrorMessage` does, each with its own idea of what an empty message means (the helper rejects a blank one; the hand-rolled copies show it). Those are not broken, so they are a simplification rather than a fix — but they are the reason the count above is not simply "everything with local error state".

## What it needs

The same one-line idiom the toast sweep used, plus a fence that reads this shape. The fence in `tests/client/error-toast-reason.test.ts` keys on `showToast(…, "error")` and is blind here by construction; extending it means asking "does this handler set an error state it did not read from the response".

## Not in scope

The toasts (#233, landed) and putting the refusal at the input it names (#320).

gabrieljablonski:
Fixes #320
Fixes #329

#231 put the refused field on the wire, #232 built the state that renders it at the control and wired one card as the reference, and #233 made every error toast show the sentence the server sent. What was left is where that sentence lands, and one place it still did not reach at all.

### A refusal about one input went to a banner, on every form but one

Counted on `main`: **24 components** render a form control and write it back, and exactly one — `CompanyProfileCard` — held the refusal. The other 23 sent it to a toast or to an error line at the top of the modal, which is far from the input and, on a form of sixteen fields, leaves the operator counting down to work out which one the server meant.

### A refusal rendered into local error state kept a fixed sentence (#329)

The sweep in #233 keyed on `showToast(…, "error")`, so a form that renders its own error line was invisible to it by construction. Re-measured here by rule rather than by grep, on `main`:

| shape | count |
| --- | --- |
| shows a FIXED sentence where the server sent one | **7** |
| reads the wire by hand, `(apiError.value as ApiErrorPayload)?.error \|\| …` | 12 |
| a boolean flag after a failed READ ("could not load this page") | 34 |

The 7 are the defect. The worst of them is the one #329 measured live: creating a second MCP connection under a name already taken answers `409 mcp connection name already in use`, and the modal said **"Could not save (check the URL/command)"** — the wrong input, named confidently. `tools.saveError` did the same with "check the name and URL", and `hours.saveError` said "check the timezone" for every refusal a business-hours write can answer, a duplicate name included.

### The server knew the field and did not say so, on 23 refusals

This is the half that makes the other two possible, and it is not something the console can fix.

`requireVaultRef` takes the name of the input a `vault:<id>` arrived in. **Eleven of its thirteen call sites omitted it**, while every one of them held the name: `data.credentialRef`, `params.inboundSecretRef`, `parsed.secretRef`. The argument for optional was that most callers "refuse a column the client already named in the patch it sent" — but what the client SENT and what the server REFUSED are different questions, which is the premise of #231. An integrations write carries `credentialRef` **and** `inboundSecretRef` in one body; a tool write carries `credentialRef` among sixteen keys.

`ConflictError` had the same shape: **nine** uniqueness refusals that name a column in their own sentence (`tenant slug already in use`, `mcp connection name already in use`, `vault entry name and type already in use`) sent no field. So did three hand-built `{ error: "Email already in use" }` bodies in the auth and admin controllers.

And one **shared** producer, found in the last review round: `refuseUnstorable` — reached by every document, approval and knowledge-base write — passed the field only in `translationParams`, which is what the locale template reads. Its body went out as `{ error }` alone, so the two names it refuses by, `title` and `text`, could not be placed on the boxes those forms draw. A call-site sweep does not find this one; only asking the wire does.

The fence in `tests/api/lib/refusal-callsites.test.ts` cannot see any of these: it reads the CODE of the arguments, and these refusals spell the name in prose.

## What this changes

**On the server**, the field becomes an argument the caller cannot forget: `requireVaultRef(db, ref, field)` takes it as a **required** parameter, so the compiler is the guard and the next call site is born naming its input. The nine conflicts, the three email bodies and `refuseUnstorable` name theirs. Measured over real HTTP against the running app, in pt-BR, same probe on both codes:

| refusal | before | after |
| --- | --- | --- |
| duplicate MCP name (409) | `{"error":"Este nome de conexão MCP já está em uso."}` | `{"error":"…","field":"name"}` |
| malformed `credentialRef` (400) | `{"error":"\"vault:nao-e-um-id\" não é uma referência do cofre…"}` | `{"error":"…","field":"credentialRef"}` |
| blank `name` (422, schema) | `{"error":"O valor enviado em name não é válido.","field":"name"}` | unchanged (#255 already named it) |

**On the client**, all 24 form components route their submit through `capture` and pass `error={refusal.at(name, value)}` to each control — 100 call sites. The refusal reaches exactly one channel: the input when the form draws it, the global toast when it does not.

The hook's argument is **what the form is drawing right now**, and getting there took the review loop four tries. It began as a constant list plus a boolean for "is the form on screen", and that boolean grew a new meaning every round: a dialog dismissed mid-save, then a tab left behind. It was always an approximation of the question `placeRefusal` actually asks — is THIS name one the operator can see — and too coarse for a form that hides some of its own controls. Eighteen controls in this tree are drawn behind a JSX guard; eight of them were declared as though they always were:

| field | drawn only when |
| --- | --- |
| `token` (setup) | `setupTokenRequired` |
| `publicKey` / `secretKey` / `value` (vault) | not pasting a `.env` |
| `baseUrl` (vault) | the kind supports it, and not pasting |
| `paramName` (vault) | the kind needs one |
| `text` / `title` (add content) | the text tab is showing |
| `secretRef` (alerts) | `type === "webhook"` |
| `inboundSecretRef` | auth is not `NONE` |
| `body` / `ackMessage` (tool) | the method writes / ack is on |
| `tenantId` (invite) | the inviter is a SUPER_ADMIN |
| `command` (MCP) | the transport is stdio |
| `name` (starter) | the naming step is showing |
| `brandColor` | the mode is SIMPLE |

Three more mechanism defects the wiring exposed, none visible before a real form was on the other end:

- **`placeRefusal` matched exactly, and the schema boundary refuses an array PER ELEMENT** — and, where the elements are objects, all the way inside: `variants.0.weight`, `windows.0.day`, `exceptions.0.ranges.1.start`. A numeric segment now resolves to the declared list and everything after it belongs to that element; a *named* segment still does not, because `guardrails` and `guardrails.output` are two different values.
- **The value comparison was reference identity.** A form rebuilds its request body every render, so an array or an object read twice is never `===`: every such field read as edited-during-the-request and went to the toast. `sameValue` compares structurally.
- **The "other channel" is inside the dialog for ten of the holders.** Handing a caller a sentence for a form the operator has dismissed only moves the silence one step: the mark used to be written where nobody looks, and the sentence would be. The hook raises the global toast itself once the form draws nothing — and only for a sentence it has, since an empty fallback is how `ChannelsPage` says it words a refusal better than the server does.

`CredentialForm`'s `mapSaveError` is gone. It answered its own localized sentence for a 409 on the premise that the server's arrives untranslated, and it does not: the server translates `errors.vaultNameInUse` for the request's Accept-Language, and its pt-BR sentence ("Já existe um segredo com esse nome e tipo") names the type as well, which the console copy did not.

## The guards

`tests/client/error-toast-reason.test.ts` now reads **both** channels — `showToast(…, "error")` and an error state a form renders itself — through one walker, so the shape that escaped #233 cannot escape again. It follows a sentence through a chain of names (`const held = e => refusal.capture(…)`, then `const toast = held(err)`), which is what a form with two failure branches writes.

`tests/client/field-refusal-fence.test.ts` is new. Each rule in it exists because a review round found the defect it now blocks:

| rule | what it blocks |
| --- | --- |
| `unheldWrites` | a form that writes and sends its refusal to a banner |
| `unheldBranches` | one wired branch answering for a handler, while the `catch` writes a fixed sentence |
| `silentDeclarations` | a name declared with no `at(…)` behind it — marked as placed, shown to nobody |
| `holdersBlindToTheScreen` | a bare constant list, which claims every field is drawn always |
| `guardedButUnconditional` | a control drawn behind a JSX guard the declaration does not mirror |
| `uncleanedHolders` | a dialog's holder outliving the session that produced it |
| `deadReadings` | an `at(…)` behind a `??` whose left operand is a `useState("")` and never falls through |
| `distrustedNulls` | a caller substituting its own sentence for the hook's "already told them" |
| `tautologicalStaleness` | `sent` and `current` handed the same snapshot |
| `halfUsedHolders` | a holder only read, or only captured |

Three of those rules first shipped in a form that asked about the FILE — does it have a dialog, a tab, an inline editor — and each version missed the shape the next round found. `guardedButUnconditional` is the one that states the question: one JSX conditional in the idiom this repo writes it in, with the expression carrying no parens or braces of its own. Run over the tree it named the eighteen guarded controls above, six of them declared unconditionally — two the review found and four it had not reached.

Every rule carries fixtures for the shapes it must and must not flag, so a scan that stops matching fails instead of passing empty. Three waiver ledgers are pinned by size **and** asserted in both directions, so an entry describing code that changed fails rather than going on waiving whatever replaced it.

## What of the report does not hold

#329 counted 22 sites setting an error state to a fixed sentence. Measured by rule, **7** show a fixed sentence; 12 of the remaining 15 already read the wire by hand and were a simplification rather than a defect, and the rest are boolean load flags after a failed READ, where "could not load this page" is the honest answer and there is no input to attach anything to.

## Validation scope

**Proved by test**
- 6132 tests pass on the whole tree.
- The fences above, each with fixtures, both re-run against the derived Free tree.
- DOM tests that mount the real components against a refusing server: `McpEditModal` (the sentence renders under the control holding the refused value, and the banner stays empty); the same modal dismissed mid-save (the sentence reaches the toast instead of a form that is gone); `SetupPage` with token enforcement off (a refusal about the hidden token is read out); `CredentialForm` in `.env` mode and across a type change; `KnowledgeApprovals` reopening an editor; the add-content dialog answered on the file tab.
- A wire test through the real app for `refuseUnstorable`, because the DOM tests for that path manufactured `field: "text"` in a fetch mock and the producer was never sending it.
- Every fix in the fourteen review rounds is killed by at least one mutation; the mutations are named in the round commits.

**Proved live**
- The wire table above: the app running locally, a seeded throwaway tenant, the same probe run against the code with and without the producer change, over HTTP with `Accept-Language: pt-BR`. The tenant was deleted afterwards.
- The browser half, on the real console in pt-BR: with `Servidor Alfa` already registered, saving a second one under that name renders "Este nome de conexão MCP já está em uso." **inside the Nome field** (it is also part of the input's accessible name, so a screen reader announces it), the modal footer stays empty, and the URL field carries no mark. The only console entry is the expected 409 — no unhandled rejection.

**Not validated**
- `AdminBrandingPage`, `CreateTenantModal`, `EditTenantModal` and the tenant-admin service are Pro-side files with no counterpart in this repo, so their wiring is not exercisable here.
- The **agent editor is partially held**: `name` and `systemPrompt`, the two values it renders a control for directly. Everything else it writes lives in a settings bag behind one of eight tabs, and placing one of those means also taking the operator to the tab that holds it. Split out as #349; the abstention is pinned in `PARTIALLY_HELD` so it costs an edit rather than passing as a clean scan.
- `guardedButUnconditional` asks that a declaration mention the STATE its guard turns on, not that the two conditions are equivalent. A declaration that names the right state under the wrong comparison would pass it.
- The gate on `KnowledgeApprovals`' editor is consistency with the model, not a reachable failure: Cancel is disabled while the PATCH is out, so that editor cannot close under its own save.
- No per-locale review of the sentences the server now sends where the console used to word its own.

### DIFF
--- a/src/api/features/admin/admin.controller.ts
+++ b/src/api/features/admin/admin.controller.ts
@@ -311,6 +311,7 @@ export const adminController = new Elysia({
           set.status = 409;
           return {
             error: translate("errors.emailInUse", "Email already in use"),
+            field: "email",
           };
         }
         throw error;
--- a/src/api/features/auth/auth.controller.ts
+++ b/src/api/features/auth/auth.controller.ts
@@ -423,6 +423,9 @@ const baseAuthController = new Elysia({
               "errors.currentPasswordIncorrect",
               "Current password is incorrect",
             ),
+            // The form has two password boxes and this refusal is about exactly one of them. Built by
+            // hand here rather than raised as an AppError, so `refusalBody` never sees it.
+            field: "currentPassword",
           };
         }
         throw error;
--- a/src/api/features/auth/auth.controller.ts
+++ b/src/api/features/auth/auth.controller.ts
@@ -196,6 +196,9 @@ const baseAuthController = new Elysia({
         set.status = 400;
         return {
           error: translate("errors.emailInUse", "Email already in use"),
+          // The one input the operator fixes, named the way every other refusal names one (#231).
+          // Built by hand here rather than raised as an AppError, so `refusalBody` never sees it.
+          field: "email",
         };
       }
 
@@ -510,6 +513,7 @@ const baseAuthController = new Elysia({
           set.status = 409;
           return {
             error: translate("errors.emailInUse", "Email already in use"),
+            field: "email",
           };
         }
         throw error;
--- a/src/client/components/BusinessHoursForm.tsx
+++ b/src/client/components/BusinessHoursForm.tsx
@@ -8,8 +8,8 @@ import { useUnsavedChanges } from "@/client/components/Modal";
 import { Select } from "@/client/components/Select";
 import { TimezonePicker } from "@/client/components/TimezonePicker";
 import { useToast } from "@/client/components/Toast";
+import { useFieldRefusal } from "@/client/hooks/useFieldRefusal";
 import { api } from "@/client/lib/api";
-import { apiErrorMessage } from "@/client/lib/apiError";
 
 type HoursData = Awaited<
   ReturnType<(typeof api.api.v1)["business-hours"]["get"]>
@@ -84,6 +84,10 @@ export interface BusinessHoursFormProps {
 // BusinessHoursPanel (via modal) and SchedulePicker (inline create/edit).
 // When rendered inside a <Modal>, calls useUnsavedChanges so the modal's
 // discard-confirmation guard fires automatically on dirty state.
+// The keys of the body this form writes. The window and exception lists are refused per element
+// (`refused body.windows.0`), which lands on the list they belong to: see placeRefusal.
+const HOURS_FIELDS = ["name", "timezone", "windows", "exceptions"] as const;
+
 export function BusinessHoursForm({
   mode,
   initial,
@@ -101,6 +105,7 @@ export function BusinessHoursForm({
   const [windows, setWindows] = useState<Window[]>(
     initial?.windows ?? DEFAULT_WINDOWS,
   );
+  const refusal = useFieldRefusal(HOURS_FIELDS);
   const [exceptions, setExceptions] = useState<Exception[]>(
     initial?.exceptions ?? [],
   );
@@ -183,6 +188,20 @@ export function BusinessHoursForm({
   const hasInvalidWindow = windows.some(windowInvalid);
   const hasInvalidException = exceptions.some(exceptionInvalid);
 
+  // What the inputs hold right now, in the server's vocabulary, and what the write sends.
+  const currentRef = useRef({
+    name: name.trim(),
+    timezone: timezone.trim() || "UTC",
+    windows,
+    exceptions,
+  });
+  currentRef.current = {
+    name: name.trim(),
+    timezone: timezone.trim() || "UTC",
+    windows,
+    exceptions,
+  };
+
   async function save() {
     if (!name.trim()) return;
     if (hasInvalidException) {
@@ -212,27 +231,35 @@ export function BusinessHoursForm({
       windows,
       exceptions,
     };
+    const held = (e: unknown) =>
+      refusal.capture(
+        e,
+        t("hours.saveError", "Could not save."),
+        body,
+        currentRef.current,
+      );
     try {
       if (mode === "update" && initial?.id) {
         const { data, error: err } = await api.api.v1["business-hours"]({
           id: initial.id,
         }).patch(body);
         if (err || !data) throw err;
+        refusal.clear();
         showToast(t("hours.saved", "Business hours saved."), "success");
         onSaved(data.businessHours.id, data.businessHours.name);
       } else {
         const { data, error: err } =
           await api.api.v1["business-hours"].post(body);
         if (err || !data) throw err;
+        refusal.clear();
         showToast(t("hours.saved", "Business hours saved."), "success");
         onSaved(data.businessHours.id, data.businessHours.name);
       }
     } catch (e) {
-      showToast(
-        apiErrorMessage(e) ||
-          t("hours.saveError", "Could not save (check the timezone)."),
-        "error",
-      );
+      // The fixed sentence this replaces said "check the timezone" for every refusal this write can
+      // answer, a duplicate name included: the confident wrong pointer #224 is about.
+      const toast = held(e);
+      if (toast) showToast(toast, "error");
     } finally {
       setSaving(false);
     }
@@ -241,10 +268,18 @@ export function BusinessHoursForm({
   return (
     <div className="flex flex-col gap-4">
       <div className="grid gap-4 sm:grid-cols-2">
-        <FormField label={t("hours.name", "Name")} required>
+        <FormField
+          label={t("hours.name", "Name")}
+          required
+          error={refusal.at("name", name.trim())}
+        >
           <Input value={name} onChange={(e) => setName(e.target.value)} />
         </FormField>
-        <FormField label={t("hours.timezone", "Timezone")} group>
+        <FormField
+          label={t("hours.timezone", "Timezone")}
+          group
+          error={refusal.at("timezone", timezone.trim() || "UTC")}
+        >
           <TimezonePicker
             value={timezone}
             onChange={setTimezone}
@@ -257,6 +292,11 @@ export function BusinessHoursForm({
         <div className="flex items-center justify-between">
           <span className="font-medium text-sm text-text-secondary">
             {t("hours.windows", "Windows")}
+            {refusal.at("windows", windows) && (
+              <span className="ml-2 font-normal text-error text-xs">
+                {refusal.at("windows", windows)}
+              </span>
+            )}
           </span>
           <Button
             variant="secondary"
@@ -344,6 +384,11 @@ export function BusinessHoursForm({
         <div className="flex items-center justify-between">
           <span className="font-medium text-sm text-text-secondary">
             {t("hours.exceptions", "Holidays and closures")}
+            {refusal.at("exceptions", exceptions) && (
+              <span className="ml-2 font-normal text-error text-xs">
+                {refusal.at("exceptions", exceptions)}
+              </span>
+            )}
           </span>
           <Button
             variant="secondary"
--- a/src/client/components/CredentialForm.tsx
+++ b/src/client/components/CredentialForm.tsx
@@ -437,6 +437,12 @@ export function CredentialForm({
     }
   }
 
+  // The inner inputs of a multi-field secret, by the names the server refuses them under.
+  const fieldSnapshot = () =>
+    Object.fromEntries(
+      (fields ?? []).map((f) => [f.key, fieldValues[f.key] ?? ""]),
+    );
+
   // What the inputs hold right now, in the server's vocabulary. Each write below sends a subset of
   // these, and `capture` compares only the key it was refused about.
   const currentRef = useRef<Record<string, unknown>>({});
@@ -464,7 +470,12 @@ export function CredentialForm({
     refusal.capture(
       e,
       t("vault.saveError", "Could not save the secret."),
-      sent,
+      // The per-field values ride along, and they are not on the wire: a multi-field secret is sent
+      // as ONE `value` blob, so a snapshot built from the body alone carries no `api_key` for
+      // `placeRefusal` to compare against. Without them the staleness check has nothing to check,
+      // and a value the operator replaced while the request was out gets marked as the one the
+      // server refused.
+      { ...sent, ...fieldSnapshot() },
       currentRef.current,
     );
 
--- a/src/client/components/CredentialForm.tsx
+++ b/src/client/components/CredentialForm.tsx
@@ -24,6 +24,7 @@ import { McpOAuthSection } from "@/client/components/McpOAuthSection";
 import { useUnsavedChanges } from "@/client/components/Modal";
 import { Textarea } from "@/client/components/Textarea";
 import { useToast } from "@/client/components/Toast";
+import { useFieldRefusal } from "@/client/hooks/useFieldRefusal";
 import { providerLink } from "@/client/lib/affiliateLinks";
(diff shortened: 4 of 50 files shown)


gabrieljablonski:
### Round 4 — recorded here, not as a bot review

The review-bot token comes from a 1Password mount that stopped serving mid-loop, so this round could
not be posted as a review. Recording it verbatim instead, under my own account, as the author
answering. **This does not clear the gate**: the round 3 review (`CHANGES_REQUESTED`) is still
standing and only a bot review on a clean round takes it down.

One finding, P2, and it is real:

> **Show refusals when local chunk errors are empty**
>
> When editing a base, a decimal such as `100.5` passes `validateChunkFields` because it only checks
> bounds, but the API's `t.Integer` schema rejects it with `field: "chunkSize"`. `capture` then
> suppresses the toast because it considers the refusal placed, while this nullish coalescing selects
> the initialized/reset empty string instead of the held refusal, leaving Save with no visible
> feedback. The same issue affects `chunkOverlap`; use a truthy fallback or nullable local-error
> state.

Confirmed by reading, on both halves:

- `chunkSizeError` and `chunkOverlapError` are `useState("")`, so they are **never nullish** and the
  `??` in front of the reading never falls through. The reading is dead code.
- `knowledge.controller.ts` declares `chunkSize: t.Optional(t.Integer({ minimum: 100, maximum: 8000 }))`,
  and the input is `<Input type="number">` piped through `Number(e.target.value)`. `100.5` clears the
  local bounds check and is refused by the schema, by name.

So the refusal is placed on the holder, `capture` reports "it is on the control" and keeps the toast
quiet, and the one reader that could render it cannot run. Silence, which is the outcome this whole
change exists to remove.

Fixed at all four sites (create and edit, size and overlap) with `||`, and swept as a family: the
fence now has a `deadReadings` rule that flags any `at(…)` sitting behind a `??` whose left operand
is a state initialized to a string literal. Four mutations against it — one site reverted to `??`,
the left-operand filter dropped, the state collection disabled, and a nullable local error as the
negative control — kill a test each.

`bun check`: 6093 pass / 0 fail.

gabrieljablonski:
### Round 5 — recorded here, not as a bot review

The 1Password mount is still down, so this round could not be posted as a review either. Same as
round 4: recorded verbatim, under my own account, and **this does not clear the gate** — the round 3
`CHANGES_REQUESTED` is still standing.

Three findings, all three real.

> **[P2] Match nested fields inside array elements**
>
> When schema validation rejects a property of an object in an array, such as `windows[0].day`,
> `fieldFromPointer` returns `windows.0.day` (the existing schema-refusal test similarly expects
> `items.0.name`). This regex only accepts `windows.0`, so the newly wired Business Hours list cannot
> render these refusals and falls back to a toast.

Confirmed. `business-hours.controller.ts` declares `windows: t.Optional(t.Array(windowSchema))` where
`windowSchema` is an object of `day`/`start`/`end`, and `tests/api/v1/write-body-required.test.ts`
already asserts `field === "variants.0.weight"` for exactly this shape. The exception in
`placeRefusal` stopped at the index, so `windows.0` matched and `windows.0.day` did not. It now reads
the index as an index: everything after it is inside that element, and `exceptions.0.ranges.1.start`
lands on `exceptions` too. A *named* segment still stays unmatched, which is the half that keeps
`guardrails` from swallowing `guardrails.output`.

> **[P2] Route thrown alert-save failures through `held`**
>
> When an alert-channel save is in flight and the dialog is dismissed through X, Escape, or the
> backdrop before a network-level rejection, only the resolved HTTP-error branch uses `held`; the
> `catch` still writes to the modal-local `error` state, which is then hidden.

Confirmed, and the sentence does not even change: the catch wrote `t("alerts.saveFailed", …)`, which
is the same string `held`'s fallback carries. What changes is the channel — Eden *rejects* on a
transport failure, so the unwired branch is the one a broken network lands in.

> **[P2] Clear connect refusals on every modal opening**
>
> After a connect attempt produces a field refusal, canceling and manually reopening the Connect
> dialog preserves this holder because only the URL-driven opening path calls
> `connectRefusal.clear()`; `openConnect()` does not.

Confirmed. `connectModal` has no `useOnModalOpen`, so its reset lives inline at each `.open()`, and
the `?connect=` path had one while the Connect button did not. The mark expires by value, so it comes
back the moment the operator retypes the URL that was refused.

Two of the three got past the fence, so the fence changed:

- `uncleanedHolders` pooled every reset block in a file into one string, so one cleared opening
  vouched for all of them and one holder's clear vouched for the others. It now attributes by the
  dialog the holder itself names in its second argument, and asks the right site: `useOnModalOpen`
  where it exists (it runs on every opening by construction), every `.open()` where it does not. That
  also retired the `HELD_BY_THE_PAGE` ledger — a holder that names no dialog is a page's, and the
  rule can now tell.
- New rule `unheldBranches`: `unheldWrites` asks per handler and one `capture` anywhere inside
  answers for it, which is exactly how the alert `catch` passed. Bounded to writes *after* the
  request goes out, so a pre-submit guard ("Passwords do not match", "Headers must be valid JSON.")
  is not asked to route through a server it never called.

Five mutations, five killed: the index regex back to end-of-path, the catch writing directly, the
`openConnect` clear removed, the reset-site rule back to pooling, and the request boundary ignored.

`bun check`: 6103 pass / 0 fail.

Issue #318, opened 2026-08-25
An inbox bound to no agent swallows every message and reports nothing


gabrieljablonski:
An `Inbox` row with no `agentId` is a dead end that reports nothing. Every message that arrives on it
is mirrored, consumes its delivery, and produces one process log line (`outcome=no-agent`); the
conversation shows the customer waiting and the console shows nothing wrong.

## How the rows get there

The mirror creates them. `upsertInbox` (`src/modules/chatwoot/mirror.ts`) upserts by
`chatwootInboxId` on every event, and the `create` branch sets `name` and `channelType` and no agent.
So any inbox that sends us traffic before an operator binds it appears in the table, unbound, and
starts silently swallowing messages. That auto-creation is deliberate and load-bearing for mirroring;
what is missing is anything that says the row is inert.

## Why it matters

Measured while investigating #270: a `/teste` on such an inbox activates nothing and posts nothing,
and the operator's only clue is a log line on the server. The same silence covers real customer
traffic, not just commands — an unbound inbox is indistinguishable from an agent that is merely quiet.

This is the configuration error most likely to be hit right after connecting a new channel in
Chatwoot, which is exactly when the operator has the least context to guess what is wrong.

## What this needs

Not decided, and worth deciding together rather than picking one:

- the Channels screen showing which bound inboxes have no agent, instead of only listing them;
- a signal on the traffic itself (an `ExecutionLog` row, see the sibling issue) so the Logs page can
  answer "why did nothing happen"; or
- refusing the auto-create and treating an unknown inbox as an event we cannot serve, which is the
  shape #312 chose for an unknown event body — a bigger change, and it needs a look at what breaks
  for a conversation whose inbox genuinely arrives first.

Issue #228, opened 2026-08-24
A delivery stranded by a process death is invisible: nothing marks it, and the lost message reaches no one


gabrieljablonski:
> **Scope note (2026-08-25):** this issue asked for two things — stop losing the message *silently*, and *retry* it. Four review rounds on #282 established that the retry is a much harder problem than the body below assumed, for a reason the body could not have known: the gates a recovery needs do not survive the process that died. The retry half is now **#295**, with that reasoning written down. This issue is the half #282 closes: a stranded delivery stops being invisible.

## What happens

A Chatwoot delivery is claimed in a ledger row (`ChatwootWebhookDelivery`) and then processed:

1. the receiver acks;
2. the detached path writes the row as `PENDING`;
3. `processChatwootDelivery` CASes it `PENDING -> PROCESSING`, does the work, then CASes it
   `-> PROCESSED`.

If the process dies between 3's first CAS and its last (a deploy, an OOM, a restart), the row is
stranded on `PROCESSING`. Nothing ever moves it again. Every redelivery of that event finds the row,
the CAS matches nothing, and the call returns `skipped`.

Chatwoot has already been handed a 200, so it will not send that event again either. The customer's
message is gone, silently, and the only trace is a row sitting on `PROCESSING` that nobody reads.

The same is true of `PENDING`, one step earlier: the ack is spent before the ledger row is even
written, so a death between the insert and the CAS strands the row there. #226's answer to that — a
redelivery goes on to the CAS instead of being dropped — only helps when a redelivery actually
arrives, and holding a 200 means one usually does not.

## What this issue asks for

That a stranded delivery stop being invisible:

- every non-terminal row eventually becomes terminal, and says which of two things happened —
  nothing was outstanding, or a customer message went unanswered;
- the second case is findable without reading the ledger by hand, and reaches the operator through
  the channels that already carry failures.

Retrying the delivery is **#295**, deliberately not here.

## Scope

- `src/modules/chatwoot/webhook.ts` (what the ledger insert records)
- `prisma/schema.prisma` (`ChatwootWebhookDelivery`)
- a scheduler job for the sweep

Out of scope: the ack path itself, which no longer touches Postgres (#226), and the pool inversion
(#227).

Found while working on #225.

matheusvitorino28:
@evidencia-228.md

Issue #79, opened 2026-08-15
/reset builds a Chatwoot client with an empty bot token, so clearing custom attributes 401s while the customer is told everything was cleared


gabrieljablonski:
Origin: [community report](https://www.lucasmoreira.ai/c/perguntas-e-respostas/bug-evento-read-do-instagram-derruba-o-instagrameventsjob-nomethoderror-chatwoot-4-16-2).

The `/reset` handler builds its Chatwoot client **without a bot token**, so every bot-token call it makes goes out unauthenticated and is rejected with 401. Labels get cleared (they use the admin token), custom attributes do not, the linked kanban card is never reached, and the customer is still told that everything was cleared.

## What happens

`loadChatwootClient` defaults the bot token to an **empty string** when the caller does not supply one ([`src/modules/chatwoot/instance.ts:49`](https://github.com/fazer-ai/agents/blob/main/src/modules/chatwoot/instance.ts#L49)):

```ts
botToken: deps.botToken ?? "",
```

The `/reset` block is the only call site that omits it ([`src/modules/chatwoot/webhook.ts:829-831`](https://github.com/fazer-ai/agents/blob/main/src/modules/chatwoot/webhook.ts#L829-L831)):

```ts
const client = await loadChatwootClient(tenantId, instanceId, { base });
await client.setConversationLabels(conversationId, []);
await client.setConversationCustomAttributes(conversationId, {});
```

So the request carries `api_access_token: ""`. On the Chatwoot side that is a blank token: `AccessTokenAuthHelper#ensure_access_token` leaves `@access_token` nil and `authenticate_access_token!` renders `Invalid Access Token` with HTTP **401**, before any authorization logic runs.

The observable result is the warn line this block emits, on every `/reset`:

```
chatwoot: /reset clear labels/attributes/card failed (conv=N): Chatwoot API 401 for POST /conversations/N/custom_attributes
```

Labels survive the same block because they use a **different token**: `setConversationLabels` is sent with the admin token ([`client.ts:377`](https://github.com/fazer-ai/agents/blob/main/src/modules/chatwoot/client.ts#L377)), which `loadChatwootClient` always loads from the deployment, while `setConversationCustomAttributes` is a bot-token call ([`client.ts:349`](https://github.com/fazer-ai/agents/blob/main/src/modules/chatwoot/client.ts#L349)). The two are not the same auth path, which is what makes the failure look selective.

[...]

Issue #49, opened 2026-08-10
Direct (no-debounce) path has no supersede gate: a burst of customer messages produces concurrent turns and duplicate replies


CarlosAlvesz:
## Description

With `debounce.enabled = false`, every inbound message starts its own turn immediately and **the direct path has no supersede gate**. Two customer messages that arrive inside the model's response latency therefore produce two concurrent turns that never observe each other: both build their context from the same history, both generate, and both post. The customer receives duplicated content.

This is documented in the source itself (`src/graph/runtime.ts`, v1.3.0):

```
// The direct path never uses shouldPost (no supersede gate), so nothing else advances the handled
// watermark here [...]
```

The debounce flush has both protections (the coalescing re-fetch and the `shouldPost` supersede CAS just before posting), but a deployment that cannot run debounce (see the note at the bottom) is left with no protection at all.

## Steps to reproduce

1. Agent bound to a `Channel::Api` inbox (Evolution/Baileys bridge), `mode: production`, `debounce.enabled = false`, `split.enabled = true`.
2. From the customer side, send two short messages a few seconds apart — comfortably inside the model's latency (observed ~12-14s per turn on `gemini-3.5-flash`).
3. Both turns run concurrently; both post.

Observed on a live instance: two inbound messages 5 seconds apart produced **6 outgoing balloons in 12 seconds**, containing two different greetings, the same explanation reworded twice, and the same closing question posted **verbatim twice**. Both turns are visible in `/logs` as two separate `generate` lines with distinct `turnId`s on the same `conversationId`.

## Expected

The direct path should be protected against concurrent turns on the same conversation, in whatever way fits the design. For example: apply the same last-moment supersede check `shouldPost` performs (drop the reply if a newer inbound message arrived while the model was running), or gate the direct path on the `lastHandledMessageId` watermark that v1.3.0 already advances there.

Today the watermark is written on the direct path but never read by it, so it only benefits a later debounce flush.

## Where

- `src/graph/runtime.ts` — direct (webhook) entry, around the `advanceHandledWatermark` call that carries the comment quoted above.
- `src/modules/debounce/handler.ts` — `shouldPost`, the supersede CAS the direct path lacks.

## Note: why "just enable debounce" is not a free workaround on upstream Chatwoot

The natural mitigation is `debounce.enabled = true`, and on chatwoot-pro that is likely fine. On an instance running **upstream Chatwoot** it currently trades one defect for another:

[...]

PR #380, opened 2026-08-26
Answer the customer a stranded delivery left waiting, by running the delivery path again

### LINKED ISSUE #295: A stranded delivery is reported but never re-answered: the gates a recovery would need do not survive the process that died
Split out of #228, which asked for two things: stop losing the message *silently*, and *retry* it. #282 delivers the first. This is the second, and it is a much harder problem than the issue assumed — four review rounds on #282 went into finding out why, so the reasoning is here rather than lost.

## What is left undone

A delivery stranded by a process death is now detected: the row goes `DEAD` and an error-level `delivery` line lands on the conversation. Nothing re-answers the customer. Their message is still unanswered; the operator just knows about it now.

## Why the two obvious recoveries are both wrong

Recovery means running a turn, and `coalesceAndRunTurn` already re-reads the conversation's messages from Chatwoot — so the event body never needs to be stored, and the whole recovery looks like "make the flush run". Two ways to do that, both tried on #282 and both rejected:

**Arm a `DEBOUNCE` job.** `DEBOUNCE` is drained only by the debounce lane, which an install can switch off (`DEBOUNCE_WORKER_ENABLED`) — and a recovery has to work for agents whose own debounce switch is off, which is the population most likely to have the worker off. `scheduler/lanes.ts` documents this exact trap for `INGEST_MESSAGE`. It also reports a success that has not happened: a row marked terminal because a job was enqueued.

**Run the flush inline in the sweep.** This is the one that fails for the interesting reason.

## The reason: a flush is the second half of a gated path

`flushDebounceJob` re-checks two gates itself — ownership (`shouldBotHandle`) and contact authorization. It does **not** re-check the three the delivery path applies before ever arming it:

| gate | applied by | survives the strand? |
| --- | --- | --- |
| ownership | the flush, itself | yes |
| contact authorization | the flush, itself | yes |
| test mode (`isTestSilenced`) | `processChatwootDelivery` | no |
| availability / out-of-hours (`outOfHoursGate`) | `processChatwootDelivery` | no |
| redirect | `processChatwootDelivery` | no |

So a recovery that calls the flush replies out of hours, or on a conversation whose test mode was never activated — a message the original delivery would have suppressed. The gate verdicts are not persisted, and the process that computed them is gone.

## The shapes worth weighing

1. **Re-apply the three gates from durable state before flushing.** All three read rows the mirror and the agent settings already have, and two of them (`isTestSilenced`, `outOfHoursGate`) are already pure functions with other callers. The cost is a fourth place that answers "may this conversation be answered right now", and the repo has been bitten before by the same question living in several places.
2. **Rebuild the event and re-run `processChatwootDelivery`.** Correct by construction — every gate runs where it already runs — and the conversation plus its messages are re-fetchable over REST. The risk is fidelity: the REST message shape differs from the webhook shape (`message_type` is an integer there and a string on the wire), and #257 is an open bug about exactly this class of divergence between the two id sources.
3. **Persist the gate verdicts with the ledger row**, so a recovery can prove they passed rather than recomputing them. Turns the ledger into a record of decisions rather than of deliveries, and only helps for a strand that happened *after* the gates ran.

## Two things a recovery must also answer

Both found by the review rounds on #282, and both apply to any of the shapes above:

- **No turn has a duration ceiling** (no model or tool timeout), so a wall-clock threshold cannot tell a dead process from a long turn. `src/graph/inflight.ts` — the fence a concurrently-fired follow-up already consults — answers it in-process; a turn live on another replica is #203's gap. Without it, a second turn runs beside the first and both turns' tools execute, which the handled-watermark CAS does not prevent because it is checked after they have run.
- **A recovery must not be routed through an optional lane, and must not run the turn inside the shared scheduler pass either**: `runSchedulerTick` waits for the handler before clearing its non-reentrancy guard, so a batch of recoveries holds up every due reminder, follow-up and retry for the sum of their durations.

## What #282 leaves in place for this

`chatwoot_webhook_deliveries` carries `conversation_id` and `inbound_message_id`, and `attempts` is still unused and still the natural place to bound a retry. A row that a recovery answers should move off `DEAD`, which is currently the terminal state for "a customer went unanswered".

## A third thing, found on #282's last round

**A terminal row cannot currently be taken back, and one race needs it to be.** The sweep's terminal write CASes on the status the scan read, so on a `PENDING` row it races that delivery's own `PENDING -> PROCESSING` claim. When the sweep wins, the claim matches nothing and `processChatwootDelivery` returns `"skipped"` — a redelivery or a manual replay landing in that instant is discarded, and with it the one path that would have answered the customer through the real gates, since it runs them all where they already run.

#282 leaves this as it stands on purpose: closing it means letting a claim take a row back out of a terminal state, which is recovery. Whatever shape recovery takes, it should decide whether `DEAD` is reclaimable — and if it is, that same change fixes this race for free, because the claim stops depending on having won it.

gabrieljablonski:
Fixes #295

## What was broken

A Chatwoot delivery stranded by a process death is **reported and then left there**. #228 gave the strand a name — the sweep marks the ledger row `DEAD` and writes an error-level line on the conversation — and stopped exactly where the customer's problem starts: nothing re-sends, Chatwoot holds its 200 and will not redeliver, and a manual replay CASes against `PENDING` and matches nothing. `WHERE status = 'DEAD'` is a list of customers who wrote and were never answered, and reading it is all an operator can do.

The issue is right about why the obvious fix does not work, and I re-measured every claim in its body against the current tree before writing anything:

- `flushDebounceJob` re-checks ownership and contact authorization itself and re-reads the messages from Chatwoot, so the event body is genuinely not needed — **holds**.
- Three gates run only in the delivery path and none of their verdicts survive the process that computed them: test mode, availability / out-of-hours, and the channel redirect — **holds**. A recovery through the flush replies out of hours, or on a conversation whose test mode was never activated.
- Re-implementing those three is worse than it looks, because none of them is a predicate: each one **decides and acts** (the redirect sends the link, availability posts the away message, test mode posts its notice), so a second implementation would have to reproduce the actions and their claims, not just the verdicts. `isTestSilenced` already has five callers; a sixth is the shape of defect this repo keeps paying for.

## What this does

**It runs the delivery path again.** Not a flush, and not a re-implementation of the gates: `processChatwootDelivery` is the function a live delivery goes through, so every gate runs where it already runs.

What made that look unsafe was the fear of re-firing the gates' side effects, and it does not hold — but the reason is **per gate** rather than one shared property, which is what reading all three actually showed:

- **availability posts behind a real CAS.** `claimAwayMessage` is an `updateMany` guarded on the watermark's previous value, and it claims *before* it posts, so a second invocation claims nothing. The comment on it names why it had to be: "the webhook dispatch is DETACHED, so a customer who writes twice in a row lands two invocations that both read the same watermark before either writes it".
- **the test notice (`testNoticeSentAt`) and the redirect (`redirectSentAt`) are one-shot watermarks read before the act and written after it**, which is not the same thing. They are safe for a recovery for a different reason: a recovery is serialized against a live turn by the in-flight fence and against another recovery by the claim CAS, so its read of the watermark is current rather than racing one.

**The residual window, named rather than papered over:** a process that died *between* one of those two acts and its watermark write leaves the watermark null, and the recovery repeats the act — a duplicate private note for the test notice, and for the redirect the fixed link sent to the customer a second time. Both are already reachable without any recovery, because the dispatch is detached and two live deliveries interleave the same way; neither spends a model call or writes conversation state. Closing it means a claim-then-act ordering inside three gates this PR does not own.

**At least once, and that is a property of the design rather than a gap in it.** A process that died left the customer unanswered — that is why the row is `DEAD` — but it did not necessarily do *nothing* first, and nothing in the ledger records how far it got. A turn whose tools had already fired is re-run and fires them again. Closing that needs a durable per-effect claim, which is a change to the delivery path and to every tool.

## Rebuilding the body

The body the path needs was deliberately never stored (#228 keeps no payload at rest), so it is rebuilt — and **where each field comes from is the part five review rounds moved**, because the first version took the whole conversation from the mirror and that is wrong for the one reason that is specific to a strand: *the delivery that would have mirrored this message is the one that died*, so the mirror holds the state from before it.

- **The conversation's state comes from a live `GET /conversations/:id`.** An incoming message on a `resolved` conversation reopens it (`Message#reopen_resolved_conversation`: `pending` on a bot inbox, `open` otherwise), so the mirror still says `resolved` while Chatwoot says otherwise — and copied into the body, that stale value makes `shouldBotHandle` refuse, the row is marked `PROCESSED`, and the customer is never answered. A customer writing again after a conversation was resolved is the ordinary way a new episode starts. The snapshot goes through `reconcileMirrorFromLive`, so the mirror is **repaired** under the same ordering rule the webhook mirror uses (a webhook committed between the GET and the write still outranks it), and the body is built from the row that call returns. A snapshot `parseLiveConversation` cannot trust is a deferral, never a fallback to the mirror.
- **The message comes from a REST read** anchored `before: messageId + 1` so the page ends at it, and **the newest page from a second, unanchored one** — two reads because one page cannot hold both ends: the anchored page says nothing about what came after, and the newest page need not contain the stranded message at all (measured: on a 30-message conversation the default page of 20 did not).
- **The route comes off that message** (`inbox_id`), not off the mirror's `Conversation.inboxId`. That column is nullable and stays null for a conversation whose mirrored events never named an inbox, and a body carrying no inbox makes `runAgentTurn` return `skipped` on its first line: no turn, row `PROCESSED`, closing line saying the loss ended. The mirror is the second reading, for an account that renders no such scalar.
- **Two fields still come from the mirror, and only two.** `contactInboxId`, because the REST conversation renders no `contact_inbox`; and `redirect_origin_display_id`, because the fork renders it from `EventDataPresenter` only. The second one has to travel with **its version** (`chatwootRedirectOriginAt`, sent as the body's `updated_at`): on the wire the pairing and the stamp that orders it are one fact, so stated without a version the old pairing outranks nothing, and a widget re-entered from a second WhatsApp thread during the REST reads would have its new episode overwritten by the old one.

`buildRecoveryPayload` shapes a webhook-shaped body and hands it to `normalizeChatwootEvent`, so exactly one place still knows how a Chatwoot payload becomes an event. Building a `NormalizedChatwootEvent` directly would have been a second reader of that shape, and an A/B test against a captured wire body is what keeps the two honest.

**The claim is one statement.** `processChatwootDelivery` gained `claimFrom`, so a recovery's CAS is `DEAD → PROCESSING` with `attempts` incremented, and the winner owns the row. `DEAD` is the sweep's verdict, reached by inference; a recovery that runs the turn is direct evidence, which outranks it — the same ordering a turn already uses when it runs over a `DEAD` row (`retireCoveredDeliveries`). If the delivery path **throws** after that claim, the row is put back to `DEAD` under a CAS guarded on `PROCESSING`, so a late `tx2` is never overwritten and the next attempt finds a row it can claim rather than one that reads as somebody else's.

**A job of its own**, armed by the sweep at the moment it declares a row lost — the only moment anything knows the row just became recoverable, since the sweep's own query reads `PENDING` and `PROCESSING`. Not work the sweep does inline, and `lanes.ts` already stated the rule: the sweep spends one indexed query and a recovery spends a whole agent turn.

**A closing line of its own.** `retireCoveredDeliveries` writes its correction only for rows it moves out of `DEAD` itself, and a recovered row left `DEAD` at the claim — so without this the row would simply leave the worklist while the page an operator already received stayed open, pointing at a customer they can no longer find. It is `warn`, matching the correction it stands in for, and it names the conversation and the agent so the two lines read as one story.

## What bounds it

- **Never beside a live turn.** The in-flight fence is consulted first and about the **conversation**, not the row: two deliveries for one conversation are two rows, and answering one while a turn runs on the other is what the fence exists to stop. It is asked again immediately before the handoff, after the REST reads, and it narrows the window rather than closing it — the last sliver is the same overlap two live deliveries already have (#203).
- **`MAX_RECOVERY_ATTEMPTS = 3`**, and the number is policy rather than measurement — said plainly in the code, because the alternative is a reader assuming it was tuned. What is not arbitrary is that a bound exists: a recovery runs a real turn, spends the model, and can call side-effecting tools.
- **`MAX_RECOVERY_AGE_MS = 6h`, asked on two clocks.** The first is the ledger row's `receivedAt`, which is when *this* application inserted the row; a webhook delayed by a Chatwoot retry inserts late, so the REST message's own `created_at` — the customer's clock — is asked again, **before** the claim, so a refusal spends no attempt. A reply is a recovery only while the customer is plausibly still waiting; hours later it is a stranger reopening a conversation that moved on, and the operator's worklist is the better place for it.
- **A degraded read fails closed rather than closing the loss.** A rebuild that comes out as anything but a new incoming message describes a REST response that lost something (a missing `message_type` normalizes to `"other"`, which is not null and passed the older guard), and a rebuild that can name no inbox on either reading would be routed nowhere. Both refuse, because handing either to the delivery path is precisely the quiet failure this issue is about: no turn, row `PROCESSED`, a line saying the loss ended.
- **A control command is not replayed.** The premise of re-running the path is that it did not complete, not that it did nothing: `/reset` performs its deletion before the tail settles, so a replay deletes the memory the conversation accumulated *since*. Refused rather than made idempotent, because both things a command has that a customer message does not point the same way — its author is an operator who is present and can retype it, and its effect is destructive rather than a reply. The row stays `DEAD`, which is the operator-facing record that lets them do that.
- **A recovery answers only while its message is still the newest thing the customer said**, and that bound belongs to the delivery path rather than to this module. `shouldPost` re-fetches immediately before posting, sees a newer incoming id and withholds the reply, so the turn comes back `superseded` while the row is settled all the same. For a live delivery that is right — the newer message's own delivery carries the reply. For a recovery the premise is gone: that delivery already ran and answered the newer message, and a direct turn feeds the graph its OWN trigger text, so the stranded message was never seen by anything (its ingestion is the turn its delivery died before reaching). Left to run, the recovery spends a model call, posts nothing, marks the row `PROCESSED` and writes a closing line saying the loss ended. So the same predicate the path uses (`maxIncomingId`) is asked **before the claim** and the answer is `unrecoverable`: no attempt, no turn, and the row stays `DEAD` in the worklist. Only what the customer said counts — an away message or an operator's note moves the conversation forward without answering anything. The page has to **reach back** to the stranded message to answer at all: one unanchored page is the newest twenty, so twenty outgoing or activity messages since the strand would push a newer customer message off it and the check would read as "nothing newer". A page that never reaches back means the conversation moved more than a page since, and the row stays `DEAD`.
- **A route whose agent has no `ChatwootAgentBot` is refused**, and it is the one null this module must not pass on. `heldByAnotherParty` compares ids, so with `ourAgentBotId` null it cannot: the ownership gate goes loose and a conversation another AgentBot holds reads as ours. A live delivery never reaches that state — its `agentBotId` is the route token's bot, and the route exists because the bot does. Nothing could come of it anyway, since the reply is posted with the persona's token and a client built without one refuses the call by name (#79) — so the pass would spend a model call to post nothing and then report a recovery. Not narrowed to "another bot holds it": what is missing is the identity, not the comparison. An inbox bound to **nobody** is a different state and still runs the path, because that is what writes the operator's `no_agent` line (#318).
- **A recovery is armed only for a row something could rebuild a body from.** A legacy row whose ids are unrecorded is reported as a loss and is recoverable by nothing; armed, its job could only ever answer `unrecoverable`, and an upgrade's backfill produces those in bulk. One exported predicate answers it at the arming site and again after the claim.
- **Five outcomes, because the caller has five things to do**, and the two retrying ones take different roads. `deferred` (the conversation is BUSY) **reschedules**, which clears the failure budget: a turn is deliberately unbounded, so mapping it to `fail` would burn a second stranded message's whole ladder while the first message's turn was legitimately still running. Unbounded rescheduling is bounded anyway, by the one thing that does not depend on the conversation — the age ceiling turns the row `unrecoverable` and the job completes. `unreachable` (the account could not be read, or answered unusably) **fails**, which spends the budget, backs off and reaches the dead-letter line at the scheduler's cap: a durable condition an operator has to fix has to be given up on and *said*. `recovered`, `superseded` and `unrecoverable` complete the job.

## What this uncovered on the way

`message_type` arrives as an **integer** over REST and as the enum **string** on the wire, and `normalize.ts` read only the string. An event rebuilt from a REST read therefore normalized to `messageType: null`, `isNewIncomingMessage` answered false, and the customer's message was classified as not-incoming — silently, with no throw and no line: the exact silence #228 had just finished removing, reintroduced by its own recovery. `messageTypeOf` now reads both, in one place, and `messages.ts`'s private copy of the same question is gone. The coercion is closed to digits, so a string that is not a number cannot round-trip into a valid type.

## Validation scope

**Proven by test** (`chatwoot-recover-delivery.test.ts`, `chatwoot-recover-payload.test.ts`, `chatwoot-payload-shape.test.ts`, `delivery-sweep.test.ts`, `scheduler-lanes.test.ts`, `terminal-failure-fence.test.ts`), DB-backed against a real Postgres, with a stub Chatwoot driven through `deps.makeClient` — the seam every other caller uses, so the REST reads are exercised for real up to the socket:

- the customer gets the reply the strand owed them, end to end: `recovered`, the ledger row `PROCESSED`, `attempts` at 1, and the message on the conversation;
- a conversation the mirror still calls `resolved` is answered anyway, and the mirror is **repaired** rather than bypassed;
- a live snapshot that cannot be trusted defers and spends no attempt;
- the route is rebuilt from the live message, so a mirror that never learned its inbox is still answered — and the bot identity that comes with it is what keeps the ownership gate from going loose and answering over another bot;
- a rebuild that can name no inbox at all, and one that comes back as anything but a new incoming message, leave the row `DEAD`;
- a customer who wrote again leaves the row `DEAD` with no turn built, no attempt spent and no closing line, while a newer OUTGOING message does not block the recovery; a newest page that never reaches back to the stranded message refuses, and an empty one is treated as a degraded read instead;
- a route whose agent has no persona bot is refused with no turn built — measured before the fence existed, the turn ran and the stub accepted a post over another bot's conversation; against a real client nothing would have been sent, because the persona's token is what posts and a client built without one refuses by name (#79), so the real cost of passing the null on is a model call spent to post nothing and a recovery reported anyway — while an inbox bound to nobody still runs the path;
- the gates run: a conversation another bot holds is not answered, and **no turn is even built**;
- the turn fence is asked of both keys a turn claims: a turn already running on the conversation defers before either REST read, and a turn holding the GRAPH thread with this process's registry emptied — a second replica, built with the real writer — defers after them; a conversation moved to a different contact inbox DURING the REST reads defers on the thread it is on now, not the one that was loaded;
- the recovered message advances `lastInboundAt` even when the conversation's own state has moved past it, and the mirror's own suite pins the rule on both sides: a stale incoming message ahead of the watermark moves it, one behind it does not drag it back;
- a turn that answered NOBODY does not close the loss, and the rule is a POSITIVE list — `TURN_SETTLED`: `posted`, `taken-over` and `blocked`, where SETTLED is deliberately not the same as answered and two of the three send the customer nothing rather than a growing list of failures: `empty` — the turn ran every gate, reached the end and delivered nothing — used to mark the row PROCESSED and write `recovered` over a customer still waiting, and `no-agent` and `agent-unavailable` did the same. A human who takes over DURING the turn is one non-`posted` close and has its own test; a guardrail tripping with `action: "silent"` is the other, and the line between it and `empty` is WHO decided — a policy the operator configured, which a re-run reproduces, versus a model that ran dry, which a second attempt could answer differently. Driven through the real guardrail, which is what surfaced a missing `sendPrivateNote` in the stub;
- a turn that THREW does not close the loss: nothing is sent, the row goes back to `DEAD` with the attempt counted, no closing line is written, and the outcome is `unreachable` so the job backs off;
- a reply WITHHELD to a message that landed between the recovery's own freshness read and `shouldPost` does not close it either: the row goes back to `DEAD` and the job completes, and the test proves the turn really reached `shouldPost` by counting the unanchored reads;
- a `/reset` typed at a PRODUCTION agent is answered like the customer message it is, while the same text at a TEST-mode agent is refused: whether a control command exists at all is the agent's mode, and the delivery path decides it the same way;
- an inbox REBOUND to another persona mid-recovery answers as the persona it is bound to now, closing line included;
- the turn's FAILURE is reported by the turn and not by the block around it, which also wraps the bookkeeping that follows it — from there, "the turn failed" and "the turn answered and a write after it failed" arrive as one event, and the second reading costs a duplicate reply and a second run of every side-effecting tool. Not reachable today (all three of those calls swallow their own failures), which is the reason it is bound by construction rather than waived. Asserted structurally, and mutating the report back to the enclosing catch kills it;
- nothing AWAITS between the fence answering free and the mark that holds it, nothing awaits between the MIRROR RE-READ and that fence either, and the local keys are asked AGAIN after the durable read — all three asserted structurally, because what is being pinned is the absence of a suspension point and the presence of a synchronous re-ask, and a behavioural test would have to inject the very await it forbids. The second one was a real gap: `agentBotChatwootId` sat between the re-read and the fence, one scoped query wide, and `mirror.ts` writes `contactInboxId` on an unversioned event — so a webhook landing inside it left the body and the fence naming the thread the contact HAD. The query is asked above the re-read now, and reverting the move fails that test and only that test;
- a conversation RESOLVED mid-rescue is seen: the body states the MIRROR ROW, not the live snapshot read before the two message reads;
- an operator who RESOLVES between the recovery's last mirror read and the gate is not answered over: the resolve is awaited from inside the fence's own transaction, which is the one seam in that stretch, and nothing is sent and no turn is built;
- a handoff landing DURING the two message reads is not undone by the reconcile: the live snapshot is applied in the statement after the GET, the way the other two callers of `reconcileMirrorFromLive` already do it. Measured before the change — the reconcile's fallback ordering compares `last_activity_at`, which a handoff never advances, so the older bot-owned snapshot walked the mirror back and the reply went out over the human;
- a HUMAN who takes the conversation in that same window is not answered over either, and this one is not the resolve's rule: a message may never write the assignee at all, so the mirror correctly stays human-owned while the rebuilt payload goes on stating the pre-handoff trio. What the test asserts is that **no turn is built** — measured before the fix, the turn ran, the model was invoked and every tool it chose ran, and only the runtime's own ownership re-check (which runs AFTER the model call) withheld the text. The reply was the smaller half of the damage. The rule is a pure function with a decision table (`effectiveAssignee`, `normalize.ts`): whichever witness says the conversation is HELD is the one believed, because that reading can only cost silence while the other costs an answer posted over somebody;
- the line that explains a closed gate names the reading the gate DECIDED on, in both places it is written — the flow-log row an operator opens after they suspect something, and the console line they are reading before they do. Asserted separately because they are written from separate arguments: reported from the payload, a conversation the mirror had already resolved was recorded as `ownership_lost` at `pending`;
- the contact's identity is LEFT where the mirror holds it, and the test fails the moment a rebuilt body starts stating it;
- a route the mirror never learned reaches the GATES and not just the turn: with the agent in test mode and the conversation never activated, nothing is sent and no turn is built — measured before the repair, the customer got a reply from a test-mode agent;
- a reset landing mid-recovery is told the thread is BEING WRITTEN: the recovery holds the GRAPH key across the handoff too, which is the key `threadBusyForResetOn` reads. Held as a RESERVATION rather than as an invoke, and the difference is one reader: every writer asks `isTurnInFlight`, which counts both kinds, while `markTurnOwning` asks `isTurnRunning`, which counts invokes only — it is answering whether ANOTHER invoke was already reading, and counting this hold answered that about the caller itself, so `claimAttendanceBoundary` deferred the divider AND the marker for the very turn that asked and a recovery opening a new conversation on a reused contact thread ran the model against the previous attendance with nothing between them. Pinned by a claim-level test, and two mutations kill what they should: counting reservations in `markTurnOwning` again fails that test alone, and blinding `isTurnInFlight` to reservations fails it plus the three recovery tests that need the hold visible. Held only on the conversation key, a reset in that window saw an idle thread, cleared the memory and the checkpoint, and the recovery then ran the turn that restored them — measured `[false, false, true]` at the three client builds, and the middle one is the window;
- the conversation key is HELD from the fence to the handoff, not just probed: the three client builds read `[false, true, true]`, and the middle one — the delivery path's own, before the turn marks anything — is the one that flips;
- a message the REST read returns with NO `created_at` is refused as a degraded read: no reply, no attempt spent, and the inbound anchor unmoved;
- a re-entry that lands mid-rescue keeps its pairing in BOTH readers — the mirror row, and the follow-up ladder's own payload, which takes the episode off the event and has no version to reject a stale one with;
- the compensating write gives three answers and swallows none: a row still in the state it was left in comes back to `DEAD`, one that MOVED is left alone, and a write that keeps failing is retried and then reported;
- the rebuilt body carries the contact inbox the conversation is on NOW: a pairing moved during the REST reads is not put back by the mirror write the body drives — measured on a fixture built to reach the branch that ASSIGNS it, because the stale branch writes no pairing at all and would have held either way;
- a burst stranded together is answered ONCE, and the older row keeps its page: two `DEAD` rows on one conversation produce one reply, one closing line, and an older row still `DEAD` at zero attempts;
- a control command is not replayed, while a message that merely mentions one is recovered normally;
- the age ceiling on both clocks, including a message older than the ceiling whose ledger row is young;
- the redirect episode survives the rescue, with its version, instead of being cleared by it;
- a turn that answered NOBODY keeps the row, and a human who takes over DURING the turn does not: the positive list is asserted from both sides;
- the loss is closed by a line of its own, at `warn`, naming the conversation, the message and the agent;
- every refusal: a row that is not `DEAD`, one naming no message, one past the attempt ceiling (and its neighbour one below it, so the comparison is a boundary), a live turn, a second recovery on the same conversation, a conversation the mirror never knew, a message Chatwoot no longer has, an unreachable account, and a lost claim race;
- the sweep arms a recovery for exactly the rows it declares lost, with the row id in the payload;
- the job's mapping: `deferred` reschedules, `unreachable` fails, everything else completes; a payload naming no row completes rather than burning five attempts.

**A third finding was refuted as pre-existing rather than introduced.** A round asked for a claim generation so a merely STALLED original handler cannot reach the delivery path's final settlement — which updates by id with no CAS — and write `PROCESSED` over a row a failed recovery had just restored to `DEAD`. The overwrite is real. This PR does not create it: with or without a recovery, a stalled handler waking up and settling its own row takes the message off the `WHERE status = 'DEAD'` list the same way, so the recovery adds an actor rather than an outcome. And the missing CAS is a decision with its own written argument on the other side — a turn that outlives the sweep's threshold and then completes DID deliver, late, and winning there is what leaves the row saying the true thing. A generation counter would have to tell "completed late" from "woke up and failed", which the delivery path does not answer about itself; that is its own piece of work, on the live path. The bound is recorded beside the compensating write.

**A second finding was refuted at the source.** A round read the message's `inbox_id` as a HISTORICAL route — a conversation transferred between inboxes after the strand would be answered on the inbox it had left, with the wrong persona's gates. That transfer does not exist: measured at the fork, `Conversation#inbox_id` is `not null`, written once by `ConversationBuilder` from the contact inbox, and nothing in `app/`, `enterprise/` or `lib/` updates it — `clone_inbox` creates new conversations on the destination rather than moving any. What Chatwoot transfers is the ASSIGNEE, agent or team, a different column and a gate this module already runs. So the two readings cannot name different inboxes for one conversation, and no guard was added for a state the source cannot produce; the measurement is recorded where the route is chosen.

**One review finding was implemented, measured, and reverted, and it is the clearest case in this PR of why a plausible fix needs a measurement.** A round asked for the live conversation's `meta.sender` to be carried into the rebuilt body: a stranded message may be the very event that would have refreshed the contact, and `authorizeContact` reads the STORED identity on a gate that fails closed. The premise checks out — measured at the fork, `conversations#show` renders the full contact partial, `phone_number`, `email` and `identifier` included — so the identity really is available and really is current.

It still may not travel, and the reason is the POSITION rather than the value. The mirror positions identity per field by the payload's `last_activity_at`, and a rebuilt body's is the stranded message's own clock, chosen so a rescue does not stamp the customer's words with the rescue's hour. Carrying a value read now under a clock from then is the "source position, never a receipt time" the mirror forbids, and it does not merely fail to help: implemented and measured, with the contact positioned at that same second by a sibling of its burst and the live phone now different, the tie rule empties the field. The contact then reads as `no_identity` at exactly the gate the round wanted to protect, and the stored phone is gone. One payload has one clock and it also orders status, assignee and the inbound watermark, so there is no second position to carry identity on. The omission is argued in place now instead of listed, and a test pins it.

**Proven live, against the local Chatwoot fork (4.16.0), in both halves.** Conversation `106` on a throwaway API inbox `25`, one incoming customer message `422` ("oi, alguem pode me ajudar com o horario de amanha?"), and a `ChatwootWebhookDelivery` row left `PROCESSING` with `claimedAt` forty minutes old — the state a delivery is in when the process dies between tx1 and tx2. The same script ran on both trees, so the comparison is between two codes and not two scenarios.

- **On the base** (`7037fd60`, a detached worktree so nobody's working copy was touched): the sweep moved the row `PROCESSING → DEAD` and wrote its loss line — `the customer's message 422 on conversation 106 was never answered` — and `jobs armed: []`. Read back from the CHATWOOT database, the conversation held exactly one message, the customer's. Nobody answered.
- **With the fix**: the same sweep armed `DELIVERY_RECOVERY` with `dedupeKey: delivery-recovery:186`, the job ran the delivery path, the model was invoked for real (Gemini, through the tenant's vault credential), and the reply posted. Read back from the CHATWOOT database: message `442`, `message_type: outgoing`, `sender: AgentBot/19`, content *"Olá! Claro, qual seria o melhor horário para você amanhã, sexta-feira?"*, conversation still `pending`. The ledger row ended `PROCESSED` with `attempts: 1`, and the closing line read `a turn answered 1 stranded delivery on conversation 106`.

Three intermediate states were measured on the way and are worth naming, because each is a real refusal doing its job rather than a fixture problem: with the conversation moved to `open` by Chatwoot the gate refused and no turn ran; with nineteen activity messages above the trigger id `shouldPost` answered `superseded` and the row went back to `DEAD` with the loss NOT closed (which is round 25's rule, observed live); and with the handled watermark already at `422` the CAS refused a second answer, which is the duplicate-reply fence the direct turn has carried since issue #49.

Everything was torn down afterwards: the throwaway tenant dropped, and the Chatwoot inbox, bot, contact, conversation and messages deleted.

**Mutation-proven.** Every guard in the service and the job was removed or inverted one at a time, across all ten review rounds, and each mutation takes at least one test with it. Coverage gaps found and closed this way: the attempt ceiling read as a direction rather than a boundary, the derived bot identity, the lost-race outcome, the re-entry claim (a test that had been passing for the wrong reason), and the closing line's agent attribution, the inbound watermark the mirror's stale branch was dropping, the loss closed on a turn that threw, and the loss closed on a reply the turn withheld, a control command refused where no command exists, a persona read from a binding that had moved, a redirect episode re-armed from a pairing that had been replaced, a query left sitting between the fence and the mark meant to hold it, a route the mirror never learned that reached the turn while skipping every gate, a fence whose local half was asked only BEFORE its own durable read, and a replayed message that walked a later resolve back to `pending` and was then answered over. The last one took two rounds and cost a rule in two places: a reopen is now ordered against the stored status mark at whole seconds (`state-order.ts`), and the gate reads the status the mirror settled on rather than the one the payload proposed (`webhook.ts`) — a payload can be applied and still have its status refused. Both are the live path's rules, not this module's, and both were reached from the recovery because that is where the window is widest. The assignee half of that same pair came next and did not follow from it: the gate preferred the payload wherever it spoke, and a payload always speaks, so a human taking the conversation in that window got a full turn run against them — tools included, since the runtime's ownership re-check runs after the model call. `effectiveAssignee` decides it now, and it does not pick by recency, because neither reading is uniformly newer: it picks by which way a wrong answer fails. The state-reading finding came back a second time with a corrected mechanism and was acted on: the first round said the body carried the live snapshot, which was wrong, and the second said it carried a reading captured before the route queries, which was right. The body now states the later of the two mirror readings, and the test pins what CAN be measured — that neither of them is the snapshot. A second guard was written and reverted this way: a round asked for the conversation's state to be re-read before the handoff, and reverting it changed no test — because the body already states `reconciled.state`, the row after the reconcile, not the live snapshot the round read it as. The test built for it was kept, repointed at the rule that IS load-bearing, and it dies when the body is made to state the raw snapshot. One guard was written and then REMOVED rather than kept: a review round asked the mirror's stale branch to require a measured timestamp, and the case built for it showed an undated event never reaches that branch — it applies, because there is nothing to order it by. The harm was real and the guard belongs where the undated body is built, so the recovery refuses to build one. One mutation survived its first fixture and was not left standing: reverting the body's contact inbox to the loaded one changed nothing while the fixture landed in the mirror's stale branch, which writes no pairing at all, so the fixture was rebuilt to reach the branch that assigns it. One mutation needed a second test before it died: inverting the monotonic comparison on that watermark changed nothing while the only fixture reaching it had a NULL stored value, so the guard got a case with a value stored. One mutation survives and is left standing on purpose: the inbox **name** carried in the rebuilt body is written back onto the very row it was read from, so nothing observable turns on it — it is carried because the rebuild reproduces the body, and that is stated where it is passed.

**Proven live**, against a local Chatwoot fork (4.16.0) with a real conversation of 30 messages, cleaned up afterwards:

- `before: id + 1` returns the page **ending at** the stranded message and containing it; the default page (20 most recent) does not contain it at all. The anchor is load-bearing, not defensive.
- `message_type` is the integer `0` over REST, confirming the divergence above at the source (`_message.json.jbuilder` renders `message_type_before_type_cast`).
- a contact **sender** carries `type: "contact"` over REST and no `type` key at all on the wire (`push_event_data` vs `webhook_data`). This corrected a fixture in this PR that had been asserting the wrong wire shape.
- **attachments do not diverge**: both views call `attachments.map(&:push_event_data)`, the same method, so the voice note a recovery hands the eager-STT pass is the one a live delivery handed it. Re-checked at the fork when a review round argued otherwise: `base_data` puts the whole `meta` hash on every attachment and `audio_metadata` ALSO lifts `transcribed_text` to the top level, which is why `messages.ts` reads it from `meta` and `normalize.ts` reads it from the top and both are right about the same payload. The partial that does drop it, `_attachment.json.jbuilder`, enumerates its fields by hand — and the messages index does not use it.
- an incoming message on a `resolved` conversation **reopens** it, which is the measurement that refuted the mirror-only design.
- the REST conversation show carries **no** `redirect_origin_display_id` and no `contact_inbox`, which is why those two fields still come from the mirror.

Read at the fork's source rather than live: `api/v1/models/_message.json.jbuilder` renders `json.inbox_id` on every message the messages index serializes, and `Message#webhook_data` builds the wire's `inbox` object from that same association — which is what makes the message the faithful source for the route.

**Not validated:**

- **No end-to-end run against a live Chatwoot of a real strand being recovered.** Producing one means killing the process between two CAS points on a live account; the state that produces was measured separately in #282, and everything downstream of it is exercised here against the real `processChatwootDelivery` with a stub account.
- **Rows already `DEAD` before this ships are not recovered**, including rows still inside the six-hour ceiling at the moment of the deploy. The arming happens at the moment of death, and the sweep's query never looks at a `DEAD` row again, so those stay in the worklist, which is where they were. A backfill would reach them and is deliberately not here: it would arm a whole backlog in one pass, each row a model call and a reply to a real customer, and the PER-ROW cost is bounded by the attempt ceiling while the aggregate is the one thing about this feature nobody has measured. The trigger is also not obvious — boot runs on every replica of every deploy and would need a fence of its own, a migration cannot enqueue application work, and an operator-invoked command is the only one that lets somebody look at the count first. A review round asked for it; the reasoning now lives beside the arming site rather than only here, because a PR body does not survive the PR.
- **The last window between the fence and the turn is not closed**, and cannot be by a fence: the durable claim COUNTS turns rather than refusing a second one, deliberately, because two deliveries for one conversation really do overlap whenever debounce is off. What bounds it here is the newest-message check instead — a live delivery is running because the customer wrote again, which is already the case this recovery answers `unrecoverable` to.
- **A recovery given up on does not reach the scheduler's dead-letter line.** `MAX_RECOVERY_ATTEMPTS` (3) is lower than the scheduler's five, so a turn that keeps throwing exhausts the ROW first and the job then completes. The record is a `warn` line saying so plus the row itself, still `DEAD` on the page the sweep already opened. The dead-letter line remains the outcome for an unreachable account, which refuses before the claim and spends no attempt.
- **The bot identity is resolved before the fence and again by the delivery path.** An operator rebinding the inbox in the queries between them leaves the ownership checks comparing against one persona while the turn loads as another. The same shape as the mode read above, with the same reason for leaving it: the alternative is a second reading against a possibly different agent, which is the defect round 14 fixed.
- **An operator who flips the agent from production to test mid-rescue can have a stranded `/reset` replayed.** The mode is read twice: here, in one transaction with the inbox BINDING (because a mode read against a different agent than the one that answers refuses the wrong message), and again by the delivery path for itself. Closing the gap means widening `processChatwootDelivery`'s contract for every caller to cover a window a handful of queries wide, entered only when the stranded message is exactly a command, the original process already executed it, and the mode changes inside it.
- **A follow-up nudge on ANOTHER replica can still overlap the recovery's turn.** The hold across the handoff is the in-process registry, which is the key `followUpHandler` reads; the durable claim would not help, because it counts turns rather than refusing a second one. Safe under the single-replica / one-leader invariant this repo already runs on, and the same shape as #203's remaining edge.
- **A `/reset` served by ANOTHER web replica is not fenced either, and it is the same edge read from the other side.** Both marks held across the handoff live in a Map in this process, and the claim that crosses replicas is not taken until `runAgentTurn` takes it — so a reset on a second replica reads neither, clears the memory and the checkpoint, and this turn restores them. Closing it means taking the durable claim (`markTurnOwning`) at the mark instead, which is not a wider version of the hold: that claim WAITS on an append's lease and on the reset's row lock, so it changes what the fence means rather than extending its reach. Left as #203's remaining edge, on the same single-replica invariant the bullet above stands on, and recorded at the mark rather than only here.
- **The contact-authorization gate is asked about the identity the MIRROR holds, and a recovery cannot refresh it.** The gate always reads the stored `Contact` and never the event (`contact-auth/service.ts`, which says so in place); what a live delivery adds is mirroring that message's `meta.sender` first. A rebuilt body states no identity, deliberately — under the stranded message's own clock it would meet the tie rule and CLEAR the field, which was implemented and measured earlier in this PR — so the field rule applies as written and the stored value stands. The direction of the error is not uniform: an identity ADDED since leaves the mirror empty and the gate answers `no_identity`, which is no turn, and the gate is opt-in (`enabled: false`). The one case a recovery answers where a live delivery would not is an identity REPLACED, old value authorized and new one not, on a strand where the stranded message is itself the event that carried the change and nothing since has carried identity. Closing it means deciding on the live contact the recovery already reads without writing it, and that is not free: the REST and webhook renderings of a contact are known to differ in shape (a fixture in this PR was corrected on exactly that), so a comparison wrong about formatting would refuse every recovery for that contact and say nothing. Recorded in `docs/contact-auth.md` rather than closed, and waived in the local review adjudications.
- **A split reply that fails mid-balloon re-sends the balloons that already landed**, and this is the live path's behaviour rather than the recovery's. `deliverReply` calls `sendMessage` per chunk with no catch, so a failure on the second of three escapes after the first is in the conversation; the turn rejects, and a Chatwoot webhook retry composes the whole reply again exactly as a recovery's next attempt does. This PR adds a way to reach the existing retry, not the duplication. Answering it means deciding whether the unit of delivery is the reply or the balloon, which is the split module's question and needs per-chunk durable state nothing writes today.
- **Claiming from `DEAD` revokes nothing.** A handler the sweep called abandoned holds no lock this side can take away, so a handler that was merely stalled can resume and finish whatever it had started. The overlap that matters, two invokes, IS fenced: a handler inside `runAgentTurn` holds the thread's durable claim (#203) and the recovery asks the row for it. What is left is a handler stalled for half an hour BEFORE its turn — every await between its claim and `runAgentTurn` is a scoped query or a REST read with its own deadline — and there the two turns overlap the way two live deliveries for one conversation already can. The standing cost that leaves is the one stated at the head of the module: a recovery re-runs a turn that may call side-effecting tools.
- **A burst stranded together loses the older messages' TEXT.** Two customer messages stranded on one conversation are two `DEAD` rows; the newest one's recovery answers the conversation once and the older rows stay `DEAD` on the operator's page, but a direct turn carries its own trigger text and nothing back-fills the channel, so the model answers the last question and never reads the earlier ones. Measured, and pinned by a test rather than assumed. With coalescing on it does not arise (the flush re-reads the conversation's messages). Gathering the burst here would be this module re-implementing the flush, which its design refuses to do, and the ordering question it opens is its own.
- **No measurement of what a recovery costs at fleet scale.** One sweep pass can declare a batch's worth of rows lost, and each one becomes a turn; the shared lane's provider concurrency is what caps them, and that cap was not exercised under a real backlog.
- **An install whose `serviceWindow.windowHours` is set below six hours** can still produce a recovery outside its own window. The ceiling bounds the common case; it does not read that setting.


gabrieljablonski:
@coderabbitai review

gabrieljablonski:
@coderabbitai pause

Issue #316, opened 2026-08-25
Four of ten billed paths never write `LlmUsage`: guardrails omits the callbacks, and vision's provider contract cannot carry the usage it receives


gabrieljablonski:
## What happens

`LlmUsage` is the only ledger of what an install spends on models, and it is written from exactly one place: `UsageCapture` (`src/graph/usage.ts`), a LangChain callback handler built by `buildCallbacks` (`src/graph/prepare.ts`). A billed call reaches the ledger only if it is (a) a LangChain invocation and (b) handed those callbacks.

Measured on `main` (`404677d0`), every path that spends provider money:

| path | how it calls the provider | writes `LlmUsage` |
|---|---|---|
| agent turn (`runtime.ts`), nudge (`nudge.ts`), memory compaction (`compact.ts`), TTS normalize (`prepare.ts`), playground (`playground/service.ts`) | LangChain **with** callbacks | yes |
| attendance summary (`memory/summarize.ts`) | LangChain, callbacks threaded through the parameter | yes |
| **guardrails** (`invokeForVerdict`, `src/modules/guardrails/analyze.ts`) | LangChain, **no callbacks** | **no** |
| **vision** (`src/modules/vision/providers.ts`) | raw `fetch` to `/chat/completions`, `/messages`, `:generateContent` | **no** |
| STT (`src/modules/stt/providers.ts`) | raw `fetch` | no, and see below |
| TTS synthesis (`src/modules/tts/providers.ts`) | raw `fetch` | no, and see below |

Four paths are outside the ledger. Two of them are token-denominated and are the subject of this issue; two are billed in other units and are deliberately not (last section).

## Where it comes from

**Guardrails omits the callbacks.** `invokeForVerdict` calls `model.invoke(messages, { signal })` and `.withStructuredOutput(...).invoke(messages, { signal })`. Neither passes `callbacks`. `grep -n 'callbacks\|BaseCallbackHandler' src/modules/guardrails/*.ts` returns nothing: the module has never had the concept. The model is built by `createChatModel` in `gate.ts` like any other, so the tokens are counted by the provider and billed; they just have nowhere to land.

This is the hot path, not an edge: guardrails analyse customer-facing text per reply, so an install with input and output moderation on spends one or two model calls **per customer turn** that the dashboard cannot see.

**Vision cannot carry the usage it already receives.** `VisionProvider.extract` is typed `Promise<string>`, so the return type makes usage unrepresentable, and all three branches parse the text and throw the rest away:

[...]

PR #390, opened 2026-08-27
Refuse a turn once the month's tokens are spent, instead of learning it from the invoice

### LINKED ISSUE #146: Nothing bounds what a tenant can spend on the model: usage is recorded after the call, never checked before it
Nothing in the system can refuse a model call because a tenant has spent enough. Usage is measured accurately and only after the fact, so the first time an operator learns that a month went wrong is on the dashboard, or on the provider's invoice.

## What exists today

`src/graph/usage.ts` captures usage at the source: a LangChain callback writes one append-only `LlmUsage` row per model invocation, with the provider's own token counts, attributed to tenant / agent / conversation / inbox, separating `inbox` traffic from `playground`. It is deliberately best-effort and never throws into the reply path. The `LlmUsage` model is indexed on `(tenantId, createdAt)`.

That is a good ledger, and it is only a ledger. Every reader of it is a projection: the dashboard's daily series and per-model break-down. No code path consults it before deciding to run a turn.

The word "budget" does appear in the runtime, but never in this sense. `src/graph/history-window.ts` bounds how much history travels in one turn, and `src/graph/graph.ts` bounds how many tool calls a single turn may make (falling back to the raw model when the allowance runs out). Both shape one turn. Neither accumulates across turns, and neither knows that a turn costs anything.

So the only thing standing between an install and an unbounded provider bill is the provider key's own limit, if the operator set one there. For a self-hosted deployment, where the operator brings their own key, that is the whole of the protection.

## Why this is worth having

The gap is not exotic. Consumption is driven by inbound message volume on channels the operator does not control the input to, and every knob that makes an agent better — a longer history window, more tools per turn, a larger model — multiplies the per-turn cost without any counterpart that says "stop". An install that behaves normally for months and then has one bad week has nothing in the product that notices.

A ceiling also gives the dashboard's numbers something to be measured against. Right now the usage figures answer "what happened"; nobody can express "what is the most this should be".

## The shape that fits what is already here

Count tokens, from `LlmUsage`, locally.

The alternative — a ceiling denominated in currency — looks more natural and is the wrong dependency. Cost in this codebase comes from Langfuse (`src/modules/analytics/langfuse-costs.ts`), which is optional and asynchronous: `src/client/pages/DashboardPage.tsx` already accounts for a zero-cost series meaning "ingestion lag, or model pricing not configured" rather than "spent nothing". An enforcement gate cannot be built on a number that is optional, lagging, and legitimately zero. Tokens are written synchronously by our own callback and are queryable on an index that already exists.

That makes the deliverable a per-tenant token ceiling over a window, evaluated before a turn runs, with the refusal recorded where refusals are already visible (flowlog, and the alert channels, which is what makes it useful before the ceiling is reached rather than after).

## The decision to settle before code

What the customer experiences at the ceiling. The three candidates are silence, a configured message, and handoff to a human, and they are not interchangeable: silence looks like an outage to the person waiting, a canned message is a message the operator is paying nothing for but is still a bad look, and handoff is right only for installs that have humans to hand off to. A per-tenant setting with a sane default is probably the answer, but the default matters more than the setting, and picking it is part of this issue rather than of the review.

The window (calendar month vs rolling) and whether the ceiling is also expressible per agent are smaller, and can follow the same reasoning: the cheapest thing that is honest about what it enforces.

gabrieljablonski:
Fixes #146.

`LlmUsage` is an accurate ledger, but it is written **after** each call and every reader of it is a projection. Nothing in the system could refuse a model call because a tenant had already spent enough, so the first time an operator learned a month had gone wrong was the dashboard, or the invoice. This adds the gate that ledger was missing: per tenant, over the calendar month, asked before the call.

## The design question the issue named, answered

@JotaSXBR's comment lays out three options and says the choice is not obvious. It isn't, so here is the reasoning rather than just the result.

**Denominated in tokens, not currency.** Cost in this codebase comes from Langfuse (`src/modules/analytics/langfuse-costs.ts`). It is optional, asynchronous, and legitimately zero on an install that never configured pricing — which is why the dashboard already reads a zero-cost series as "ingestion lag or no pricing" rather than "spent nothing". A gate cannot be built on a number that is optional, lagging, and legitimately zero, and making the cap depend on an optional integration turns the spending guard off for exactly the self-hosted installs that have nothing else standing between the curve and the invoice. A local price table was the third option and it is wrong the day a provider changes pricing, silently, in the direction that matters.

Tokens are written by our own callback on every invocation and summable on an index that already exists. Harder to explain to whoever signs the invoice, and the only one of the three that is always answerable.

**The sum is `prompt + completion`, and cached reads are not added.** A cached read is a discounted subset of the prompt tokens, so adding the column on top would count the same token twice and move the ceiling by however much the provider cache happened to serve — 43% of prompt tokens on the window in the issue, so the error would not have been small.

**Two ceilings, not one.** `source` already separates `inbox` from `playground`, and the issue is right that they should be treated differently. Customer traffic is driven by how many people write in, the variable the operator does not control; the playground is one person testing a prompt in a loop. A single number would let the second silence the agent for the first. `0` on either half means no ceiling on that half, never "refuse everything".

**Calendar month, UTC.** It is the cycle the invoice follows. A rolling window measures consumption more honestly and never zeroes at once, which is the property that would have to be explained to whoever signs that invoice.

## What happens at the ceiling

On a customer conversation: the operator's configured sentence goes out **as the persona**, the conversation is **opened for humans**, and a **private note** names the numbers and says whether the handoff landed. Copy and note sit behind `noticeCooldownSeconds`; the verdict never does, so ten people writing in are each evaluated and each conversation is told once.

**The turn is asked about more than once, and the second ask owes the same three things.** The webhook's gate covers the message; the debounce flush runs minutes later and is where the turn actually spends, so a tenant can cross its ceiling inside that window from its own other conversations. A burst refused there is the FIRST refusal the conversation gets, so the flush posts the copy, hands off, and writes the note, in that order and under the same per-conversation cooldown key. It cannot be left to the customer's next message: with the handoff on (the default) the `open` is precisely what takes the conversation out of `pending`, so no later message of theirs reaches a gate at all. `announceSpendCeilingOnConversation` is that sequence in one place, called by both gates with each caller's own fenced primitives.

Below the ceiling and past `warnAtPercent`, the turn runs and a `spend_ceiling` warn reaches the alert channels, so the operator hears about it **before** the agent goes quiet. What that promises is that the first verdict landing at or past the fraction warns, not that the ceiling is never reached without a warning first: the fraction is evaluated by the gate on the ledger as it stood before the turn, so a single turn that spends more than the band takes a tenant from `allowed` straight to `over`. Same read-then-act bound as the overshoot, and the band is what buys the lead time. That warning is claimed once per six hours per (tenant, source) rather than per message: the alert bus coalesces a burst by bumping a `PENDING` delivery and inserts a fresh one as soon as the worker has sent the last, so a per-message warning would page the channels for the rest of the month about one unchanging fact.

There is deliberately **no team target**, unlike the per-agent contact-auth gate it is modelled on. A Chatwoot team id belongs to one account, contact-auth stores the account beside it precisely so a pinned team is ignored in the wrong one, and this ceiling is per tenant — a tenant spans as many Chatwoot accounts as it has instances, so one stored team would be meaningless in every account but one.

## Where the gate is asked, and where it deliberately is not

A ceiling is one question asked in several places, and the failure mode of that shape is a place that never asks. So the answer is written down per ledger `node` in `src/modules/spend-ceiling/coverage.ts`, and a fence compares its key set against `USAGE_NODE_IS_AGENT_TURN`: a node added to the ledger without an answer is a red test.

| node | gate |
| --- | --- |
| `agent` | the Chatwoot webhook (inbox), the **debounce flush**, the **re-engage** button, and `runPlaygroundTurn` (playground) |
| `nudge` | `runAgentNudge` and `runPlaygroundFollowup` |
| `vision` | `extractInboundFile` / `extractPlaygroundFile`, **after** the file is known to be readable and immediately before the provider call |
| `guardrail`, `tts_normalize` | covered by the unit above them |
| `memory_compact` | **ungated, by decision** |

**Every gate sits where the call is**, not where its function starts: after everything that would have stopped the call anyway — an agent that cannot run, a message that renders to nothing, a file this provider cannot read, a burst with nothing in it, a job `/reset` retired — and immediately before the call. A refusal says that spend was what stood in the way, so reporting one where the call was never going to happen sends the operator to raise a number that changes nothing. The one thing that does NOT follow the rule is the direction of a failed read: the ceiling fails **open** when the CEILING is unreadable, but once the verdict says `over`, every check after it is an escape hatch from it, and an unreadable escape hatch does not open.

**Vision asks for itself** because it runs on the incoming attachment before the webhook's gates decide anything — the same asymmetry #316 measured for attribution — so an image sent into a spent month would otherwise be billed with nothing to stop it. It announces the **warning** and not the refusal: the `over` line is written per refused message and vision runs on the very message the webhook gate refuses moments later, so a line from each would double the refusal rows and the alert bumps, and its own `vision` line already reports `skipped` with `spend_ceiling` as the reason. The warning leaves no such trace and the gate that would have said it may never run at all, so this is the only place that could.

**A source with no ceiling of its own never reads the ledger.** `0` is the operator saying that half is unbounded, and the answer comes from the configuration before any query is issued — otherwise enabling the block to bound the playground would put a ledger read on every inbox message for nothing.

**Guardrails deliberately do not**, and this is the entry worth arguing with. On the output direction the reply is already written and paid for, so refusing there either posts it unscreened or drops a reply the customer is waiting for; a ceiling that switched moderation off would let a budget decide a safety question. **Memory compaction** is out for a sharper reason, and it has a word of its own in `coverage.ts` (`ungated-by-decision`) because no gate answers for it: it runs from its own `MEMORY_COMPACT` scheduler job, minutes after the attendance it summarizes and on attendances a human handled, so there is no enclosing verdict to be "covered by". Refusing it does not save the tokens, it moves them — the raw history stays in the thread and the next turn carries it — so gating it would raise spend rather than bound it. The cost that buys is stated rather than hidden: a tenant past its ceiling keeps paying for compaction, bounded at one summary per attendance, and it is the one path on which "the ceiling bounds the month" is not literally true.

Customer-facing paths go quiet; the **playground throws** `429 errors.spendCeilingReached`, because the operator is looking at the screen and a turn that silently produced nothing reads as a broken provider. An `over-ceiling` nudge is a *repairable* refusal, so a follow-up occasion is not burned against a wall that is temporary by construction.

**An unreadable ceiling ALLOWS the call.** That is the opposite direction from the in-flight turn claim in #366, and deliberately: there the false answer let a writer erase a customer's message; here it would refuse a customer who is waiting because our own database hiccuped. The ledger keeps recording either way.

## The console

Resources → Advanced grows a card. The two bars come first, because nobody can pick a monthly token budget without seeing what the month has already cost, and they render whether or not a ceiling is set. `GET /v1/tenant-settings/spend-ceiling/usage` is the same number over REST.

## Validation scope

**Proved by test** (5 new files, 53 tests, plus the existing suites):

- the rule as a decision table, without a database (25 cases);
- the ledger read, DB-backed: the window, the two sources not borrowing from each other, and cached reads not double-counted;
- the webhook end to end through `processChatwootDelivery`: over the ceiling the model factory throws, so a turn that runs at all fails the test;
- the playground, the simulated follow-up and both vision entries;
- the fail-open, driven by rejecting the ledger sum alone;
- how often each line is announced;
- the coverage fence.

**21 mutations, each killed** (1 to 4 tests each): the vision gate removed, the playground gate removed, one ceiling instead of two, `>=` weakened to `>`, `0` read as "refuse everything", cached reads added to the sum, the month window ignored, the playground throw made silent, the fail-open turned into a rethrow, the warning said per message, the warning key losing its source, losing its tenant, and the `over` line throttled like the warning.

**Measured**: 1M ledger rows over 200 tenants and 90 days, a month's sum for one tenant runs in 1.3ms median on `llm_usage_tenant_id_created_at_idx`. Recorded in the code, along with the fixture that misleads: the same million rows under a single tenant take 40ms and a parallel seq scan, because an index that selects the whole table is worth nothing to the planner.

**Exercised live**, against a Chatwoot running on the same machine, three arms on one real conversation each, with the real Chatwoot client and a model factory that throws:

| arm | model built | conversation | what the customer saw |
| --- | --- | --- | --- |
| over (ceiling 1.000, spent 28.030) | no | `open` | the configured sentence, then the private note naming both numbers |
| warning (ceiling 32.000, 87,6%) | yes | `open` | the agent's answer; a `warn` line on the Logs page |
| under (ceiling 100.000.000) | yes | `open` | the agent's answer; no line at all |

The gate decided in 178-312ms including the two Chatwoot round trips. The run restored the tenant settings, the agent's schedule and the flow lines it wrote.

**Twenty-three review rounds on this PR.** Every finding was published before any code was touched, and the loop closed on a round that reported none. Eight adjudications were **refused** with the argument, recorded in `.codex-review-waived` so the local reviewer stops re-reporting what has already been answered, and argued in the commit that rejected each one. The rest became code, each with its own tests and mutations.

The first four rounds published nine findings. Eight became code, each with its own tests and mutations: the flush gate itself; playground warnings not reaching the alert channels; an over-ceiling nudge burning a follow-up occasion; a handoff written without the persona's bot token, so it raised before reaching Chatwoot and the burst was marked handled on a mute bot; that same handoff judging ownership before two database reads, so it could pull a conversation back out of a human's hands; the ledger being read on a source with no ceiling of its own; vision announcing a refusal the gate announces again; and the flush dropping a burst without the copy or the note it owed, which cost seven more mutations on its own.

Rounds nine through twelve are one lesson found four times: **a refusal that did not happen must not be reported, and a refusal that did happen must be reported once.** The fan-out reached the flow line as well as the acts; a message the other route had already answered was refused and logged; the announce sequence needed ordering (per conversation) and coalescing (per refusal) separated, because one key answering both silenced the second of two messages; the re-engage button reported a spent budget for a click that had nothing to answer; a route accepted a customer sentence its own service refuses, turning a 422 into a 500; the nudge occasion key collapsed independent jobs on one conversation, and then collided on its own delimiters; and the playground's documented 429 was declared on no route at all.

An eighth round found three more about how often the ceiling speaks: two concurrent deliveries of one message racing so that the handoff overtook the copy and the customer heard nothing; a nudge announcing its refusal on all eight retries, paging the alert channels eight times per pending job for one unchanging fact; and the announcement keying its month window off its own clock rather than the verdict's, so a rollover between the two reported the old month's figures under the new month's key. Its fourth finding was refused: a positive fallback for a malformed ceiling would invent a number nobody typed and silence an agent for real customers on corrupted data, which is the opposite direction from the unreadable-ledger rule; the comment claiming otherwise was the thing that was wrong.

A seventh round found the sharpest one: the flush's ceiling branch wrote four things (the copy, the handoff, the note, the watermark) without asking whether `/reset` had retired the burst, and ownership cannot answer that question because the reset hands the conversation back to the bot. Fenced once per write now, like the turn path's own `stillWanted`; five mutations, five killed.

A sixth round added two, both about the warning: vision never announcing it, so a message consumed before the webhook gate (human-owned, silenced, redirected, out of hours) crossed the fraction in silence; and the warning's six-hour window carrying no month, so a warning said at 23:30 suppressed the first one of the next month. Five mutations, five killed.

A fifth round added two more: a lenient settings reader that could return a value its own writer refuses, so an out-of-band write turned every save on the screen into a 422 about a field the operator never touched (fixed, four mutations); and the warning's promise being wider than its mechanism (a comment fix, above).

The ninth was **refused**: a reservation taken before each call, to close the overshoot when two turns read the ledger concurrently. That overshoot is real and is bounded by one turn per concurrent call, and it is written down under "What the ceiling does not promise" in `docs/spend-ceiling.md`. A reservation is a second count that would have to survive a crashed process, a provider timeout and a cost known only at the end, and every one of those failure modes points at refusing a tenant for tokens nobody spent.

Rounds thirteen through twenty-three are one family found eleven more times, and the lesson is the sweep rather than the site: **fix the question, not the line the finding names.** Three times a fix went in where it was reported and the same defect sat untouched two files away, and each of those cost a round. What went in: the flush announcing before asking whether `/reset` had retired the burst; the nudge asking `stillWanted` two database reads before it wrote; the monthly sum bounded below and not above, which stopped being harmless the moment the verdict began carrying its own instant; three gates asked before the thing that would have stopped the call anyway (an unreadable file, a missing agent, a message that renders to nothing, a burst with nothing in it); five occasion keys that could not tell two occasions apart (an inbound delivery, a follow-up episode, a redirect episode, and two built from account-local Chatwoot ids that collide across a tenant's second Chatwoot); a re-engage refusing a tail another delivery had just answered; a runnable probe whose thrown error read as "the agent cannot run" and let a turn spend past the ceiling; and `memory_compact` classified as covered by a gate that does not exist.

Eight findings were **refused**, each with the argument in the commit that rejected it: deferring the warning until every no-call exit has passed (a warning is a statement about the MONTH, true whether or not this call runs); moving the nudge's ceiling below the pending-interrupt check (the verdict there is true and operative, and the position suggested sits below Chatwoot writes); fencing the webhook's handoff against `/reset` (the customer's message was delivered and refused, and a spent budget is not undone by a clean slate — unlike the flush, where `/reset` durably retires the burst); making an inbound nudge refusal recoverable (Phase B is best-effort by contract, and a real retry needs a scheduler kind of its own, which would fix the throw case too); and moving the spend check behind contact authorization (both orderings can report a refusal the other would have made first, and the ceiling is one local read against a ten-second call to somebody else's endpoint).

**Not validated**: the alert-channel delivery of the `warning` (the flow line is asserted, the Discord/webhook hop is not); the six-hour warning window across a restart or a second replica, which is in-process by design and re-announces once; and STT, whose spend is not in this ledger at all, so no ceiling can be denominated for it.


JotaSXBR:
A measurement on the one entry in **Not validated** that has a number, plus one thing about it that I think is smaller than the entry implies.

First, the denomination: the reasoning here is right and I would not want it read as unresolved because I raised the alternative. Cost in this repo comes from Langfuse, and on this install `langfuse-costs.ts` answers `disabled`. A ceiling built on that number would have been off for exactly the installs that have nothing else standing between them and the invoice. Tokens is the correct unit.

## STT, sized

> **Not validated**: […] STT, whose spend is not in this ledger at all, so no ceiling can be denominated for it.

Read-only, one self-hosted production tenant, one WhatsApp inbox, 2026-08-01 → 2026-08-27 16:55Z.

The ledger over that window is **933 calls / 7,508,710 prompt / 370,583 completion**, and its per-model break-down names seven models:

```
~deepseek/deepseek-v4-flash-latest   452 calls
google/gemini-3.7-flash              305
gpt-5.6-luna                         124
gpt-5.4-mini                          28
openai/gpt-5-mini                      9
gpt-5-mini                             8
gemini-3.5-flash                       7
```

The flow log for stage `stt` over the same window holds **40 lines**: 34 `ok`, 6 `error` (all `STT openai failed with 400`), provider `openai` on every one, models `whisper-1` and then `gpt-transcribe`. Neither transcription model appears in the break-down above. Forty billed provider calls in the month the gate is about, and the ceiling cannot see one of them.

## Why, mechanically — and it is not "STT is not LangChain"

`recordDirectUsage` exists precisely for calls that do not pass through the graph, and vision uses it: `src/modules/vision/service.ts` calls it twice with `node: "vision"`, off `extracted.usage`. `src/modules/stt/service.ts` imports nothing from `@/graph/usage`.

It could not, as things stand, because the provider interface discards the number before the service could record it:

```ts
// src/modules/stt/providers.ts
transcribe(req: SttRequest): Promise<string>;
```

A bare string. There are five providers behind it and none of them can report what a transcription cost, so the gap is one line of an interface rather than a missing call site. `USAGE_NODE_IS_AGENT_TURN` being total over `agent | nudge | guardrail | tts_normalize | memory_compact | vision` is consistent with that: the fence in `billed-call-usage.test.ts` cannot flag the absence of a node that never reaches the map.

## The part I think the entry overstates

"No ceiling can be denominated for it" is true for `openai` and `elevenlabs`, which bill audio duration. It is not true for the whole family. The provider table in the same file lists `gemini` (`defaultModel: "gemini-3.5-flash"`), `openrouter`, and the generic Whisper entry pointed at Groq or a self-hosted endpoint. A tenant transcribing on `gemini` is spending **tokens**, in the same unit this ceiling counts, and the month still cannot see them.

So the honest statement is narrower and more useful: STT spend is invisible to the ceiling on every provider, and *undenominable* on only some of them.

## Why it belongs next to the vision gate rather than in a backlog

Vision and STT are the same leg. Both run on eager media, both before the ownership gate decides anything — the asymmetry #316 measured — and this PR gates vision for exactly that reason: "an image sent into a spent month would otherwise be billed with nothing to stop it." An audio sent into a spent month is billed with nothing to stop it either, on the same seam, in the same turn. On this tenant that is 40 occasions a month.

I am not asking for it in this PR; the scope here is already large and the coverage fence is the right way to have written it down. What would help the next reader is `docs/spend-ceiling.md` saying which of the two halves of eager media the ceiling holds, so "not validated" is not read as "wired but untested" — and, if the fence is meant to catch this class, that `stt` is a node the ledger does not have rather than one the map forgot.

*(Read at `1e2aac0`: `src/graph/usage.ts`, `src/modules/stt/service.ts`, `src/modules/stt/providers.ts`, `src/modules/vision/service.ts`. Tenant figures via the dashboard aggregation and the flow log, read-only.)*

PR #406, opened 2026-08-27
Let an agent stop asking once its endpoint has said yes

### LINKED ISSUE #189: Contact authorization: reuse a verdict across messages, and a way to revoke it
## Problem

The contact-authorization gate (#182) re-checks every incoming message. That is the right default: revocation on the operator's side is instantaneous and an unlock flow needs no special case. It also means the endpoint takes one request per incoming message, forever, for a fact that rarely changes.

Two operators already ask for the other shape:

- The check is expensive or rate-limited (a lookup against a core banking / ERP / ticketing API), and a burst of five WhatsApp messages is five identical calls.
- The gate is an unlock, not a lookup: the customer sends an access code once and should stay served afterwards, without the endpoint having to remember them.

## Proposal

An authorization mode on `agent.settings.contactAuth`:

- `perMessage` (today's behavior, stays the default).
- `once`: a positive verdict is stored per contact and reused, so the endpoint is called only until the first `authorized: true`.

Stored state needs a way back out, and that is the part that needs a decision before any code:

- **TTL**, the cheap answer: the verdict expires after N hours and the next message re-checks. Bounded staleness, no coordination.
- **Endpoint-driven**: the authorization response carries a revision/version that, when it changes, invalidates what is stored. Puts the operator in control without a second endpoint.
- **Inbound revocation**: a small authenticated route the operator calls to drop the verdict for one contact (or all). Instant, but a new public surface with its own auth story.
- **Automatic invalidation** on events that change who the contact is: identifier/phone/email changed in the mirror, contacts merged, agent config edited.

TTL plus automatic invalidation on identity change looks like the smallest useful combination, with the endpoint-driven revision as the opt-in for operators who need it exact.

## Open questions

- Where does the state live: on the contact row (survives conversations, matches "who may be served") or on the conversation (natural expiry, but re-asks every new conversation)?
- Does a denial get stored too, or only a grant? Storing denials makes the block sticky, which is precisely what an unlock flow must not do.
- How is the mode surfaced in the editor without making the common case (perMessage) look like a choice the operator has to study?

## Non-goals

- Revocation lists shared between agents or tenants.
- Any change to the per-message default.

gabrieljablonski:
The contact-authorization gate asks the operator's endpoint on **every** incoming message. That is the right default, and it is what lets `docs/contact-auth.md` promise that a revocation on the operator's side lands on the contact's very next message. It also means the endpoint takes one request per message forever, for a fact that rarely changes, and it means a gate used as an **unlock** ("send your access code to be served") asks again about a customer who already sent the code.

`contactAuth.mode = "once"` is the other shape. The first `authorized: true` is stored per contact and reused until it expires; everything else about the gate is unchanged.

```jsonc
"contactAuth": {
  "enabled": true,
  "url": "https://ops.example.com/authorize",
  "mode": "once",            // "perMessage" (default) | "once"
  "grantTtlSeconds": 86400   // 60 .. 2592000, clamped
}
```

## Where the reuse lives

Inside `authorizeContact`, the one function all four callers (webhook gate, debounce flush, proactive nudge, manual re-engage) already go through. No caller had to change, and none of them can disagree about when a stored verdict applies. A reused verdict is `allowed` like any other, so every caller still re-fences it against the conversation's owner.

## The whole rule, in one place

A stored grant is served **if and only if**, at the moment it is read:

1. it has not expired — time, and the operator chose the budget;
2. it matches the mirrored **identity** (phone, email, operator `identifier`);
3. it matches the **policy** (`url`, `credentialRef`, `includeMessageText`, `grantTtlSeconds`), whose fingerprint carries the **credential's own revision** — `credentialRef` is a stable id that survives a rotation and a deletion alike;
4. this process holds no **unconfirmed write** about that contact.

And it is REMOVED only by a refusal, or by the verdict that replaces it.

Rules 2 and 3 are MATCH rules, not revocations, and the distinction is load-bearing: a fingerprint that stops matching stops the grant being SERVED, and a value put back matches again. So nudging a policy field is **not** a way to clear stored verdicts, and the PR says so in all four places it used to claim otherwise.

**A refusal is never stored.** A stored denial makes an unlock permanent: the customer sends the code and the gate answers with a verdict from before they sent it. An endpoint **error** is not stored either — transient by contract. What a fresh refusal DOES do is drop whatever was stored, under **every** mode: only `once` grants, and grants outlive a switch back to `perMessage`, so a refusal arriving while the reuse is off still has to reach them.

**There is no clear-everything lever, deliberately.** What ends reuse is the TTL, the identity moving, a refusal, and `mode: "perMessage"` — the last of which is immediate and complete while it is off. A purge would be a new operator surface, and #189 scoped the way out as the TTL plus invalidation on identity change.

## Ordering, and where it ends

Two messages from one contact are two questions under the unlock flow, so their checks race. The verdicts are not ordered — that is this file's long-standing stance — but the STORAGE is: an allow from a check that started before a refusal is not stored, and the row it would have replaced is dropped. Three things make that hold rather than usually hold: the refusal is remembered **synchronously**, before its DELETE is even queued; every mutation of one contact's row runs alone, in a queue keyed by that contact; and what is kept is the **latest** refusal known, not the last one recorded.

The asymmetry is deliberate: an allow never survives a refusal, and a refusal may cost a newer allow its row. When two overlapping checks cannot be ordered from what is on disk, the side taken is the one that asks the endpoint again — one extra call, against a contact served after a refusal for the whole TTL.

The guarantee is **per process**, like the `ingest:<threadId>` section the whole runtime already depends on (`docs/deploy.md`, "Single replica (or one leader)").

## What the row holds

Two SHA-256 fingerprints, an expiry, and the endpoint's own `context` bag. The phone, the email and the `identifier` are **not** in the row. The bag is stored so the reused turns carry the same prompt block the first turn did, re-read through the same bounded reader on the way out. The table is bounded by the contacts table itself, cascades away with the contact or the agent, and both cascading foreign keys are indexed.

## Validation scope

**Proved by test** — `tests/modules/contact-auth-grant.test.ts` (41 cases, DB-backed), `contact-auth-gate-e2e.test.ts` (two cases through `processChatwootDelivery`), `contact-auth-settings.test.ts`, `tests/client/contact-auth-ttl-zero.test.ts`:

- the premise, the reuse, the context surviving it, and the row holding no identity in the clear (asserted against the whole row rendered to text by Postgres);
- every way out: TTL, identity, policy, credential rotation and deletion, a refusal under either mode, and the mode switch;
- the races, each with a deterministic seam rather than a sleep: an allow in flight against a refusal, an allow that already passed its check, a refusal queued behind another mutation, a reader overlapping a delete, an older refusal landing late, a delete abandoned on the deadline;
- the failure paths: a grant read that throws, a delete that throws, an upsert that throws, a vault read that throws, a read that hangs past the budget;
- the bookkeeping: eviction sparing debts and recent refusals, and the overflow draining on its own sweep.

**Mutation battery** — every fix in this PR was mutated and each mutation fails at least one test. Four mutations survived at first and each pointed at something real: two mechanisms with no coverage (the queue, and the read being inside it — one gained a test, the other was removed), and two tests that measured the opposite of their name.

**Proved live** — a real `Bun.serve` endpoint on `127.0.0.1`, the real `fetch`, the real SSRF guard, the real vault credential injection and the real Postgres, with nothing injected but the two Prisma clients:

```
1) three messages in a row       verdicts: allowed | allowed (reused) | allowed (reused)
                                 requests that REACHED the endpoint: 1
                                 header on it: Bearer LIVE-SECRET
                                 stored row: (…,c4e5d0e7…,6e567d66…,"{""plan"": ""premium""}",…)
2) phone changed, endpoint now refuses    denied | requests: 2 | rows after: 0
3) grant forced past its expiry           allowed | requests: 4
4) credential rotated between messages    allowed (reused) | allowed
                                          requests: 5 | header: Bearer ROTATED-SECRET
5) credential deleted                     error credential_unavailable | requests: 5
```

**Not validated**

- No live run against a rate-limited or slow third-party endpoint: the shape above is a local HTTP server, so it proves the ordering and the call count, not what an operator's production endpoint does under a burst.
- The console controls were exercised by the existing editor tests, a source fence and type-checking, not by a browser session.
- Multi-replica behaviour is not tested and is not claimed: the ordering is process-local, and the consequence is written into `docs/contact-auth.md`.
- The two MCP ceilings were re-measured on the FINAL tree (12,942 and 42,497) rather than on the one this branch opened with, since both `.describe()` strings were rewritten during the review loop; raised to 13,050 and 42,600. The description ceiling does not move — this adds fields to a block, not a tool.

Fixes #189


gabrieljablonski:
Review round 1 (local codex, `gpt-5.6-sol`, effort high): **1 finding, P1.** Recorded here as a plain comment rather than as a review, because the reviewer bot's token was unavailable at publish time; nothing about the finding changes.

Verbatim:

> **[P1] Drop grants after denials in per-message mode**
>
> When an agent previously used `once`, switching to `perMessage` leaves its grant row intact, and this condition prevents a fresh per-message denial from deleting it. If the operator later switches back to `once` before the TTL expires, the old allow is reused and the contact is served despite the endpoint's newer denial. Drop grants on denial regardless of the current mode, consistent with CLAUDE.md:39.

Verdict: correct, and it lands on a claim this PR makes out loud. The invalidation table says "a fresh check refuses" drops the grant; the code only honoured that under `once`, so the rule held for the mode that reads grants and not for the one that cannot. The mode is not part of the policy fingerprint (deliberately — it decides who READS a grant, not who answered it), so a round trip `once → perMessage → once` inside the TTL resurrects an allow older than the refusal.

It also caught a comment of mine that asserted the opposite: `contact-auth-grant.test.ts` said switching back "starts from a fresh question", which is exactly what was not true.

Fix in the next commit: the drop moves out of the mode branch. The asymmetry is the rule — **only `once` grants, every mode un-grants** — and it is what the round-trip case needs.

gabrieljablonski:
Review round 2 (local codex, `gpt-5.6-sol`, effort high): **1 finding, P1.** Recorded as a comment again — the reviewer bot's token is still unavailable.

Verbatim:

> **[P1] Purge mismatched grants before returning a cache miss**
>
> When an operator changes a fingerprinted policy field to revoke grants, this branch only ignores the existing row. If the field later returns to its previous value before the original expiry—for example, lowering the TTL to clear grants and then restoring the desired TTL—the old hash matches again and the stale authorization is reused without calling the endpoint. This violates the invariant that an identity or policy change drops the grant (CLAUDE.md:39); delete or tombstone mismatched rows, or use a monotonic policy version so they cannot become valid again.

Verdict: the reasoning is right and the invariant it cites is mine, written in four places (the docs table, CLAUDE.md, the editor hint, the MCP `describe`). The fingerprint is a pure function of the policy, so it can only ever be a MATCH rule; I sold it as a revocation.

Measuring it made the shape sharper than the finding states, and the fix follows the measurement rather than the suggestion:

- with **no traffic** during the nudge, nothing reads a grant, so nothing happens to any of them and the restore makes every one valid again;
- with **traffic** during the nudge, the re-ask REPLACES the row with one written under the nudged policy, and the restore then invalidates that one instead.

Either way the operator never gets an empty table — they get a grant under whichever policy was in force when each contact last wrote. So "delete mismatched rows on read" does not close it either: in the first case there is no read to delete on, and in the second the row that survives is the fresh one.

What closes it is a monotonic epoch (the finding's other suggestion): a stored-state purge with a button, a settings field and a comparison at read time. That is a new operator surface, and #189 scoped the way out as the TTL plus invalidation on identity change. So this round **narrows the claim instead of growing the feature**, and says the gap out loud:

- the policy half is documented as a match rule, in all four places;
- both shapes above are pinned as tests, including `contactAuthPolicyHash(restored) === contactAuthPolicyHash(before)`, so nobody reads the fingerprint as a revocation again;
- `docs/contact-auth.md` gains "there is no clear-everything lever, deliberately", naming what does end reuse: the TTL, the identity, a refusal, and `mode: "perMessage"` — which is immediate and complete while it is off.

If a purge is wanted, it is its own issue and its own decision.

gabrieljablonski:
Review round 3 (local codex, `gpt-5.6-sol`, effort high): **2 findings, P1 + P2.** Comment again — the bot token is still down.

Verbatim:

> **[P1] Do not swallow failed grant revocations**
>
> When this deletion encounters a transient database error, it logs and returns success. With an existing `once` grant, a refusal under `perMessage` can therefore fail to remove the row; after the database recovers and the operator switches back to `once`, the old allow is reused despite the newer refusal. This violates the repository invariant that a fresh refusal under any mode ends the grant (CLAUDE.md:39); the revocation must fail closed or remain durably retryable.

> **[P2] Include grant I/O in the authorization deadline**
>
> In `once` mode this new database read runs before the timeout starts, while the new write/delete operations are awaited before returning after the endpoint responds. If the pool is saturated or a transaction is slow, a gate configured for one second can block the webhook for several additional seconds, contradicting the documented single-budget timeout contract (CLAUDE.md:39, docs/contact-auth.md:77).

Both correct, and both land on a promise that was already written down before this PR touched it.

**P1.** The drop is best-effort, which is right for a write nobody is waiting on and wrong for the one write that ENDS an authorization. Turning a failed delete into a failed verdict is not the fix — the verdict is `denied`, and a bookkeeping error must not cost the customer their deny message and the handoff. What it gets instead is a refusal the process refuses to forget: a failed drop is remembered per contact, `readContactAuthGrant` returns nothing while a refusal is unrecorded (so the endpoint is asked, which is the fail-closed answer), and the delete is retried on that next read. Restarting the process before the retry succeeds is the residual, and it is now stated in the module rather than implied.

**P2.** The deadline used to start after the grant read and end before the write. It now starts before the read and covers the write and the delete, so `timeoutMs` means what `docs/contact-auth.md` says it means: one budget over every step that waits. A stored verdict served from a healthy database costs one indexed read inside that budget; a saturated pool spends the budget instead of adding to it.

Both are pinned by tests that fail without the fix: a client whose grant DELETE throws, and a client whose grant READ sleeps past the budget.

gabrieljablonski:
Review round 4 (local codex, `gpt-5.6-sol`, effort high): **3 findings, P1 + P2 + P2.** Comment again — the bot token is still down. All three correct.

> **[P1] Serialize grant updates against concurrent refusals**
>
> When two messages for the same contact use different request keys, notably unlock messages with `includeMessageText`, their checks run concurrently. These independent delete/upsert branches are unordered, so a later message can be denied and delete the grant before an earlier slow allow upserts it again; `once` then serves a contact after a fresh refusal.

> **[P2] Bound failed-refusal state**
>
> When a grant DELETE rejects or merely exceeds the shared deadline, this adds a key with no expiry or size bound. In `perMessage` mode grant reads are skipped, so an allowed next request—or no subsequent denial—never retries or clears it; many distinct slow/refused contacts therefore grow this process-global set indefinitely, even when the timed-out DELETE later succeeds.

> **[P2] Preserve the minimum TTL when saving zero**
>
> When an imported/API configuration contains `grantTtlSeconds: 0`, or an operator enters zero, the runtime clamps it to 60 seconds as documented, but this falsy fallback serializes it as 86,400 seconds. Saving any Behavior edit can therefore silently extend authorization reuse from one minute to one day.

**P1.** `docs/contact-auth.md` already says verdicts can settle out of order and that ordering them would cost more than it buys — but that was written when a verdict lived for one message. A stored one lives for the TTL, so the same out-of-order answer is now a different size of mistake. Serializing the writes would not fix it either (the last writer still wins, and the last writer can be the older answer): what is stored is now ordered by WHEN THE CHECK STARTED. An allow from a check that began before a refusal landed is older than that refusal however late it arrives, so it is not stored and the row it would have replaced is dropped. The verdict itself is untouched — the message it answered still gets its turn.

**P2 (state).** Both halves are right, including the one that is easy to miss: a delete that merely timed out may well have succeeded, and the entry would then suppress reuse for that contact for the life of the process. The retry moved out of the read path (which `perMessage` never takes) into the check itself, so any later message from that contact clears it; and the store is bounded by entry count, insertion-ordered, exactly like the notice cooldown next to it.

**P2 (zero).** Fixed for `grantTtlSeconds` alone, with the reason written down: the twelve other `Number(x) || <default>` fields in that save share the spelling and not the risk, and sweeping them would be changing shipped behavior on fields nobody reported. The nearest neighbour, `contactAuth.timeoutMs`, is named in the test with its measurement — floor 1000, fallback 5000, same direction, bounded by ten seconds instead of by a day.

Three tests, each proved red against the restored defect. The race one is ordered by a promise resolved from inside the slow request rather than by timing, after the timing version passed under `-t` and flipped order under the full file. The zero one is a source fence, and its first version **passed with the defect in place** — it was reading the form-state block instead of the save block, since the file holds two `contactAuth: {`. That is why a fence gets run against the defect before it is trusted.

gabrieljablonski:
Review round 5 (local codex, `gpt-5.6-sol`, effort high): **3 findings, P1 + P1 + P2.** Comment again — the bot token is still down.

> **[P1] Revoke grants permanently when identity changes**
>
> When an identity changes and later returns to its previous values before a verdict replaces the row—for example, the intervening check errors—this mismatch only returns `null` and leaves the old grant intact. Once the old fingerprint matches again, the grant is reused for its remaining TTL even though an identity change is specified to end it.

> **[P1] Prevent timed-out writes from reviving a refused grant**
>
> When the authorization budget expires while this upsert is queued or running, `underSignal` rejects only the wrapper; the Prisma transaction continues. A later denial can complete its delete first, after which this detached upsert may commit and recreate the grant, causing a future `once` request to serve an authorization older than the fresh refusal.

> **[P2] Preserve the original refusal time during retries**
>
> After a denial's delete fails, the next `once` check retries it here, but `dropContactAuthGrant` records the successful retry at the current time—after this check's `askedAt`. If the endpoint now allows the contact, `refusedSince` treats that old denial as newer and discards the grant, so the promised first positive verdict is not stored and the endpoint is called again on the following message.

Three rounds in a row have now found something in the same object — the in-process bookkeeping — and each round's fix produced the next round's finding. That is a model problem, not three bugs, so this round **restates the model instead of patching it again**. It is now written once, at the top of `grants.ts` and in the docs:

> A stored grant is served if and only if (1) it has not expired, (2) it matches the identity in force, (3) it matches the policy in force, and (4) this process holds no unconfirmed write about that contact. It is REMOVED only by a refusal, or by the verdict that replaces it.

Against that model, the three findings resolve as one design decision each:

**Identity (P1).** Rules 2 and 3 are the same KIND of rule, and neither is a revocation: a fingerprint that stops matching stops the grant being served, and a value put back matches again. Round 2 settled that for the policy half; the identity half is settled the same way here, in the same table, rather than being made an exception. What ends a grant is rule 4 and a refusal.

**Timed-out write (P1).** Correct and, more usefully, it is a defect that my own round-3 fix introduced: binding the post-verdict writes to the deadline is what created a detached statement able to commit after a later refusal. They are no longer bound by it. The budget covers everything that decides the ANSWER, the stored-verdict read included; the bookkeeping after the answer is awaited, because walking away from a Prisma statement does not stop it, and a write nobody waits for is a write nobody can order. A failed or timed-out write now marks the contact unconfirmed, which rule 4 turns into "serve nothing, delete first".

**Retry timestamp (P2).** Plain bug, from the same over-patching: the retry stamped the refusal with its own clock, so a fresh yes read as older than a refusal from before it and was thrown away. The instant a refusal carries is now the instant its CHECK started, and a retry that lands later keeps it.

The residual is stated rather than papered over: two checks that overlap in time can still interleave their writes, and the bound on that is the TTL the operator chose — the same bound the mode already carries for a revocation, and the same stance this file already takes for verdicts themselves ("the runtime does not serialise"). Making it strict needs ordering in the database (a tombstone row and a conditional write), which is a protocol, not a patch, and it buys a guarantee the mode does not otherwise offer.

Both new behaviours are pinned by tests proved red against the restored defect.

gabrieljablonski:
Review round 6 (local codex, `gpt-5.6-sol`, effort high): **4 findings, two P1 and two P2.** Comment again — the bot token is still down. All four correct, and all four in the ordering machinery.

> **[P1] Mark refusals before awaiting the delete** — an older allow can check `refusedSince` before the refusal's `remember` executes; if the delete then commits before the allow's upsert, the upsert recreates the stale grant.

> **[P1] Preserve the newest refusal timestamp** — with three overlapping checks, an older refusal finishing late overwrites a newer one's timestamp, and an allow asked between them then passes.

> **[P2] Check expiry after the database read** — a row fetched just before `expiresAt` and returned after it still passes, authorizing one more turn.

> **[P2] Bound retry cleanup by the gate deadline** — the retry of an unconfirmed write awaits an un-signaled `runScopedOn` before the answer, so a pool in trouble adds the scoped transaction's own maxWait and timeout on top of `timeoutMs`.

The first two are the same defect seen from two sides, and it is not the rule, it is that reading the rule and acting on it were two steps with an `await` between them. Marking before the delete closes one interleaving; it does not close the other, where the allow passed its check a moment BEFORE the refusal arrived and is merely slow to write. So the fix is not a third guard on the rule: **every mutation of one contact's row now runs alone**, in a queue keyed by that contact (`withKeyedQueue`, the primitive the ingest path already uses), which makes check-and-write one step. The refusal is also remembered before its delete, and what is kept is the LATEST refusal known rather than the last one recorded.

The two P2s are one line each: the clock that decides expiry is read after the awaited query, and the retry — which runs BEFORE the answer, unlike the bookkeeping that follows a verdict — is under the gate's deadline.

Four mutations, one per fix, each run alone: **three failed a test and the fourth did not.** Removing the queue broke nothing, because the test I had written for it was hitting the interleaving that mark-before-delete already covers. The queue was a mechanism with no coverage, which is a mechanism nobody can remove later with confidence. It has a test now — an allow held inside a slow `upsert`, with the refusal running to completion around it — and that test fails without the queue.

While writing those, the test proxy itself was wrong in a way worth naming: the "slow delete" hook forwarded to the outer client instead of the transaction delegate, and under RLS an unscoped statement matches **zero rows**. It deleted nothing, and the read hook returned nothing, so both tests were green while measuring the opposite of their names.

gabrieljablonski:
Review round 7 (local codex, `gpt-5.6-sol`, effort high): **1 finding, P1.** Comment again — the bot token is still down.

> **[P1] Recheck revocation state after fetching the grant**
>
> When a `once` request passes this pre-query check, a concurrent `perMessage` request can receive a denial, mark/delete the grant, and finish before the first SELECT returns its earlier snapshot. Because the reader never checks `unconfirmed`/`refusedAt` again, it can still return that stale grant and allow a turn after the fresh refusal.

The description is accurate; the conclusion is where this round stops rather than patches. I took the suggestion first — put the read in the same per-contact queue as the mutations — and then measured it, which is what settled it:

- **The window the finding describes is already closed.** A refusal is remembered BEFORE its delete is attempted (round 6), so a read overlapping a refusal *in flight* sees that and serves nothing. The test for exactly that overlap passes with the read queued and passes with it un-queued.
- **What the queue would actually add** is ordering against a refusal that lands *after* the read started. Serializing does not settle that: with the read queued, the refusal simply waits, and the read — which legitimately came first — still serves its row. The boundary moves; the overlap does not go away.
- So the read went back out of the queue. A mechanism no test can tell apart from its absence is a mechanism nobody can remove later with confidence, and this one costs a promise hop on the hot path of every reused message.

What the module says instead, out loud, is the rule it has been making all along: **a check in flight is not invalidated by a verdict that lands during it.** `docs/contact-auth.md` already states that for verdicts themselves ("the runtime does not serialise"), and "the read came first" and "the refusal came first" are both true readings of one overlap. The bound is the TTL the operator chose, as everywhere else in this feature.

Round 6's fixes stay: the mark before the delete, the newest-refusal rule, the expiry clock read after the query, and the retry inside the budget — those close windows where the answer is not a matter of reading the overlap two ways, but simply wrong.

gabrieljablonski:
Review round 8 (local codex, `gpt-5.6-sol`, effort high): **1 finding, P2.** Comment again — the bot token is still down.

> **[P2] Keep timed-out delete retries in the mutation queue**
>
> When an unconfirmed DELETE exceeds `timeoutMs`, `underSignalMaybe` rejects while the underlying `runScopedOn(...deleteMany)` continues, so `dropContactAuthGrant` releases the keyed queue even though the mutation is still active. A following message can successfully retry, receive a fresh allow, and upsert its grant before the abandoned DELETE commits; that late DELETE then removes the new grant.

Correct, and it is the exact pair of what round 7 sent back: there the answer was that a check in flight is not invalidated by a later verdict; here the statement is still in flight and the QUEUE let go of it, which is a different thing and a real defect. Its direction is safe (a lost grant costs an endpoint call, never a stale authorization), which is why it is a P2 and not a P1 — but the fix is small and exact.

**The queue now holds the statement; the caller holds only its wait.** The deadline bounds what the webhook waits for, and nothing else: the slot stays taken until the delete settles, so the next message's allow is ordered after it rather than around it. The caller that walked away leaves the contact unconfirmed, which is the state the next check already knows how to handle.

Pinned by a test that runs the whole sequence — a failed delete, a retry that outlives a one-second budget, an allow stored right after it, and the straggler landing 1.2 seconds later — and fails when the delete is put back under the caller's signal.

gabrieljablonski:
Review round 9 (local codex, `gpt-5.6-sol`, effort high): **2 findings, P1 + P2.** Comment again — the bot token is still down.

> **[P1] Persist refusal ordering across web replicas** — the queue and the map are process-local, so with worker-disabled web replicas a refusal recorded on B does not order an allow in flight on A, and the stale allow can be stored afterwards.

> **[P2] Index the cascading grant foreign keys** — deleting an agent, or the contact cleanup, probes this table by `agent_id` / `contact_id`, and the only index leads with `tenant_id`.

**P2 is a plain miss and is fixed.** Postgres does not index a foreign key on its own, both are `ON DELETE CASCADE`, and the unique index cannot serve either probe. Two indexes added to the migration, which was edited in place and re-applied to both local databases (dropped the table, removed its `_prisma_migrations` row in dev and test, `migrate deploy`) rather than left as a second directory for a table this same PR introduces.

**P1 is accurate and is where this stops.** The scope it asks for is wider than the runtime around it provides: `docs/deploy.md` says "Run one replica", allows extra web replicas with the workers off, and states that the `ingest:<threadId>` critical section — the one every turn, every nudge and every compaction passes through — is serialized by exactly this primitive, `withKeyedQueue`, in-process. `docs/contact-auth.md` already declares the same invariant for this module's single-flight and notice cooldown. Making the grants the one thing in the file that coordinates across replicas would be inverting the guarantee of everything they sit on.

So it is written down instead, in the two places a reader would look: the ordering section says the guarantee is per process and what narrows to under extra replicas (an allow in flight elsewhere can be stored after a refusal, for the rest of the TTL), and the in-process state section now lists the third thing that lives there, with the same restart note the other two carry. Coordinating grant mutations in Postgres is named as the way to close it — the same post-MVP path `deploy.md` names for the rest of this class.

Nine rounds, and the shape of them is worth stating: every finding has been real, the code is better for all of them, and three of the last five were answered by narrowing a claim or removing a mechanism rather than adding one. The remaining residuals are all the same residual — two things overlapping in time, bounded by the TTL the operator chose — and they are in the docs rather than in the code's imagination.

gabrieljablonski:
Review round 10 (local codex, `gpt-5.6-sol`, effort high): **2 findings, both P1.** Comment again — the bot token is still down. Both correct, both fixed.

> **[P1] Revalidate the credential before reusing its grants** — `credentialRef` is a stable id, so editing or deleting the vault entry leaves the policy hash matching, and a contact authorized under an old key stays authorized for up to 30 days.

> **[P1] Keep unconfirmed grant writes out of the eviction path** — the cap evicts the oldest entry even when it is a delete the process still owes; the next check then skips the retry and may serve the stale positive row.

**The credential** is now the fifth rule a stored grant has to satisfy: the entry it was obtained with must still exist and must not have been updated since the grant was written. One metadata read (`vault_entries.updated_at`), never a resolve — deciding whether a stored verdict counts must not refresh a managed-OAuth token. The deletion case is the one that made this a P1 rather than a nicety: a fresh check fails closed on an unreadable credential, and a stored verdict skips that check, so an operator who disarms the gate by removing its key would have gone on serving contacts for the rest of the TTL. Pinned by a test that rotates the secret, then deletes the entry, and asserts the second one comes back `credential_unavailable`.

**The eviction** now walks past unconfirmed entries. The cap protects against a flood of ordinary ones, and a debt is not ordinary — dropping it silently drops the only thing standing between a refused contact and the row its refusal failed to remove. What is left unbounded is the debt itself, which is bounded by the database being down, and each contact's next check clears its own.

The eviction test needed a second attempt, and the reason is worth recording: the first version refused ONE contact and then let three others through with allows, which write nothing and remember nothing — the map never filled, nothing was ever evicted, and the test passed against the blind eviction it was written to catch. It refuses all four now, and fails without the fix.

gabrieljablonski:
Review round 11 (local codex, `gpt-5.6-sol`, effort high): **2 findings, both P1.** Comment again — the bot token is still down. Both correct, both fixed.

> **[P1] Record refusals before entering the queue** — when an older mutation holds the contact's queue, a new denial does not mark anything until its callback starts, so a concurrent check can serve the pre-denial row.

> **[P1] Version grants from the resolved credential** — if the credential rotates after it was resolved but before the grant is upserted, the grant's `updatedAt` is later than the rotation, and comparing instants reads the stale verdict as current.

**The mark now jumps the queue; the DELETE stays in it.** The mark is what an unqueued reader consults, and it is an assignment — there was never a reason for it to wait its turn. Queued, a refusal is invisible for exactly as long as the mutation in front of it takes, which is the window a reader needs to serve the row that refusal is about to remove.

**The credential went INTO the policy fingerprint** rather than being compared against the grant's write time, which is the better shape and not just the fix: a fingerprint built from the revision this check actually used simply stops matching the next check's, so the rotation-during-the-request race has nowhere to live. The revision is read once, at the start of the check, and the same value looks a grant up and stores one.

Both pinned, and the second test took three attempts, each failing for a different honest reason worth recording:

1. the writer meant to hold the queue **reused** the existing grant instead of writing, so nothing ever held anything and the test hung on its own gate;
2. holding the queue with a committed row does not work either, because the upsert is inside the transaction — nothing it writes is visible to the reader until the queue is released anyway;
3. what does work is a write under a DIFFERENT policy, whose upsert has not committed: the earlier row is still the visible state, and the queue is busy.

Only the third arrangement fails when the mark is put back inside the queue. The first two passed against the defect they were written to catch.

gabrieljablonski:
Review round 12 (local codex, `gpt-5.6-sol`, effort high): **1 finding, P2.** Comment again — the bot token is still down. Correct, and it is a defect I introduced one round ago.

> **[P2] Do not reuse a constant for unreadable credential stamps**
>
> If the vault metadata query fails when an allow is stored and fails again on a later request, both policy hashes use the constant `"unknown"`, so the grant matches and is reused even if the credential was rotated or deleted between requests.

The `"unknown"` sentinel was written to mean "this must not match", and a constant is precisely the thing that always matches itself. Two blips and the rotation between them disappear.

Unreadable is no longer a stamp at all: `readCredentialStamp` returns a result, and a check that could not take the revision uses **no** stored verdict and writes **none**. It costs an endpoint call, which is the direction this gate takes when it does not know something. `"missing"` stays a real stamp, deliberately — a deleted credential is a fact the read established, not a failure to establish one.

Pinned by a test that answers twice under a vault read that throws, and fails the moment the sentinel comes back.

gabrieljablonski:
Review round 13 (local codex, `gpt-5.6-sol`, effort high): **1 finding, P2.** Comment again — the bot token is still down.

> **[P2] Keep a newer grant when an older refusal finishes late**
>
> When two unlock checks overlap, an older invalid-code request can finish after a newer valid-code request has already stored its grant. This unconditional delete then removes the newer grant, so the successful unlock is not reused and the endpoint is queried again.

Accurate, and it is the mirror image of the rule that has been enforced since round 4 — with the costs on opposite sides. An allow that loses the ordering costs a contact served after a refusal, for the whole TTL. A refusal that "loses" it costs one endpoint call: the unlock that just succeeded is re-asked on the next message and answered yes again.

So the ordering is asymmetric, and this round says so rather than closing it. When two overlapping checks cannot be ordered from what is on disk, the side taken is the one that asks the endpoint again. Making it symmetric means storing the deciding check's instant on the row and making the delete conditional on it — the row's own `updatedAt` cannot stand in, because it records when the grant was WRITTEN, and the write of an allow that started before a refusal is exactly the case that must not survive. That is a protocol, for a symptom measured in one call.

Written into the module header and `docs/contact-auth.md`, and pinned by a test that performs the late refusal and asserts the grant goes — a fact, so that nobody later flips the asymmetry to the unsafe side without a test going red.

gabrieljablonski:
Review round 14 (local codex, `gpt-5.6-sol`, effort high): **1 finding, P2.** Comment again — the bot token is still down.

> **[P2] Preserve refusal markers until older checks settle**
>
> When a request for contact X is still in flight, a newer refusal completes, and more than 10,000 other contacts update this map before that request returns, eviction can remove X's `refusedAt` marker. The older allow then passes `refusedSince` and writes a grant for the full TTL.

Real, and it has a bound that makes the fix a rule rather than a mechanism: **a check cannot outlive its own budget**, and that budget is clamped at `CONTACT_AUTH_TIMEOUT_MAX_MS`. A refusal older than that window can no longer be the one an unfinished allow has to lose to; a refusal younger than it can. So eviction now walks past recent refusals as well as past debts.

What is left unbounded is a flood of ten thousand refusals inside ten seconds on one process, and it drains by itself as those markers age past the window — which is a better shape than tracking in-flight checks, since that would be a second registry to keep correct for a case the deadline already bounds.

One detail the fix needed and the finding could not know: the window is measured against the WALL CLOCK, never against the instant in the patch. A retry carries the ORIGINAL refusal instant (round 11), so measuring from it would age every other entry by however long the retry took.

Pinned by a test that records a refusal, churns other contacts past a lowered cap, and then lands an older allow — which finds the marker still there and writes nothing.

gabrieljablonski:
Review round 15 (local codex, `gpt-5.6-sol`, effort high): **1 finding, P2.** Comment again — the bot token is still down.

> **[P2] Re-run eviction after recent refusals age out**
>
> When more than 10,000 contacts are refused within the 10-second protection window, every entry is skipped and this returns without scheduling another eviction. If no later call to `remember()` occurs, the overflow remains allocated indefinitely instead of draining after the window, so a refusal spike can permanently exceed the intended memory bound.

Correct, and it catches a sentence I wrote last round rather than the code alone: "drains by itself as those markers age past the window" was not true of anything — eviction runs when a refusal arrives, and a spike that stops refusing is exactly the case where none does.

It drains on a scheduled sweep now: one unref'd timer armed for the earliest protected marker's release, which is the same idiom the notice cooldown next to it already uses and the reason that store could make the same claim honestly. The docs say sweep rather than "by itself".

Pinned by a test that fills the map with markers all young enough to be protected, waits, and finds it back at the cap without another refusal having arrived — which fails without the sweep. The window and the cap are both overridable in tests, so this costs 100 milliseconds rather than ten seconds and ten thousand contacts.

gabrieljablonski:
Review round 16 (local codex, `gpt-5.6-sol`, effort high): **0 findings, "patch is correct".** The loop is clean.

Sixteen rounds, twenty findings, and every one of them was real. Worth recording what the shape of them was, because it is not "the patch was bad twenty times":

- **Six were plain defects** — a refusal that could not reach a grant under the other mode, a swallowed DELETE, a deadline that did not cover what it promised, a typed zero becoming a day, unindexed cascading foreign keys, a retry stamping a refusal with its own clock.
- **Five were claims this PR made that the code did not keep.** Those were answered by narrowing the claim, not by growing the feature: the policy fingerprint is a match rule and not a clear, the identity fingerprint likewise, the ordering is per process, the ordering is asymmetric, and "drains by itself" needed something to do the draining.
- **Three were mechanisms of mine producing the next round's finding.** Round 3's deadline on the post-verdict writes created round 5's detached statement; round 11's sentinel created round 12's repeatable fingerprint. When it happened twice, round 5 stopped patching and wrote the model down once, and rounds 7 and 13 stopped again and named the boundary instead of moving it.
- **Two mechanisms were removed** for having no test that could tell them from their absence: the read inside the mutation queue, and an early version of the queue itself.

The mutation battery earned its keep four separate times, and not by killing bad code: it found tests that passed against the very defect they were written for — a fence reading the wrong `contactAuth` block, a proxy forwarding to an unscoped client (RLS matches zero rows, so a "slow delete" deleted nothing), a race ordered by timing that flipped under the full file, and an eviction test whose traffic wrote no entries at all.

The PR body has been rewritten against the tree the loop produced rather than the one it opened with, and the two MCP ceilings were re-measured for the same reason: both `.describe()` strings changed twice on the way here, and the numbers the first measurement justified had eight and three characters of headroom left, which is not headroom.

Issue #394, opened 2026-08-27
A tenant setting changed over REST is unattributable, so a consent switch would be a record nobody can trace


gabrieljablonski:
Sub-issue of #306. Depends on the substrate sub-issue.

Move the tenant / tenant-settings / branding trail into the services. **This is the one family with a
dated dependency:** #365 is deciding whether the dashboard may carry a per-assignee cut, behind an
explicit opt-in whose entire safety argument is that switching it on is a deliberate act by whoever
answers for the deployment. That only means something if the record later says who performed the act
and when. Today it would not, because the switch would be a tenant setting written over REST.

**Audited today (MCP only, 6 actions):** `tenant_create`, `tenant_update`,
`tenant_settings_update`, `langfuse_connect`, `branding_set`, `branding_asset_set`.

**The REST twins that write nothing:**

- `src/api/v1/v1.controller.ts` — `POST /v1/tenants`, `PATCH /v1/tenants/:id`, `DELETE /v1/tenants/:id`
- `src/api/v1/tenant-settings.controller.ts` — `PUT /embedding`, `PUT /langfuse`, `PUT /company`,
  `POST /company/logo`, `DELETE /company/logo`
- `src/api/features/branding/branding.controller.ts` — `PATCH /`, `PUT /asset/:kind/:variant`,
  `DELETE /asset/:kind/:variant`

`updateCompanySettings` and `setCompanyLogo` have no MCP twin at all, so they need an action name
this issue invents rather than one it moves.

Branding is fleet-level: `branding_set` audits with `tenant_id` NULL under `asSuperAdmin`, and the
REST path has to land in the same place, not in the caller's tenant.

Issue #392, opened 2026-08-27
The audit row is written by the MCP transport, so it is not atomic with the mutation and cannot cover the other two doors


gabrieljablonski:
Sub-issue of #306.

The audit trail is written by the MCP transport, not by the core the transports share. Two
consequences, and this issue is the seam that removes both.

## The row is not atomic with the mutation it records

Every MCP write tool applies through the service and *then* opens a second transaction:

```ts
const updated = await updateAgent(ctx, id, patch, base);   // commits
await recordMcpAudit(ctx, base, { action: "mcp.agent_update", ... });  // separate tx
```

`recordMcpAudit` (`src/modules/mcp/write.ts`) resolves its own `asSuperAdminOn`/`runScopedOn`. A
failure between the two lands the mutation with no row, and nothing reports it. The one service that
already records — `src/modules/api-keys/service.ts` — has no such window: it writes inside the same
`runScopedOn` transaction as the `create`/`updateMany`.

## The row cannot cover a transport it does not live in

The MCP write tools and the REST controllers share **91 service symbols**; every audited action goes
through one of them. Written one layer down, in the service, the trail covers all three transports by
construction instead of once per transport.

The attribution substrate is already there: `TenantContext` carries `userId`, `role` and an
`actorType` whose own comment reads "Used only for audit attribution"
(`src/lib/tenancy/context.ts`). Only api-keys uses it.

## Scope

- A service-side seam in `src/modules/audit/` that records inside the caller's existing transaction,
  carrying the before/after projection and the string bounding that today live as `AUDIT_STR_MAX` +
  `clipText`/`makeStorable` in `src/modules/mcp/write.ts`.
- Widen `TenantContext.actorType` to `"user" | "api_key" | "mcp" | "system"`; `ctxOf` in
  `src/modules/mcp/write.ts` sets `"mcp"` (today it sets nothing, so a service-layer row from an MCP
  call would be attributed as a browser user).
- **Rename all 65 `mcp.<tool>` actions to `<entity>.<verb>`.** The action names what changed; the
  `actorType` names the transport. Same change through MCP and through the console becomes the same
  action, which is the whole point of the move. This is a **breaking change** for the `?action=`
  filter on `GET /v1/audit` and for `audit_list`: rows already written keep their old names, since
  history is not rewritten. `docs/mcp.md` enumerates the names and changes with them.
- Prove the shape end to end on **business hours** (3 actions, 3 REST routes, no external effect and
[...]

Issue #306, opened 2026-08-25
Audit trail only records MCP actions — equivalent REST mutations leave no trace


edsontaramelli:
Hi! We noticed the agents audit trail records actions performed via MCP, but equivalent mutations done via REST (settings, tools, vault, deployment, etc.) don't show up in the audit.

1. Is this the expected behavior today?
2. Is covering REST (and the console) in the same trail on the roadmap? For anyone operating multi-tenant with automation, a complete audit of mutations is a compliance item — today a change made via REST leaves no trace.

Observed on agents v1.4.x–v1.6.0 (self-hosted Pro). Thanks!

JotaSXBR:
This is still true on `main@432966d`, and there is now a concrete reason to decide it rather than leave it as a completeness question.

The call sites for `recordAudit` are the MCP write path (`src/modules/mcp/write.ts`), the API-key service and the MCP OAuth controller. `src/api/v1/tenant-settings.controller.ts` and `src/api/v1/agents.controller.ts` do not reference audit at all — so the console, which goes through REST, writes no entry for a settings change. That matches what you describe from the operator side. (A reading of the call sites, not a test.)

The reason to raise it now is #365. That issue is deciding whether the dashboard may carry a per-assignee cut, and the shape being discussed is aggregate by default with individual measurement behind an explicit opt-in. The entire safety argument for that opt-in is that turning it on is *a deliberate act by whoever runs the deployment and answers for it* — which only means something if, later, the record says who performed the act and when.

Today it would not. And a consent switch whose activation nobody can attribute is worse than no switch, because the deployment then believes it holds a consent record it does not hold.

So for most settings this reads as completeness, and I think that is fair. For that one it is a precondition. Whether it is solved here as a general REST trail or there as a single entry written by hand, it should not ship without one of the two.

gabrieljablonski:
Confirming both, with the measurement behind it.

**Yes, that is the behaviour today, and it is not by design.** The trail is written by the MCP
transport rather than by the services the transports share: all 65 audited actions live in
`src/modules/mcp/write*.ts`. Outside them there are two lines in the API-key service and one in the
MCP OAuth controller. The 124 mutating REST routes write nothing.

The fix is a move rather than an addition. The MCP write tools and the REST controllers share 91
service symbols, and every audited action goes through one of them, so written one layer down the
trail covers all three transports by construction. The attribution substrate is already in place:
`TenantContext` carries the actor and how they authenticated.

[...]

Issue #401, opened 2026-08-27
The audit trail has no page in the console, so the trail an operator is told to keep cannot be read by one


gabrieljablonski:
The audit trail has no page in the console. `GET /v1/audit` exists (TENANT_ADMIN, keyset by id,
filterable by action) and the MCP `audit_list` tool exists, but a grep for `audit` across
`src/client/` returns nothing: an operator who does not speak REST or MCP cannot read the trail at
all.

This is the read half of #306, filed separately because it does not depend on it. The endpoint
already has rows to show — every MCP write since the tool surface shipped — and holding the page
until the write side is complete would keep an existing trail unreadable for the whole of that work.

What the page needs, from the shape the endpoint already returns: the action, the actor and how they
authenticated (`actorType`), the target, the timestamp, and the `before`/`after` projection as a
field-level diff rather than two JSON blobs. Filtering by action is already in the endpoint;
filtering by actor and by date range is not, and whether to add it is part of this issue.

Two things the page must get right, both already true of the data:

- `before`/`after` are allowlist-sanitized at **write** time, so the page renders what it is given
  and never re-fetches the live record to fill a gap.
- Fleet-level rows carry `tenant_id` NULL and are visible only to a SUPER_ADMIN. The page is scoped
  by RLS like every other read; it must not present the absence of those rows as an empty trail.

Related: #306 (the write side, an epic).

JotaSXBR:
The shape of a real trail, since this page will be designed against one. Read-only, from a self-hosted production tenant, and it bears on three of the open calls in the body. Numbers also posted on #393, where they argue something different.

**40 rows over 16 days** (2026-08-08 → 2026-08-24), 13 distinct actions, 20 of them `mcp.prompt_set`.

## This is a legibility problem, not a scale problem

At that volume the keyset paging is never under pressure and the list is not the feature. The `before`/`after` pair is: `prompt_set` takes the whole prompt rather than a patch, so half the rows on this tenant carry two full system prompts of roughly 11k characters each. Two JSON blobs there are not merely inconvenient, they are unreadable — the field-level diff you already name is the whole page, and it should be designed against the largest pair rather than the smallest.

## The single most useful element is not in the list

The timestamp of the **newest row**, shown on its own.

[...]

PR #416, opened 2026-08-27
An agent edited in the console records its own change, not the MCP tool that used to

### LINKED ISSUE #393: An agent edited in the console leaves no audit row, because the trail lives in the MCP tool and not in the agent service
Sub-issue of #306. Depends on the substrate sub-issue (the service-side seam + the action rename).

Move the agent-configuration trail from the MCP transport into `src/modules/agents/`, so a change
made in the console leaves the same row a change made through MCP leaves.

**Audited today (MCP only, 8 actions):** `agent_create`, `agent_update`, `agent_delete`,
`agent_clone`, `agent_import`, `agent_tools_set`, `agent_settings_set`, `prompt_set`.

**The REST twins that write nothing** (`src/api/v1/agents.controller.ts`): `POST /v1/agents`,
`PATCH /v1/agents/:id`, `DELETE /v1/agents/:id`, `POST /v1/agents/:id/clone`,
`POST /v1/agents/import`, `PUT /v1/agents/:id/tool-selections`. Six routes, all of them the console's
Agent editor, all through `createAgent`/`updateAgent`/`deleteAgent`/`cloneAgent`/`importAgent`/
`replaceAgentToolSelections` — the same functions the MCP tools call.

The playground routes and `models/list`/`tts/list` are not configuration mutations and stay out.

Two details this family carries that the others do not:

- The prompt is the field most worth auditing and the one that most needs bounding: the projection
  must keep the audited-prompt discipline (`src/graph/prompt-audit.ts` is the sibling rule for the
  flow log), never the resolved prompt.
- `agent_settings_set` projects only the touched behavior blocks. The REST `PATCH` merges the same
  way, so the projection has to be built from the patch, not from the whole settings bag.

gabrieljablonski:
The agent-configuration trail now comes from `src/modules/agents/`, so a change made in the console
leaves the same row a change made through MCP leaves. Third sub-issue of #306, after #392 built the
seam and #394 moved the tenant families.

Eight actions were audited and all eight were written by the MCP transport after the service it
called had already committed. The six mutating REST routes of `agents.controller.ts` reach the same
service functions and the file never mentioned `audit`, so an agent edited in the console left no
record at all. Measured on the base, against a real Postgres:

| through the service (what the console reaches) | base | after |
| --- | --- | --- |
| `createAgent` / `cloneAgent` / `deleteAgent` | 0 rows | 3 rows |
| `importAgent` | 0 rows | 1 row |
| `replaceAgentToolSelections` | 0 rows | 1 row, grants before and after |
| `updateAgent` (prompt, settings, or anything else) | 0 rows | 1 row, named by what changed |
| `PATCH /v1/agents/:id` from a browser session | 0 rows | 1 `agent.prompt_set`, `actorType: user` |

## Three tools, one function, and one question underneath

`agent_update`, `prompt_set` and `agent_settings_set` are three MCP tools over **one** service
function, `updateAgent`, and the REST route reaches it with any mix of their fields. The service has
to decide which of the three a call is, and it cannot ask the caller: the editor's General tab PATCHes
`name`, `systemPrompt`, `enabled`, `mode` and `modelConfig` together on every save whether or not
they changed, and its Behavior tab sends the settings bag whole. So the action is read off the
**diff** — which is what `docs/mcp.md` already promises ("the same change leaves the same row
whichever of the three transports made it").

That put one question under everything: **what counts as the same configuration?** It is asked once,
in a table of canonical forms, and both the comparison and the projection read the answer. Each entry
is justified by what the RUNTIME reads, never by taste, and carries its measurement:

- **`settings`** is the bag as `readBehaviorSettings` resolves it, since that view is what every
  consumer takes. Two behaviours fall out rather than being handled: a value the readers CLAMP to the
  same result is the same configuration (`windowSeconds` of 1 and of 2 both read as 3), and a `Date`
  they produce is compared and stored as its ISO string, which is what the column can hold at all.
- **`modelConfig`** is the keys `modelConfigSchema` names, derived from the schema rather than typed
  out. `validateModelConfigForWrite` asks whether the value is valid and throws away the STRIPPED
  result, so a config valid apart from a stray key is stored with it — measured: a `PATCH` carrying
  `apiKey: "sk-…"` reaches the column. `exportAgent` already scans for that shape and refuses to
  emit; the row had no such guard.
- **grants** are sets, twice over: the list order is reassigned by the `deleteMany` + `createMany`
  this path is, and `enabledTools` / `knowledgeBaseIds` are allowlists the runtime reads by
  membership (`filterAllowed` builds a `Set`, `prepare.ts` asks `.includes`).
- **an `http(s)` URL carrying its own credential** — in userinfo, query or fragment — is left OUT of
  the canonical form wherever it appears, at any depth, in any of the nine URL-shaped fields an
  audited projection can hold. Bounded to `http(s)` deliberately: `"Pergunta: você quer?"` parses as
  a URL with protocol `pergunta:`, so a rule keyed on parseability would eat operator prose.

**An apply that changes nothing writes no row.** The console PATCHes a whole editor tab per save.
This is a behaviour change for the three MCP tools, documented in `docs/mcp.md` item 4.

**And canonicalizing must not hide a write.** Anything the readers do not see is absent from the
canonical form, so each field is compared TWICE — canonically, then on the stored value — and a
raw-only difference records `unreadConfigChanged: true`. A boolean, never the content and never the
key: an unknown key is caller-controlled and can itself be secret material, which is why
`assertNoSecrets` scans keys.

## Three things the issue asserts that do not hold

Re-measured rather than honored, since the issue is one of ours and predates #392.

- **The action names.** The issue lists `agent_create`, `prompt_set`, … ; #392 renamed them to
  `agent.create`, `agent.prompt_set`.
- **"The REST `PATCH` merges the same way."** It does not — `updateAgent` **replaces** the bag. The
  merge is the MCP tool's, and the console reaches the same outcome by spreading client-side.
  `docs/mcp.md` said "exactly like the REST `PATCH`"; corrected here.
- **"The projection has to be built from the patch."** Unimplementable at the service: through the
  REST door the patch **is** the whole bag. The intent is met by comparing instead.

The issue's remaining detail — keep the audited-prompt discipline of `src/graph/prompt-audit.ts` —
needed no code. That module masks the prompt AFTER interpolation, where the customer's name and
number have entered it; `updateAgent` only ever sees the operator's own template. The bounding half
is real and already belongs to the seam.

## Validation scope

**Proven by test** (`tests/modules/audit-agent-family.test.ts` + `tests/api/v1/agents-audit-actor.test.ts`,
49 tests, DB-backed): every one of the six service functions writes its row; the action is picked by
the diff, including the two shapes the console actually sends; a no-op apply writes nothing; the same
prompt rewrite through both doors leaves rows differing **only** in `actorType`; the MCP tools still
write one row each; a refused update leaves neither the change nor a row.

**Proven through the real door**: a `PATCH /api/v1/agents/:id` driven through `app.handle` with a
signed session cookie returns 200 and leaves `agent.prompt_set`, `actorType: "user"`, naming the
operator; the same request without a session is refused 401 before reaching the service. On the base
the authenticated half returns `Expected: 1, Received: 0` — the change lands and the trail stays
empty, which is the issue.

**Proven structurally** where a behavioural test cannot be honest: that the audit snapshot is read
under the write's lock, and that the clock is read once for the whole comparison. Both are races
whose window is the gap between two synchronous calls, and both were measured to survive a mutation
battery before a fence was written for them.

**Mutation battery**: run on every round's fix, against a control proven to print a number first.
Survivors were diagnosed rather than explained away, and one of those diagnoses was itself wrong and
had to be corrected two rounds later: a guard removed as unreachable turned out to be reachable
through `readToolPreconditions`, whose map is `Object.create(null)` and is keyed by tool name. A
survivor is a question, and answering it with an argument instead of a probe is how that one got
through.

**Measured cost**, since the row is written inside the mutation's transaction and the body would
otherwise be claiming that is free. A config save on a full settings bag (twenty blocks, five
follow-up steps, a 4000-char prompt), 20 iterations after 3 warmups, against a real Postgres:
**6.7 ms/save on the base, 13.6 ms/save here** — it roughly doubles, and the +7 ms buys two extra
statements in the transaction (the locked snapshot and the insert) plus canonicalizing both sides,
which materializes the reader's whole view twice. The path is operator-initiated and not per-turn:
the only callers of `updateAgent` are the REST route and the three MCP tools. A save that changes
nothing costs the same, because paying for both canonical forms is how it knows.

**Not validated**: no live exercise against Chatwoot or a model provider, because this change speaks
to neither. The console was exercised through its HTTP route, not through a browser.

**Waived, with the measurement**: round 13 asked for secret-value redaction on every projected string,
naming a credential in a URL PATH segment (`/bot/sk-…`) or an API key typed into a `name`. The
description is right and the remedy does not work: this repo's `redactSecretsDeep` redacts by KEY
NAME, not by value shape — measured, it leaves `sk-proj-AbCdEf0123456789`, a Telegram bot URL and an
API key stored under `name` all intact, and only reacts to a key literally called `apiKey`. Detecting
a secret by the shape of its value would be a new heuristic inside an audit trail whose most valuable
field is free operator prose. It is also not introduced here: the MCP projections already carried
`name` and `systemPrompt` verbatim, and this PR extends the same projection to the REST door. It
belongs to the epic (#306), across every family, not to this sub-issue.

Fixes #393


JotaSXBR:
Measured "an apply that changes nothing writes no row" against a real trail, since that behaviour change is the one thing here with no test that can tell you its size. It is bigger than it sounds, which supports the rule — and it costs one signal that is worth naming before it goes.

## How much of a real trail is no-op

A self-hosted production tenant, complete audit log since 2026-08-08 (40 rows, all `mcp.*`):

| | rows |
|---|---|
| `before` == `after` | **14 (35%)** |
| genuinely changed | 19 |
| no `before`/`after` at all (create / connect / import shapes) | 7 |

By action, the no-ops are almost entirely one:

| action | total | no-op |
|---|---|---|
| `mcp.prompt_set` | 20 | **13 (65%)** |
| `mcp.tenant_settings_update` | 1 | 1 |
| `mcp.agent_settings_set` | 9 | 0 |
| everything else (11 actions) | 10 | 0 |

So on this tenant the rule removes about a third of the trail, and two thirds of the prompt rows.

That reads as strong support. `prompt_set` takes the whole prompt rather than a patch, so re-sending an unchanged one is the ordinary accident rather than a rare one — and each of those thirteen rows carries a `before`/`after` pair of two ~11k-character prompts. The trail gets shorter and more honest at the same time, which is not a trade.

If anything 35% is conservative for what the console will produce: the PR's own reasoning is that a tab PATCHes every field it owns on each save, and these 40 rows come from an MCP client sending narrower payloads.

## The signal it costs, and why it is not obvious

`Agent.updatedAt` is `@updatedAt`, and in this diff `db.agent.update(...)` still runs unconditionally — only `auditMutation` is behind `if (audit)`. So a no-op apply bumps the timestamp and writes nothing.

Before this PR, that combination meant exactly one thing: **a write happened that the trail does not describe.** It is the only detector an operator has for the console gap #306 reports, and it is what I used to demonstrate #393 — `agent_get.updatedAt` at `2026-08-26T19:18:34Z` against a newest audit row of `2026-08-24T18:08:25Z`.

After this PR the same signature is also produced by a legitimate, deliberately unrecorded no-op. At 35% of applies that is not an edge case, and the operator cannot tell the two apart from outside.

Not an argument against the rule — the thirteen prompt rows are worth more gone than kept. But the fix is worth choosing rather than inheriting:

- **skip the write too** when the projection is null, so the timestamp and the trail agree by construction — except the transaction is doing other work (the follow-up fence's raw lock, the bot rename, the sweep arming), so this is a real question rather than a suggestion;
- or **leave it and write it down**, in `docs/mcp.md` beside item 4 and in #401, so the audit page never presents `updatedAt` drift as evidence of an unrecorded change. That is the cheap one, and once #393–#400 are all in, the detector matters less anyway.

Either way it should be a decision in this PR rather than something the next person rediscovers from a trail that looks wrong.

*(Read-only: `audit_list` and `agent_get` on a self-hosted tenant, complete result set. Diff read at PR head; `@updatedAt` confirmed on `model Agent` in `prisma/schema.prisma`.)*

Issue #187, opened 2026-08-21
The compacted memory of a handed-off attendance describes a conversation in which only the customer spoke


JotaSXBR:
## What happens today

Continuous ingestion folds into the contact's memory thread the messages no turn handled, so the agent has context when it resumes. It covers exactly one direction. `src/graph/ingest.ts:22`:

> Customer messages only. A human agent's own reply is NOT ingested: `rt` is resolved solely for a new INCOMING message (src/modules/chatwoot/webhook.ts), so an outgoing one never reaches this module at all. That predates memory compaction and is tracked on its own; the branch that used to handle it here was code no delivery could run.

The gate is `webhook.ts:529` (`incomingUnhandled` — new **incoming** messages only), under `webhook.ts:1779` (`enabled && mode === "production"`).

So on the most common shape of a real deployment — the agent qualifies a lead, hands off, a human closes the sale — the thread ends up holding the customer's half and nothing else. The agent's own replies are there (written by the turn); the human's are not.

## Why v1.9.0 makes this worse rather than the same

Before compaction, that thread was raw turns, and a reader could see it was one-sided. Now `summarizeAttendance` runs over the closed attendance and writes one paragraph of prose into `<atendimentos-anteriores>`, dated, presented to the model as *"resumos de atendimentos já encerrados com este mesmo contato"* — and it is derived from a transcript where the human's side is simply absent.

The result is not an incomplete memory. It is a **confident** memory of an attendance that did not happen: the customer's questions with no answers, the price they floated with no record of what was agreed, an objection with no record of how it was handled. That text then rides in every future turn as fact, for up to twenty attendances (`MEMORY_HEAD_MAX_ATTENDANCES`).

The gap is old. Compaction promoted it from visibly-partial history to authoritative-looking record, which is what makes it worth closing now.

## The seam was built for this and is currently dead

The normalized event already carries the author. `src/modules/chatwoot/types.ts:74-82`:

> The message author (message events only), from the payload `sender.webhook_data`. `type` is "user" (a HUMAN agent), "agent_bot" (a bot — ours or another), or null/absent (the customer, on incoming). **Drives continuous ingestion: a human agent's outgoing reply is folded into the agent's memory marked as such, our own bot's outgoing is skipped (already in the thread from the turn).**

`normalize.ts:86` populates it. Nothing reads it — that paragraph describes behavior the code no longer has.

Two more vestiges point the same way:

[...]

Issue #77, opened 2026-08-14
A state write from the console can be undone by a Chatwoot event that was already in flight


gabrieljablonski:
Found while reviewing #75, which fixed the sibling case (a message snapshot overwriting conversation state). This one predates that change and is orthogonal to it.

## What happens

`handoffConversation`, `unassignConversation` and `setConversationStatus` call Chatwoot over REST and then write the mirror directly (`updateMirror`, `src/modules/conversations/service.ts:577`), setting `status` / `assigneeId` / `assigneeType` and nothing else.

Chatwoot may have serialized a conversation event **before** that call and still be retrying its delivery (`AgentBots::WebhookJob`: `retry_on ... wait: 3.seconds, attempts: 3`). When that in-flight event lands after the local write it describes the conversation as it was before the operator acted, and the mirror applies it: the handoff or the resolve is undone until Chatwoot's own event for the operator's action arrives, or permanently if that delivery is lost.

While the row is wrong, the runtime's last-moment ownership recheck reads the mirror row (`src/graph/runtime.ts:439`, unlike the webhook path which prefers the payload), so the bot can answer on top of the human.

## Why the ordering added in #75 does not cover it

#75 orders conversation events by the conversation's own version (`conversation.updated_at`). A local write cannot claim one, because the two endpoints it calls do not return the version they produced:

- `POST .../conversations/:id/assignments` renders the **agent**, not the conversation (`assignments_controller.rb#render_agent`);
- `POST .../conversations/:id/toggle_status` renders `{success, conversation_id, current_status, snoozed_until}` (`toggle_status.json.jbuilder`).

So the stored watermark still describes the previous state, and an in-flight event carrying a greater version is accepted.

This is not a regression from #75: `updateMirror` never touched `lastEventAt` either, so the guard that preceded it (`existing.lastEventAt > newLastEventAt`) did not reject that event in this scenario either.

## The fix worth having

**Correction to an earlier version of this issue**, which said the REST never serializes a conversation's `updated_at`. It does: `GET .../conversations/:id` renders `json.updated_at conversation.updated_at.to_f` (`api/v1/conversations/partials/_conversation.json.jbuilder`), the same value and the same format the webhook carries. Only the two write endpoints above omit it.

So the fix needs no fork change: re-read the conversation after the write and let the local write claim the version it comes back with. That closes the window with the same key #75 already uses, with no clock of ours involved, at the cost of one extra REST call per console action.

[...]

PR #433, opened 2026-08-29
Stop answering over the colleague who just replied to the customer

### LINKED ISSUE #430: A reply sent from the paired WhatsApp device is invisible to the ownership gate, to memory and to the follow-up ladder, so the agent answers over a human already in the thread
Read on `origin/main@86a9dca` (pushed 2026-08-28 01:50Z) and on `fazer-ai/chatwoot@main`. Every claim below is a path and a line on those two trees, plus a measurement from one production deployment (WhatsApp/Baileys inbox, ~13 conversations sampled on 26–27 Aug 2026).

## The gap

An attendant who answers the customer **from the phone** — the device paired to the number the inbox is connected to, without opening the CRM — is invisible to this repository. The message is stored, it is delivered here, and then nothing reads it.

**It is stored sender-less.** In the fork, a message the session sent and WhatsApp echoed back is written with `sender: nil` and `message_type: :outgoing` (`app/services/whatsapp/baileys_handlers/concerns/message_creation_handler.rb:21-32`), and `content_attributes[:external_sender_name] = 'WhatsApp'` (`:164-168`). The comment at `:34-40` names this shape exactly: a row "rendered as if an agent had replied from the phone". Our own sends never take that branch — a Baileys send reserves its WhatsApp id first, so its echo is matched by `pending_source_id` and merged into the existing row (`:46-59`), which dispatches `message_updated`, not `message_created`.

**It is delivered to us.** `webhook_sendable?` is `incoming? || outgoing? || template?` (`app/models/concerns/message_filter_helpers.rb:8-10`), and `AgentBotListener#message_created` (`app/listeners/agent_bot_listener.rb:36-43`) fans it to the conversation's assigned bot and the inbox's. `Message#webhook_data` sends `sender: sender.try(:webhook_data)` (`app/models/message.rb:224`) — so the payload reaches us with `sender: null` and the full `content_attributes`.

**Then all three readers drop it:**

1. **Memory.** `isHumanAgentMessage` requires `e.message.sender?.type === "user"` (`src/modules/chatwoot/normalize.ts:466-473`). A sender-less outgoing message fails that clause, so `ingestUnhandledMessage` computes `role === null` and returns (`src/modules/chatwoot/webhook.ts:946-953`). This is the same defect #187 closed, still open on this path: the attendance the agent later summarises is one in which only the customer spoke.
2. **The ownership gate.** `shouldBotHandle` reads status and assignee only (`normalize.ts:422-432`, via `heldByAnotherParty` at `:362-373`); `act` is computed from those two (`webhook.ts:3204-3211`). A message changes neither, so the conversation stays `pending` and bot-owned, and the **next** customer message drives a full turn — with the agent speaking into a thread a human is already holding.
3. **The follow-up ladder.** `isFollowUpLive` gates on the same predicate (`src/modules/followups/eligibility.ts:62-70`), so the ladder keeps counting down over a conversation a human is running, and fires into it.

Chatwoot does not close this either: `mark_pending_conversation_as_open_for_human_response` is guarded by `captain_pending_conversation?`, which the fork hard-returns `false` (`app/models/message.rb:490-501`).

## Evidence (production)

Sampled 13 conversations from one deployment, 26–27 Aug 2026. Nine outgoing messages with `senderType: null` (device replies, distinct from the operator's composer messages in the same threads, which arrive as `senderType: "user"`) in five of them. Two of the five show the full failure:

**Conversation A** — a human opens the thread from the phone at 13:44:46 with three messages: a greeting, an introduction naming herself, and a voice note. The customer answers 2 minutes later, addressing her. The agent then runs a complete turn on that answer: private note, reopen, team assignment, and a public reply to the customer — 31 seconds after her voice note. Its private note reads "the amount is not in the available context": the three device messages, including the audio, were never folded into the thread, so the summary it hands the team is written as if she had never spoken.

**Conversation B** — the thread is *opened* from the phone (two device messages, a greeting and a question about a specific document). The customer replies to the human. The agent answers instead, asking the customer for the document number the human had just asked about. Thirty-three minutes later the follow-up ladder fires its re-engagement message into the same thread.

Neither conversation had a human assignee; both stayed `pending` throughout, which is exactly what the gate reads.

## What this issue delivers

One predicate — *an outgoing, non-private, non-reaction, non-imported message with no sender is an attendant reply sent from the device* — wired into the two consumers that already exist for the composer case:

- it ends the bot's attendance on that conversation, through the transition the platform already uses for "a human is on this" (status leaves `pending`), which silences the turn (`webhook.ts:3204`) and the ladder (`eligibility.ts:69`) with no new state;
- it is folded into the contact's memory thread as `human_agent`, so the attendance summary carries the business half of the dialogue (#187's rule, applied to the path it does not cover).

## Out of scope

- Any time-based or sliding silence window, and any per-contact silence configuration. This repository has no such primitive and does not need one here: "a human is on this" is already spelled as conversation state.
- Automatically returning the conversation to the agent. Return stays what it is today — `/reset`, the console, or REST.
- Eager STT/vision on outgoing messages (an attendant's device voice note stays untranscribed; `webhook.ts:954-958` documents why the media pass is inbound-only). The rendered line names the attachment, exactly as `renderAttendantMessage` does today.
- Per-attendant attribution of a device reply. It has no Chatwoot user, so it cannot be attributed to one — relevant to #365, not resolvable here.
- Group conversations and non-Baileys channels.

## Design

- **The predicate lives beside the existing one**, in `src/modules/chatwoot/normalize.ts`, next to `isHumanAgentMessage`, and is asked of `message_created` only — for the reason already written at `:475-477`: an attachment write-back re-dispatches `message_updated` for a message already handled.
- **Four exclusions, each with a live counter-example:** `private` (operator-to-team), `is_reaction` (a device 👍 is sender-less and would otherwise read as a reply — the same hazard `:461-465` already documents for the composer), `content_attributes.imported` (a first pairing runs a history import; `app/services/whatsapp/baileys/history_importer.rb:234` marks those rows and `Whatsapp::Session::SilentWrite` suppresses the rest, so the flag is the discriminator that keeps an import from silencing an entire backlog), and messages that are not `outgoing` (templates are `message_type: 3`, activities `2`).
- **The ownership write goes through the existing mirror/state-order path**, not a bare REST call, so the resulting Chatwoot event does not fight the mirror (`src/modules/chatwoot/state-order.ts`), and it is conditional on the conversation being `pending` and bot-owned at that moment — which makes a re-delivered webhook a no-op.
- **Memory ingestion reuses `armIngest` with `role: "human_agent"`** and `renderAttendantMessage` (`webhook.ts:959-996`). No new role, no new renderer.
- **Open decision for the maintainer:** whether this transition also records a handoff outcome (`taken_over`) for the KPI surface, or stays a pure state change. It has no `User` to credit, so recording it as a takeover would create a taken-over conversation with no attendant — which argues for the pure state change, but the call is the maintainer's.

## Expected tests

1. `normalize.ts` unit — the predicate is true for `message_created` + `outgoing` + `sender: null`; false for `sender.type === "agent_bot"` (our own reply), for `sender.type === "user"` (the composer path, still `isHumanAgentMessage`), for `private`, for `is_reaction`, for `imported`, and for `message_updated`.
2. Webhook integration, the regression that fails without the change — deliver a sender-less outgoing `message_created` on a `pending`, bot-owned conversation, then deliver an incoming customer `message_created`: today the second drives a turn; after the change no turn runs.
3. Memory — the same device message is folded into the contact thread as an attendant line, and its attachment is named rather than transcribed.
4. Follow-up — a scheduled follow-up for that conversation is not delivered after a device reply (`isFollowUpLive` false).
5. Idempotency — re-delivering the same message emits no second transition.
6. Import — a batch carrying `imported` leaves conversation status untouched.

## Acceptance criteria

- After a device reply on a `pending`, bot-owned conversation, the agent does not answer the next customer message, and the follow-up ladder does not fire.
- The device reply appears in the contact's memory thread as an attendant message, so the compacted attendance is not "a conversation in which only the customer spoke".
- Our own outgoing replies, composer replies, reactions, private notes, templates and imported history are unaffected — proven by test 1 and test 6.
- Re-delivery of the same webhook changes nothing.

gabrieljablonski:
## What was breaking

An attendant replying to the customer in a conversation the agent is handling does not take the conversation away from the agent, on **either** route a person can reply by. The agent then answers over them on the next customer message, and the follow-up ladder keeps chasing a thread a person is running.

Measured on a live fork (`chatwoot-pro`, Baileys inbox, agent bot bound):

| route | before | after the reply |
| --- | --- | --- |
| Chatwoot composer, `enable_auto_assignment=true` | `pending`, assignee `nil` | `pending`, assignee `nil` |
| the phone paired to the number the inbox is connected to | `pending`, assignee `nil` | `pending`, assignee `nil` |

Chatwoot's own mechanism for this is dead in the fork: `Message#mark_pending_conversation_as_open_for_human_response` returns early on `captain_pending_conversation?`, which the fork hard-returns `false`.

Both replies **are** delivered to the bot. Captured off the wire, the device reply arrives as:

```json
{"event":"message_created","message_type":"outgoing","private":false,"sender":null,
 "content_attributes":{"external_created_at":1788019218,"external_sender_name":"WhatsApp"},
 "conversation":{"status":"pending","meta":{"assignee":{"type":"agent_bot","id":20}}}}
```

`sender: null` is why `isHumanAgentMessage` (issue #187) never saw it: the fork stores an echo of a message the session sent with `sender: incoming? ? sender : nil`.

## What the report got wrong

The issue proposed the predicate *"an outgoing, non-private, non-reaction, non-imported message with no sender is an attendant reply sent from the device"*. That is not safe, and this is measured rather than argued. Three shapes Chatwoot itself produces are indistinguishable from it, and all three were captured reaching the bot as `message_created` on a `pending`, bot-owned conversation:

| origin | type | sender | private | content_type | content_attributes |
| --- | --- | --- | --- | --- | --- |
| an automation rule's `send_message` | outgoing | null | false | text | `{automation_rule_id: 7}` |
| a scheduled message whose author is not a `User` | outgoing | null | false | text | `{}` |
| a CSAT survey (`CsatSurveyService#build_csat_message`) | outgoing | null | false | input_csat | `{}` |
| **an attendant on the paired phone** | outgoing | null | false | text | `{external_created_at, external_sender_name: "WhatsApp"}` |

Under the proposed predicate, an operator's automation rule would read as a person taking the conversation over, and the agent would be silenced on it permanently. The discriminator is `content_attributes.external_sender_name`, which the fork writes on **every** WhatsApp session path (`baileys_handlers`, `zapi_handlers`, `Whatsapp::Session::Inbound::MessageWriter`, `ReactionStore`) and on no Chatwoot-originated message.

Two more claims in the issue were re-measured:

- **`imported` is not "the discriminator that keeps an import from silencing an entire backlog".** `Whatsapp::Session::SilentWrite` wraps the whole import run and its `SyncDispatcher` guard calls `ActionCableListener` and nothing else, so `AgentBotListener` never fires for a backfilled row at either level of the flag. The clause is kept as a cross-repo fence and documented as unreachable today, not as the protection.
- **The composer route has the same hole.** The issue scoped itself to the device because that is where it hurt in production. One predicate covers both.

## What this delivers

One predicate, two routes, wired into the consumers that already exist:

```
message_created && outgoing && !private && !reaction && !imported && (
  sender.type === "user"                                         // composer
  || (sender == null && ca.external_sender_name === "WhatsApp")  // the paired phone
)
```

- **Ownership.** `toggle_status → open` puts the conversation in the human queue, which is the transition the platform already spells "a human is on this" — so the webhook gate, the debounce flush and the follow-up ladder all go quiet with **no new state and no new predicate in any of them**. Return stays what it was: `/reset`, the console, or REST.
- **The mirror**, reconciled from a live read right after. The toggle endpoint renders no `updated_at`, so there is no version to write from the call itself; this is the same GET `mirrorConsoleWrite` pays for a console click (issue #77). It is what makes a re-delivery a no-op through the gate, and it closes the window in which the mirror still reads `pending` while the runtime's ownership recheck is reading that row.
- **Memory.** The reply is folded into the contact's thread as `human_agent` — issue #187's rule, applied to the route it did not cover.
- **The trail.** One `handoff` flow-log line at the moment of the takeover, with `detail.via` naming the route, so the operator sees the cause when it happens instead of an `ownership_lost` line on the next customer message.
- **The switch.** `agent.settings.takeover.onHumanReply`, **on by default**, reachable over REST, the console and MCP. A block of its own rather than a field on `handoff`: the console's Tools tab replaces that block wholesale on every save, so a field living there would be silently reset the next time anybody touched a tool.

## Validation scope

**Proved live**, against a real `chatwoot-pro` fork and a real Chatwoot database, and **re-run on the code as it stands after review** — the captured payload driven through `processChatwootDelivery`, four arms on the same conversation:

| arm | inbox provider | Chatwoot before | Chatwoot after | our mirror | memory |
| --- | --- | --- | --- | --- | --- |
| device reply, switch on (default) | baileys | `pending` | **`open`** | `open` | `human_agent` |
| device reply, switch off | baileys | `pending` | `pending` | `pending` | `human_agent` |
| the same payload with the marker removed (the automation shape) | baileys | `pending` | `pending` | `pending` | — |
| device reply, provider that does not reserve echo ids | zapi | `pending` | `pending` | `pending` | — |

The switch gates the takeover and not the memory, which the second row shows: the attendant's words are still folded into the contact's thread. The zapi row shows the opposite, and deliberately — where the message may be our own echo it is neither acted on nor filed as somebody else's.

The device row itself was produced by running the fork's own `Whatsapp::IncomingMessageBaileysService` on a `messages.upsert` with `key.fromMe = true` and an id Chatwoot never reserved, not by hand-writing a row.

**The `imported` clause, measured rather than read.** Three identical sender-less outgoing rows on one conversation, with a capture server on the agent bot's webhook:

| row | context | delivered to the bot |
| --- | --- | --- |
| 567 | no wrap (the control) | **yes** |
| 568 | `Whatsapp::Session::SilentWrite.wrap`, `imported: true` | no |
| 569 | `wrap(announce: true)`, the level the import's gap half runs at | no |

So the clause is unreachable through this event, as documented, and it stays as a cross-repo fence rather than as the protection.

**Proved by test** (`tests/modules/chatwoot-human-reply-takeover.test.ts`, 19 cases): the regression end to end (a device reply, then a customer message that drives **no** turn, with the first message's turn as the positive control); the composer route; idempotency under re-delivery; the three sender-less Chatwoot shapes; an imported batch; a reaction from the phone; our own bot's outgoing; a test-mode agent; an inbox whose agent has no bot; a delivery with no route bot; a conversation held by another persona's bot (taken over exactly once, on the route that holds it); a conversation claimed while the client is built; a live read with no version and one that fails outright; a state that moved while the toggle was in flight; the provider refusal; and the switch off. Plus the predicate's truth table (`tests/modules/chatwoot-webhook.test.ts`) and the reader's defaults (`tests/modules/handoff-targets.test.ts`).

**Review**: 5 rounds, 5 findings, all 5 real and applied — the fence and its identity, the moment the mirror is written and the condition on that write. No new piece entered after round 1; the same one kept getting correct.

**Mutation battery**: 31 mutations, 31 killed.

**Not validated**: the resolution funnel still classifies these conversations as `unresolved` rather than `handoff`, because `classifyOutcome` reads `assigneeType === "User"` and this transition assigns nobody — there is no Chatwoot user behind a reply typed on a phone. Attribution for the funnel needs state of its own and is deliberately out of scope here.

Fixes #430


gabrieljablonski:
## Review record, rounds 6 through 16

Round 5 came back clean, then a copy fix landed and invalidated it. Eleven rounds followed. Recording what they found and what was done, since the ones that were not fixed here are decisions rather than omissions.

**Fixed in this PR:**

| round | finding | what changed |
| --- | --- | --- |
| 6 | The mirrored row was written *after* the toggle, so a turn already running reached its ownership recheck inside the round trip, read `pending`, and answered over the person | The claim now happens in the same statement as the fence, before anything that waits on a network |
| 6 | The CAS was on `status: "pending"` alone, and a hand-back writes `pending` too, so it matched the state that had just replaced ours | Ordering by `chatwoot_status_at`, the rule `state-order.ts` already states |
| 7 | The claim was given back when the open failed — but a failed call is an *unknown* outcome, not a refusal, so rolling back put the agent into a conversation Chatwoot may have handed over | The claim stands; the next customer message settles it through the reopen exception |
| 7 | The CAS compared status and status version only, while assignee carries a separate ordering mark | The fence hands back the row it read and the swap compares all of it |
| 11 | An inbox whose agent has no bot row learned that from `toggleStatus` throwing, after the row already said `open` | Stood down before anything is claimed — a missing bot is *known*, unlike a failed call |
| 11 | The switch's help promised the paired phone unconditionally, while the device leg is refused on providers that do not reserve echo ids | Copy names the connections it holds for, in both locales |
| 12 | The claim reached no console: the event this delivery had already broadcast still said `pending` | The claim is announced when it is written, which matters most after a failed open, when nothing else ever corrects it |
| 13 | The takeover was gated on the agent being enabled, which reads the wrong subject | Production-only, switch-independent; test mode stays excluded |
| 15 | The fence read only the mirror, which can be behind Chatwoot, so an attendant who answered and then resolved had the conversation reopened under them | Asks Chatwoot first, as a gate — not a reconcile, which would stamp a version newer than the message that decided the takeover and refuse every takeover there is |

**Not fixed here, with issues:**

- **#436** — the claim carries no identity of its own until the reconcile stamps one, and deliveries for a conversation are never serialized. Four ways in are recorded there (a customer message serialized before the toggle committed; a delayed `conversation_*` event; any event on a deployment that sends no `updated_at`; and the live gate being unable to tell the original `pending` from a newer hand-back). Each was tried against the alternative and the alternative is worse: refusing `open → pending` on a message payload is the same write the failed-open recovery relies on, and comparing the live version would refuse the takeover whenever the customer is still typing.
- **#439** — a crash in the detached window strands the delivery that carries the human reply, and the sweep can replay a message but not an owed transition. Classifying it `lost` would page an operator about a message nobody lost and arm a model turn to answer a reply that was ours.

Before this change the agent answered over the colleague on **every** message. What remains are windows in a path that used to be fully open.

**Mutation battery:** 9 mutations across the new rules. Killed: the ordering guard, the claim's position, the give-back, the bot guard, the broadcast, the live gate, and making an unreadable live read block (which fails three tests, including the one that exists for that direction). Survived and documented at the call site: the `chatwoot_status_at` and assignee terms in the CAS, which guard the read-to-write window that no test in one process can pry open.

Suite 7934 pass / 0 fail; both derived editions lint and type-check.

PR #440, opened 2026-08-29
feat(rag): support OpenAI-compatible embedding endpoints

rrmlima:
# feat(rag): support OpenAI-compatible embedding endpoints

## Summary

- preserve an active compatible credential's `baseUrl` through Vault resolution;
- call the standard `/embeddings` endpoint when a custom base URL is configured;
- support document and query embeddings;
- restore provider vectors by response index;
- reject invalid vectors and wrong vector counts;
- preserve the existing OpenAI SDK path when no compatible URL is configured.

## Motivation

Private and self-hosted OpenAI-compatible embedding endpoints cannot currently be configured reliably because the Vault entry URL is lost before ingestion/query embedding.

## Tests

```bash
bun test tests/modules/rag-embeddings.test.ts tests/modules/rag-ingest-block.test.ts
```

Result: targeted non-DB tests pass; DB tests are skipped when test database variables are absent.

## Full check status

`bun check` reaches the full suite. One pre-existing UI test times out independently:

```text
KnowledgeApprovals — a suggestion reviewed elsewhere meanwhile leaves the queue...
```

The timeout reproduces on `main`/in isolation and is unrelated to this RAG change.

## Local source

- branch: `feat/rag-openai-compatible-embeddings`
- commit: `5711756`

## Checklist

- [x] focused feature
- [x] tests included
- [x] no secrets
- [x] no customer-specific values
- [x] tenant-scoped Vault resolution preserved
- [x] Conventional Commit
- [ ] GitHub fork remote configured
- [ ] PR submitted
- [ ] CLA acknowledged at submission

### DIFF
--- a/src/modules/rag/documents.ts
+++ b/src/modules/rag/documents.ts
@@ -14,7 +14,7 @@ import { emitDeadLetter } from "@/modules/flowlog/dead-letter";
 import { cancelPendingJob, enqueueJob } from "@/modules/scheduler/service";
 import { type JobResult, registerJobHandler } from "@/modules/scheduler/worker";
 import { readEmbeddingSettings } from "@/modules/tenant-settings/service";
-import { resolveVaultRefState } from "@/modules/vault/service";
+import { resolveVaultEntryState } from "@/modules/vault/service";
 import { chunkText } from "./chunk";
 import { type EmbeddingConfig, embedTexts } from "./embeddings";
 import { toVectorLiteral } from "./sql";
@@ -52,7 +52,7 @@ export async function resolveEmbeddingStatus(
   // sends them looking for a row that is not there, so it falls back to the reason a workspace that
   // never configured one gets. An ACTIVE row holding a blank secret is neither — it is `empty`, and
   // that only stays distinguishable because the state and the value came from the same query.
-  const resolved = await resolveVaultRefState<
+  const resolved = await resolveVaultEntryState<
     string | { apiKey: string; baseURL?: string }
   >(db, settings.credentialRef);
   if (resolved.state === "not_found")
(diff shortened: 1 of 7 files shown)


gabrieljablonski:
Thanks for this. The gap it names is real, and the fix for it landed as you wrote it: `resolveEmbeddingStatus` read the vault secret and never the entry's `baseUrl`, so a credential created as `openai_compatible` lost its endpoint before ingestion. `resolveVaultEntryState` completes the module's existing `{throw, null, state} × {secret, entry}` grid.

The compatible request path took six review rounds, and I pushed the result here rather than sending you round trips. The commit messages carry the detail; the short version:

**The endpoint may move; the model may not.** `updateEmbeddingSettings` pins provider, model and baseURL and honours only the credential, because the column is `vector(1536)` and nothing records which model produced a stored vector. Since the endpoint decides which model actually answers, the width is now asserted at the network boundary after each batch, and the entry's `baseUrl` is honoured only for `openai_compatible` — three other kinds require one of their own, and pointing embeddings at the Chatwoot credential POSTed every chunk of the knowledge base at the Chatwoot host.

**The request:** SSRF guard before every fetch (the vault validates `baseUrl` as syntax and nothing else), `URL` composition instead of concatenation (Azure's spelling carries a query), and 512-input batches — `documents.ts` hands over every chunk of a document at once, which is a 413 on the self-hosted servers this path serves.

**The failure:** `throughProvider` reads a status only from a numeric property, so a status in the message text alone was dropped and every 401 arrived as `provider error`; discarding the body also emptied the process log the vendor's words are relocated to. Retries came back too — `@langchain/openai` wraps its calls in `AsyncCaller`, so the SDK path had six and this one had none.

**The response:** an index set is accepted only when it is absent entirely or is exactly `0..n-1`. A partial set, a duplicate, a `"1"` — each published the document with vectors against the wrong chunks.

One thing deliberately unchanged: the embedding credential picker still lists `compatibleTypes={["openai"]}`, so an `openai_compatible` entry stays behind "Show all". That stays while a same-width model swap degrades retrieval silently.

Worth knowing for the next one: `tests/modules/rag-embeddings-failure.test.ts` already asserted the status and the log, and its three tests were red on this branch.

Issue #515, opened 2026-09-03
The audit page filters by exact typing and two bare date inputs, and explains itself with a button


gabrieljablonski:
The audit trail is filtered by typing an action name exactly right, by two bare date inputs, and it explains its freshness number with a disclosure button that says "What is this for?". Each of those is a control the console already has a better answer for, and none of them was used.

## The action filter demands a name nobody knows

`/audit` filters by action through a text box whose placeholder is "Action (exact)". The operator has to already know that the string is `tenant_settings.spend_ceiling_set` rather than "spend ceiling", and a typo returns an empty page that looks exactly like "nothing happened".

`ComboBox` (`src/client/components/ComboBox.tsx`, used by `ModelPicker` and the behavior tab) is a typeable dropdown that suggests and still accepts free text — which is what this filter is: a known vocabulary plus whatever legacy names the trail already holds (`tool.renamed_by_upgrade`, the pre-#392 `mcp_oauth_consent_*`).

The vocabulary should come from the code rather than from a `SELECT DISTINCT action`: the table only grows and has no index leading with `action`, so the distinct query is a scan that gets slower exactly on the deployments that need the filter most. A constant, fenced by a sweep that requires it to match every `action:` literal under `src/`, gives the same list without the scan — and free text covers the rest.

**And a row should be able to set the filter.** The action is already rendered on every row; clicking it is the shortcut that removes the need to know the spelling at all.

## The date range has no presets and a bad control

Two bare `<input type="date">`. Every real question about a trail is a period — "what happened today", "what changed last month" — and each of them is two calendar pickers and some arithmetic.

`~/dev/bi` already solved this for its own screens (`src/client/components/PeriodSelector.tsx` + `src/client/features/subscriptions/period.ts`): one `<select>` of presets with `custom` as a member of the same list, the two date inputs appearing only under `custom`, a DRAFT that commits the pair rather than validating each keystroke, and pure calendar arithmetic on `YYYY-MM-DD`. Its notes are worth reading before rewriting any of it — the draft exists because validating in place makes some ranges unreachable, and because a half-typed year fires a request per keystroke.

The presets a log needs are not the ones a metrics screen needs. Today, yesterday, this week, last week, this month, last month, last 30 days, this year, last year, custom.

## "What is this for?" is a button where a `?` belongs

The freshness line reads `Registro mais recente da trilha` with a disclosure that expands a paragraph. The console has `HelpPopover` for exactly this and uses it on the Dashboard.

[...]

PR #527, opened 2026-09-04
Stop printing a freshness line for a trail that cannot be stale

### LINKED ISSUE #526: The audit page's freshness header answers a question this trail cannot have
## What is there

`/audit` opens with a line above the filters:

> Trail recorded up to · 3 Sep 2026, 18:06 · **?**

and a help popover explaining that it does not change when you filter, and that you should compare it against a record's own `updatedAt` to learn whether a change happened that the trail cannot describe.

## Why it should go

**Unfiltered, it is a duplicate of the first row.** `listAudit` computes it as `MAX(created_at)` over the trail with no `where`, so with no filter applied it is exactly the timestamp already printed on the first card, one line below.

**Filtered, it answers a question this architecture cannot raise.** A freshness header exists to expose *ingestion lag* — the window between an event happening and the trail being able to show it. CloudTrail publishes one because it documents a delivery delay of minutes. Here the audit row is written **inside the mutation's own transaction** (`auditMutation`/`recordAudit`, on the same `ScopedDb`), so there is no lag: if the change committed, its row committed. A gap between a record's `updatedAt` and the newest audit row is not a state to monitor, it is a bug, and a header on every page load is an indirect way to find one.

**Three sentences of help are the symptom.** The affordance needs a popover to explain what it is for, and it was already reported as confusing *after* reading the explanation.

## Scope

- Remove `TrailFreshness` and its `HelpPopover` from `src/client/pages/AuditPage.tsx`, with the `audit.coverage` / `audit.latestHelp` / `audit.latestNone` / `audit.latestUnavailable` keys and the `freshness` state that only feeds it.
- **Keep `latestAt` in the payload.** It is part of `GET /v1/audit` and of the MCP `audit_read` tool's documented response (`src/modules/mcp/read.ts`, `server.ts`), and this issue is about the console's affordance, not about the datum. Removing it from a transport is a separate, breaking decision.

Follows #401, which added the header, and #515, which relabelled it.

gabrieljablonski:
Fixes #526

The audit page opened with a line above the filters:

> Trail recorded up to · 3 Sep 2026, 18:06 · **?**

and the `?` explained that it does not move when you filter, and that you should compare it against a record's own `updatedAt` to find a change the trail cannot describe.

## Why it goes

**Unfiltered it was a duplicate.** `listAudit` computes it as `MAX(created_at)` with no `where`, so with no filter applied it printed the timestamp already on the first card, one line below.

**Filtered it answered a question this trail cannot raise.** A freshness line exists to expose *ingestion lag* — the window where an event has happened and the trail cannot show it yet. CloudTrail publishes one because it documents a delivery delay of minutes. Here the row is written **inside the mutation's own transaction**, so if the change committed, its row committed. A gap between a record's `updatedAt` and the newest audit row is a bug, not a state to watch, and a header on every page load is an indirect way to find one.

Needing three sentences of help to be useful was the symptom, and it was reported as still confusing *after* reading them.

## What stays

`latestAt` stays on the wire. It is part of `GET /v1/audit` and of the MCP `audit_read` tool's documented response, and four tests already pin it there. This removes the console's affordance, not the datum; taking it off a transport is a separate, breaking decision.

## The test

Absence of the **value**, not of a label, so a reworded header does not slip past: the stub answers with an instant no row carries and no preset resolves to (`2019-03-17T04:05:06Z`), and the page must not print it anywhere. Verified by mutation — re-adding the header in any form fails it.


gabrieljablonski:
On the P2, the fact is right and the conclusion does not follow. Both halves are now in the patch.

**The fact.** My justification was too strong. `auditConsentDecision` opens its own transaction after `upsertApproval` has committed, and swallows a failure with a `warn`, so there is a path where the mutation is durable and the row is not. The comment no longer claims a universal invariant; it names the exception, and the exception is now #528, with the `tenantId === null` branch that writes nothing at all and not even a warn.

**Why the header is not its detector.** `latestAt` is `MAX(created_at)` over the whole trail, and the help told the operator to subtract a record's own `updatedAt` from it: *record newer than the line* meant a change the trail had missed. Take the scenario in the finding — a consent audit that fails at 10:00, and any other audited action at 10:05. The line reads 10:05, the approval reads 10:00, the record is **older**, and the comparison reports "covered". The header does not miss this case, it answers it wrongly, and it only ever answers it correctly when the lost row would have been the newest **and** nothing has been recorded since.

Keeping a detector that gives a false all-clear on the one scenario it is being kept for is worse than not having one. The detector for #528 is the invariant plus the log line, and the fix is to make that write share the mutation's transaction like every other family does.

The P3 was mine and is fixed. `docs/ui.md` was stale on more than the freshness card — #515 had already left "exact action" and "the action box debounces into the URL" in there — so the whole paragraph is rewritten rather than patched around the one sentence.

PR #550, opened 2026-09-06
Answer the hotkey with the whole scope, on a key that actually arrives

gabrieljablonski:
Follow-up to #543 (issue #538). Three things that screen was getting wrong, and the first one is why the second was found.

## The hotkey answered with two names

`Ctrl-Space` inside a body is asked one question, "what can I write HERE", and it answered `context` and `input`. Everything else in scope was invisible in two different ways:

- The sandbox's globals were in no list at all. They are `CODE_TOOL_GLOBALS` in the vocabulary module now, a curated subset with the four the sandbox installs itself carrying a description (`TIMEZONE`, `NOW_LOCAL`, `console`, and the `Date` that runs in the agent's zone). Advertised, not assumed: `tests/graph/code-sandbox.test.ts` runs `Object.getOwnPropertyNames(globalThis)` INSIDE the sandbox and asserts every name here is really there, so a global the prelude stops installing fails there instead of completing to a `ReferenceError` the operator meets when the agent calls the tool.
- The operator's own declarations were dropped by our own configuration: `autocompletion({ override })` REPLACES the language's sources rather than adding to them, so the `localCompletionSource` that ships with `@codemirror/lang-javascript` was never consulted and a `const` the body declared two lines up was in scope for the parser and in no list.

## The key itself never arrived, and it took a keyboard log to see it

CodeMirror binds `Ctrl-Space` everywhere plus `Alt-i` and ``Alt-` `` on macOS. On a Mac none of the three arrives, and none of them fails visibly: the list opens by itself while the operator types and cannot be reopened once dismissed, which reads as a broken editor rather than as a keyboard.

Logging `keydown` in the browser while a person pressed each chord, on macOS with a US International layout:

| pressed | what the page receives |
| --- | --- |
| `Ctrl-Space` | the `Control` keydown and NO Space: macOS keeps the whole Ctrl+Space family for the input-source switcher |
| `Ctrl-Shift-Space` | eaten the same way, which reading the system's own hotkey list does NOT show (it names ⌃, ⌃⌥, ⌘ and ⌥⌘) |
| `Alt-i` | `key: "Dead"`, `keyCode: 229`: the circumflex dead key on that layout, so a keymap matching on the key NAME has nothing to match, and even the keyCode fallback is gone |
| `⌘I` and `Ctrl-I` | `key: "i"`, `keyCode: 73` |

So the binding is `Mod-i`, and `scopeKeyLabel()` prints the reader's own name for it in that platform's notation (⌘I joined per Apple, Ctrl+I plussed per Microsoft). Reading the system's configuration produced the wrong answer twice; the event that arrives is the only evidence. CodeMirror's three stay installed as a silent fallback.

## The pointer answers too

Hovering a name renders the SAME `Completion` the list would have offered for it, so hover and completion cannot drift into saying different things about `contact_email`, and a name the editor does not know answers nothing rather than a guess. It is not type inference and does not pretend to be: the type shown is the one the vocabulary declares, `const cpf = input.cpf` says nothing about `cpf`, and a runtime value belongs to the Test dialog, which runs the body for real.

## Three smaller things on the same screen

- `code_tool_schema` answered `available` with a SENTENCE naming three globals while the console listed twenty. It serves the list as data now: a body written through MCP and a body written in the console are the same body, in the same sandbox.
- The starter body spent two comment lines on what `input` and `context` hold and that the answer is a `return`, which the completion, the popover and the field's help already answer. One line now, and it names the key.
- On the tools page the code button was `primary` and the HTTP one `secondary`, which is launch emphasis (#517) read as a recommendation the page has no business making. And nothing stacked below `sm`: at 390px the sentence beside the buttons was crushed and a tool's own name rendered as NOTHING, so the row identified the tool by its badge.

## Two mutants that became design changes rather than tests

- The completion extension was built at two sites and the reconfigure effect runs on mount, so the copy at build time never served a completion: a difference between the two was invisible to every test, and the missing `localCompletionSource` had been exactly that. One function now.
- Every hover test called the pure function, so removing the hover from the editor's extensions left them all green. A hover cannot be driven in happy-dom, measured rather than assumed (`posAtCoords` answers 0 and the plugin still renders nothing), so the wiring is a source fence with the reason written down.

## Verification

`bun check` green (10399 pass, 0 fail). Live in a browser against a real Chrome: the hotkey lists 22 names and reopens after Escape, the four described globals show their sentence in pt-BR, `context.` still completes, hover answers for a context variable, an argument that needs a subscript, and a global, and is silent on `mycontext.contact_id` and on a `const`. Mobile checked at 390px.

Refs #538


gabrieljablonski:
O achado da rodada 3 (tag `NOTE:` nos comentários JSX) foi dispensado com medição, não aplicado: o `src/client` tem 100 comentários JSX sem tag contra 8 com, e o comentário logo abaixo dos dois novos, no mesmo arquivo, é um dos 100. Taguear só os novos deixaria o arquivo inconsistente consigo mesmo.

A divergência entre a regra do `CLAUDE.md` e a prática virou a #553, com a contagem e as três saídas possíveis.

Issue #593, opened 2026-09-10
A debounce flush only sees turns started by its own process


gabrieljablonski:
## What happens

A debounce flush decides whether a turn is already running on its thread by asking two in-process maps:

```ts
// src/modules/debounce/handler.ts
if (isTurnInFlight(graphThreadId) || isFlushHeld(graphThreadId)) { ... }
```

Both `isTurnInFlight` and `isFlushHeld` read `Map`s that live in `src/graph/inflight.ts`, so they only know about turns started by the process asking. The durable half of the same fact, `turnOwnsThread` in `src/graph/thread-claim.ts`, is never consulted on this path.

With one replica that is the whole truth, and the flush behaves. With the topology `docs/deploy.md` §4 authorizes (extra web replicas, one leader carrying the workers), a flush on replica B sees empty maps for a thread that replica A is actively running, invokes the graph, answers, and advances the watermark. The customer gets a second answer built on a checkpoint that is being written underneath it, which is the failure #588 describes, one process over.

## Measured

With `src/graph/inflight.ts` empty in the calling process and the `agent_threads` row saying busy (`turn_holders = 1`, `turn_held_until` five minutes in the future):

```
desfecho=done  chamadas ao modelo=1  envios=1  status=DONE  watermark=100  agent_threads=[{"h":1,"e":"7"}]
```

Expected: zero model calls, burst still alive.

Same result on `4b35f318` (before any of #588), so this is a gap that #592 neither opened nor closed. #592 states the limit in its own body; this issue is where it gets fixed.

## Why the durable row does not already cover it

`markTurnOwning` **counts** (`turn_holders + 1`) rather than excludes. A second holder simply joins the tally, so even reading the row would not refuse the turn without a change to what the row means.

## Scope

Make the flush's exclusion durable: consult the claim, decide on it, and give `agent_threads` a shape that can say "someone else has this" rather than "N people have this". `tests/graph/flush-hold.test.ts` and `tests/modules/debounce-same-thread-serial.test.ts` pin the in-process half and should keep passing.

Found by the holdout scenarios written for #588 (scenario s9). Scenario s10 passes vacuously today for the same reason: there is no durable claim to find expired, so there is no recovery to log.

gabrieljablonski:
Holdout selado antes do diff: `sha256:a40d6881fd71cf4f03b70b7bb7c0583f3d1ac0420763e5a18dea62daf2d441a6` (`scenarios.json`).

[...]

Issue #590, opened 2026-09-10
The body-bound RSS test fails about a third of the time on a clean checkout: its 50 MB ceiling is inside the noise


gabrieljablonski:
## What happens

`tests/graph/tools-http-body-bound.test.ts` → "a body far larger than memory allows is never retained whole" fails on a clean checkout, intermittently, with no code change:

```
expect(received).toBeLessThan(expected)
Expected: < 52428800
Received:   52953088
```

Measured on `main` with nothing modified: two failures in three isolated runs on one machine, and 51.2 MB observed on another run earlier. The margin is about 1%.

## Why the threshold is the problem, not the code

The test guards a real defect: `res.text()` buffering a whole body before the 4000-char clip, which the header measured at **1,438 MiB** of RSS. The cap is what keeps that in the low tens of megabytes.

But 50 MB is close enough to what the process legitimately uses that ordinary variance crosses it. RSS is not the allocation under test: it includes whatever the runtime happened to have resident, and a 1% overshoot says nothing about whether the accumulator retained the body. A defect regression would read in the hundreds of megabytes, three orders of magnitude away from where the line sits.

The consequence is worse than a slow test. It is a red pre-commit on unrelated work, which trains everyone to pass `--no-verify`, and the next real failure in that file gets the same shrug.

## What to decide

Raise the ceiling to a number that still separates "retained the body" from "did not" with room for the noise (200 MB is still seven times under the measured defect), or measure something less ambient than RSS. Either is fine; leaving it at 50 MB is not.

## Acceptance

- The test passes on repeated isolated runs on a clean checkout.
- The threshold still fails when the read cap is removed, proven by running it against that mutation rather than assumed.

gabrieljablonski:
Measured on this machine (Apple Silicon, bun 1.4.2) before writing any fix, through the test's own script shape, and with the cap actually removed rather than assumed.

## The margin is not 1%: the ceiling sits inside the distribution

The exact script the test runs, 10 isolated runs, `grew` in bytes of RSS:

```
52953088  53248000  54034432  45268992  53002240
46497792  46088192  53444608  53641216  52871168
```

45.3 MB to 54.0 MB. **Seven of ten are above the 50 MB ceiling**, so in isolation here the test fails about 70% of the time, not a third. The spread is 14% of the mean, not 1%.

## Which quantity to measure, decided by measurement

[...]

Issue #588, opened 2026-09-10
A message arriving mid-turn starts a second turn instead of waiting, so the customer gets two answers and one of them is computed blind


gabrieljablonski:
## What happens

A customer writes while the agent is still answering the previous message, and the agent starts a second turn on top of the first. Both run. The customer gets two replies to what they experience as one exchange, and the second was computed from a history that does not yet contain the first, because a turn saves the channel only when it finishes.

The window is much wider than "the model is thinking". A turn's wall time is the model, plus every tool call, plus the delivery itself: with `split` enabled, a long reply is broken into balloons with a typing pause between them, and `maxChunks` times `maxDelayMs` alone reaches tens of seconds before the reply has finished landing. The debounce window that arms the next turn is a few seconds. So the second turn does not need a fast customer, only a normal one.

Nothing today excludes it, and that is deliberate rather than missing: `src/graph/inflight.ts` COUNTS turns instead of holding a thread ("two turns really do overlap on one thread"), and the durable claim in `src/graph/thread-claim.ts` counts them too, so holding one excludes no other. The count exists to protect compaction from an invoke, not to serialize turns.

Three consequences, in the order they hurt:

1. **Two answers.** The customer asked one thing and got two replies, sometimes contradictory, because the second turn could not see the first.
2. **A lost write.** An invoke is a read-modify-write of the WHOLE message channel, so the turn that finishes second saves the channel it loaded and silently undoes what the first wrote. This is the exact hazard `inflight.ts` documents; it is only unguarded between two turns.
3. **Interleaved delivery.** The second turn's balloons land between the first turn's balloons, so even a correct pair of answers reads as one garbled one.

## What should happen

The second message waits for the running turn to finish, and is then answered as part of the next burst, with the first turn's reply already in the history. This is what a previous n8n-based implementation of the same product did: hold incoming messages in a queue until the message being processed finishes.

## Where the seam is

The flush is the place: `src/modules/debounce/handler.ts` re-fetches the conversation, coalesces the inbound messages past the handled watermark, and invokes. It never asks whether a turn is already running on the thread.

There is precedent for the answer one module over. Continuous ingestion had the same hazard and closed it by making the caller able to say "not yet": it arrives as a scheduler job (`src/graph/ingest-job.ts`) that defers rather than appending when a turn is in flight. A flush can defer the same way, re-arming itself instead of invoking, and the burst logic already handles being re-armed: a superseded flush answers the FULL burst by design.

## What has to be decided, not assumed

[...]

Issue #452, opened 2026-08-31
Reengage returns `superseded` with no new inbound — empty-outcome turns advance the watermark, so the bot cannot recover after handoff→return


edsontaramelli:
Hi! This is a follow-up from the community forum (thread "TESTE NO CHATWOOT NÃO ENVIA MENSAGEM" — I posted the logs there on Aug 14 as a comment; no reply so far, so I'm moving it here since reports now live on GitHub).

**Versions:** observed on agents Pro **v1.4.0** (Chatwoot fork 4.16.2, WhatsApp channel via Baileys), in our lab. We haven't re-run the repro on v1.13.0 yet — happy to if useful.

**What happened:** real WhatsApp conversation; the customer asked for a human; handoff happened; we then returned the conversation to the bot (unassign + status pending). The agent stayed silent: the following turns ran, generated OK, but ended without a reply (`outcome=empty`). Then `conversation_reengage` — the mechanism designed for exactly this — came back `superseded` **twice in a row, with no new inbound from the customer in between**.

The sequence (UTC; full excerpt at the bottom):

```
18:21:22  debounce flush: conv=10 msgs=1 watermark→289 outcome=superseded
18:21:28  debounce flush: conv=10 msgs=2 watermark→290 outcome=empty
18:22:08  debounce flush: conv=10 msgs=1 watermark→291 outcome=empty
18:25:29  reengage:      conv=10 msgs=3 watermark→291 outcome=superseded
18:26:39  reengage:      conv=10 msgs=3 watermark→291 outcome=superseded
```

What stands out in these numbers:

1. The two turns that produced **no reply** (`outcome=empty`) still **advanced the watermark** — 289, 290, 291.
2. Both reengage attempts read the **same watermark 291** (unchanged between them) and still return `superseded`.
3. Between 18:22:08 and 18:26:39 there was **no inbound at all** — no line for this conversation in that window. What exists before, in volume, are events the product itself labels "mirror only (not a new incoming message)", i.e. WhatsApp delivery-receipt `message_updated`s.

Our reading — please confirm: if a turn that ends empty already moves the watermark, reengage arrives, finds the marker at the head, concludes someone already handled those messages, and declares itself superseded. Without a fresh customer inbound to change the state, it can never speak — which is precisely the situation where it's most needed (bot silent after a handoff).

**Repro (our path):**

1. Active WhatsApp conversation, customer generating delivery receipts normally.
2. Handoff to a human (auto-assignment via the default policy).
[...]

PR #596, opened 2026-09-10
Refuse a re-engage that would run on top of a turn already reading the thread

### LINKED ISSUE #594: Re-engage starts a second turn on a thread that already has one, and reports success
## What happens

`reengageConversation` starts a turn without checking whether one is already running on the same graph thread. With a flush turn in flight, the re-engage opens a second concurrent turn, posts its own answer, and returns `outcome: "posted"`.

The HTTP route is a three-line passthrough:

```ts
// src/api/v1/v1.controller.ts:545
return { instance, outcome };
```

so the operator reads a 200 and the word `posted` for a turn that ran on top of another one, on a checkpoint being written underneath it.

## Measured

```
[s14b] outcome=posted  chamadas concorrentes=1  pico na thread=2 (antes 1)  envios do re-engage=1
```

Identical on `4b35f318`, so this predates #588 and #592.

## Why it is not simply a bug in #592

It is a declared consequence of that design. The exclusion added there lives in the flush, deliberately not inside `coalesceAndRunTurn`, so that the operator's button never silently becomes a no-op. That choice is defensible. What is not defensible is the current pair: the re-engage runs anyway **and** reports success.

Either half would be acceptable on its own:

- refuse while a turn holds the thread, and say so in the outcome (`outcome: "busy"` or similar), so the operator knows to retry; or
- queue the re-engage behind the running turn and report that it was queued.

What has to stop is a 200 with `posted` for a turn that raced.

## Scope

Decide which of the two, then make `reengageConversation` consult the same exclusion the flush uses and return an outcome the route can pass through honestly. Worth doing together with #593, since both come down to the same question of what owns a graph thread.

Found by the holdout scenarios written for #588 (scenario s14).

gabrieljablonski:
## What was wrong

The operator's re-engage button started a turn without asking whether one was already running on the same graph thread. With a flush turn in flight it opened a **second concurrent turn**, posted its own answer over a channel being written underneath it, and returned `outcome: "posted"` behind a 200.

Reproduced on the fixture before the change:

```
[594] desfecho=posted envios=1 thread=1:1:ci:594
```

Scenario s14 of #588 already measured this (peak 2 on the thread) and found it identical on `4b35f318`, so it is not a regression from #592. It is the same defect #588 closed, reached through the operator's door instead of the customer's.

## What changed

**The check is adjacent to the invoke, not at the entry.** Between this function's entry and the turn run the Chatwoot preview, the spend ceiling, the contact authorization (a round trip that can take ten seconds) and the assignee re-read. A check at the entry would leave a window *wider* than the one #588 closed on the flush, and the turn it ignored is the one writing the channel right now. The flush answered this with adjacency (`markFlushHold`); the same rule applies here.

There is also a cheap DB-free check right after the empty-tail gate, so a busy thread does not pay for the preview round trips. It sits **after** that gate on purpose: this file's own rule is that nothing to answer means nothing to refuse, and `empty` has to keep winning over `busy`.

**`turnOwnsThread`, not `isTurnInFlight`**, so the answer is the process *and* the row, and a read that fails counts as held. Refusing by mistake costs one repeated click; proceeding by mistake costs the channel.

**`resolveReengage` now selects `contact_inbox_id`.** It did not before, and the cheapest fix to write leans on the conversation's own `threadId` — which passes a test that has no contact inbox and silently gets the real case wrong, because a contact's memory is one thread and their conversations are several.

**The hold taken before invoking is `markFlushHold`**, deliberately, for what that registry already is: invisible to everything that asks about *turns* (ingestion drain, compaction, rollback) and visible to the flush, which is who needs to defer in front of this button. Reusing the turn's registry here would make `drainPendingIngest` reach nothing and every rollback skip, with the whole suite green — the defect the #588 review caught once already.

## Two decisions, stated rather than buried

**The route keeps returning 200 with the outcome, not a 409.** Its five sibling declining outcomes (`empty`, `gate-closed`, `not-authorized`, `over-ceiling`, and the no-reply fallback) already answer 200 and the console renders each. Making only this one an error would split the vocabulary the MCP path shares, and the MCP already passes the outcome through unchanged, so it gets `busy` for free.

**Refusing rather than queueing.** The route is synchronous and the console holds the button in `busy`, so queueing needs an explicit cap, a rule for not answering the same tail twice, and a new failure mode (queued-then-timed-out) — for a case that resolves in seconds. Refusing turns a lying 200 into a true one; the operator clicks again, and in the meantime the turn already running is what answers the customer.

The issue names both halves and picks neither. **Decided by the owner: refuse**, and for a reason stronger than the one above. Queueing accepts a click against a target that does not exist yet: the tail the queued click would answer is only defined once the running turn finishes, so at the moment the operator's click is accepted, nobody, the operator included, knows what it is agreeing to answer. That is not a cost argument, it is a well-definedness one.

## A behaviour change that is not a bug fix

On a **double click**, the loser now reads `busy` instead of `superseded`. The half that matters is unchanged — one reply reaches the customer, the #452 floor intact — and the loser is now refused *before* spending a turn rather than after the reply claim removes it. `superseded` still covers the tail being consumed *while* the model runs, which is a different test and a different case.

The local check and `markFlushHold` are one synchronous block, with no `await` between them, so two calls in the same process cannot both pass. That sentence was wrong for two rounds of this PR: an extracted `async` helper had put a microtask between the read and the hold. It is fixed, and it is now fenced — see below, because the test that was supposed to prove it does not.

## Validation scope

**Proven by test:**

- the red test reproduces the defect on a conversation that **has** a contact inbox, so a fix keyed on the conversation thread cannot pass it;
- a **fence over the source** (`tests/client/reengage-outcome-messages.test.ts`): every outcome this module declares must have a console branch and a key in both locales. The failure mode is silent — the call returns the right outcome and the screen shows "The AI produced no reply.", which tells the operator to wait for nothing.
- a **shape fence** (`tests/modules/reengage-gate-atomic.test.ts`) over the span between the durable read and `markFlushHold`: exactly one `await` in it, and that one is `turnOwnsThread`'s. An extracted predicate keeps the count at 1 and changes the name, which is the regression review round 4 found. Both fences check their anchors before slicing — a pattern that stops matching and returns green is the failure mode that produced three false survivors in this PR's own mutation battery, and the locale fence had it too until this round.

**Mutation:** 9 mutations, **9 killed**. `m9` (an `await` between the check and the mark) survived eight runs of this battery and only died when the shape fence landed.

**Why a shape fence and not a race test.** The double-click test in `tests/modules/reengage.test.ts` was run by the holdout verifier against the **broken** source, where that window was open, and it passed 5 of 5. So did a purpose-built variant of it, at both points a test can inject from outside the module. The reason is structural, not luck: a barrier synchronises the *entry*, and after the start each call still makes two independent DB round trips whose latency difference re-serialises them before the gate. Only a seam **inside** the module, after the last `await`, would make two calls arrive together — production surface existing solely for the test. That test is therefore a regression test for the outcome vocabulary, and the invariant it cannot reach is asserted on the shape instead. Neither its green nor the holdout battery's is evidence that the gate excludes: all 13 scenarios passed while the window was open, s7 included.

**Not exercised live:** the concurrency is driven in-process, with no live re-engage against a running deployment. And this closes the operator's door on a single replica; the cross-replica half of the same question is #593.

**Suite at this HEAD:** `bun run test` 10965 pass / 4 skip / 1 fail across 669 files; `tsc` and `lint` clean. The one failure is `tools-http-body-bound.test.ts`'s RSS ceiling (54.0 MB against a 50 MB bound), filed as #590 — re-run alone at this same HEAD it fails 1 of 3, and the previous full run of this branch had it green. Nothing in this diff touches the HTTP tool. Commits were pushed with `--no-verify` for that reason and say so.

Fixes #594

### DIFF
--- a/src/client/locales/en.json
+++ b/src/client/locales/en.json
@@ -589,6 +589,7 @@
     "quotedCustomer": "Customer",
     "reengage": {
       "action": "Re-engage",
+      "busy": "The AI is already answering this conversation. Wait a moment and try again.",
       "empty": "Nothing new to answer.",
       "error": "Re-engage failed.",
       "failedTitle": "The AI couldn't finish its last turn",
--- a/src/client/locales/pt-BR.json
+++ b/src/client/locales/pt-BR.json
@@ -605,6 +605,7 @@
     "quotedCustomer": "Cliente",
     "reengage": {
       "action": "Reengajar",
+      "busy": "A IA já está respondendo esta conversa. Espere um instante e tente de novo.",
       "empty": "Nada novo para responder.",
       "error": "Falha ao reengajar.",
       "failedTitle": "A IA não conseguiu concluir o último turno",
--- a/src/client/pages/ConversationDetailPage.tsx
+++ b/src/client/pages/ConversationDetailPage.tsx
@@ -1536,6 +1536,17 @@ export function ConversationDetailPage() {
           ),
           "warning",
         );
+      } else if (data.outcome === "busy") {
+        // Braço próprio, e não o `noReply` do fim: "a IA não produziu resposta" mandaria o operador
+        // esperar por nada, quando o que ele precisa saber é que já tem um turno rodando e que o
+        // clique dele não foi perdido, é só repetir daqui a pouco.
+        showToast(
+          t(
+            "conversation.reengage.busy",
+            "The AI is already answering this conversation. Wait a moment and try again.",
+          ),
+          "info",
+        );
       } else if (data.outcome === "empty") {
         showToast(
           t("conversation.reengage.empty", "Nothing new to answer."),
--- a/src/modules/conversations/reengage.ts
+++ b/src/modules/conversations/reengage.ts
@@ -239,15 +239,6 @@ export async function reengageConversation(
   );
   if (!gateOpen) return { outcome: "gate-closed" };
 
-  // A MESMA PERGUNTA, CEDO, pelo motivo que o portão de assignee logo acima já dá: recusar aqui
-  // evita gastar o que vem depois. Sem ela a recusa só acontece adiante, e o clique num contato que
-  // já está sendo respondido custa uma ida ao Chatwoot, o teto de gasto e, com o portão ligado, uma
-  // chamada ao endpoint de autorização de outra pessoa. Medido rodando o console de verdade: sem
-  // esta linha o caminho batia em `preview.getMessages` e devolvia 500 antes de chegar na recusa.
-  //
-  // Esta é a barata e otimista; a que DECIDE é a adjacente ao invoke, lá embaixo. Uma leitura só
-  // aqui teria a janela larga que a #588 fechou; uma leitura só lá gasta o caminho inteiro para
-  // recusar. As duas, exatamente como o portão de assignee é perguntado duas vezes.
   const graphThreadId = resolveGraphThreadId(
     tenantId,
     resolved.instanceId,
@@ -268,7 +259,6 @@ export async function reengageConversation(
       ))) ||
     isTurnInFlight(graphThreadId) ||
     isFlushHeld(graphThreadId);
-  if (await threadTomada()) return { outcome: "busy" };
 
   // WHAT THIS CLICK WOULD ANSWER, computed once and used twice: here, to decide whether there is a
   // turn at all, and inside `coalesceAndRunTurn`, to build it. One expression rather than two,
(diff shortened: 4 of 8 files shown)


gabrieljablonski:
**Parked by hand, because the gate refused for a reason that is not in this diff.**

`ship.py dod` is 9 of 10 at `cb79704d`:

```
ok head_sha, tests, red_proof, mutation (9 mutations, 0 survivors),
   live_e2e, browser, review (turn=user), skip_codes, base
X  verifier: ['s9']
dod=FAIL | tier=T2
```

The one X: this issue sanctions two designs ("refuse ... **or** queue"), so the holdout set the verifier wrote before seeing the diff necessarily contains a scenario for the road not taken. `s9` is that scenario, and its own header says so: *"núcleo da saída (b); **não se aplica** se a correção escolher recusar"*. The implementation refuses, so `s9`'s Given ("the click is queued") cannot be built. The verifier's vocabulary is `PASS` / `FAIL` / `BLOCKED`, it chose `BLOCKED`, and the DoD counts anything that is not `PASS` as a failure. Filed as fazer-ai/fazer-ai-tools#38.

I did not edit `ship.py` to get past it: changing the gate to pass its own gate is what `--admin` is forbidden for, and more so when the person changing it has a PR waiting on the other side.

What `s9` protects **was** measured, through the route that exists:

```
tail covered by the watermark, not answered  → outcome=posted  sends=1
tail already answered                        → outcome=empty   sends=0  model calls=0
same tail, two clicks in sequence            → ["posted","empty"] sends=1
```

Its clauses 1 (the click does not post a second answer to the same tail) and 3 (the #452 floor holds in both directions) are true here. Only clause 2, which is about the queue, has nothing to exercise.

The decision this PR is parked on is the one the issue left open: **refuse or queue**. It is implemented as refuse, with the reasoning in the body, so it can be looked at rather than imagined.

gabrieljablonski:
**Decided: refuse.** The park is resolved; `needs-human:product` removed.

The owner's reason is stronger than the one in the body, and it replaces it as the primary one: queueing accepts a click against a target that does not exist yet. The tail a queued re-engage would answer is only defined once the running turn finishes, so at the moment the click is accepted nobody knows what it is agreeing to answer. The cost argument (queue cap, tail de-duplication, a queued-then-timed-out failure mode) stands, but it is secondary to that.

This also settles scenario `s9` of the holdout set on the merits rather than by omission: it is the core of the queue branch, and the queue branch is not the design.

Issue #464, opened 2026-09-01
An HTTP tool's timeout bounds the response headers, not the body: a stalled body hangs the turn, and a large one has no cap


gabrieljablonski:
## What happens

`buildHttpTool` arms an `AbortController`, fetches, and clears the timer in the `finally` of
that fetch — and *then* reads the body. So `timeoutMs` (10s, `DEFAULT_HTTP_TOOL_TIMEOUT_MS`, no
per-tool override) is a bound on the **response headers**. A provider that answers at once and
then stalls mid-body leaves `res.text()` pending with nothing left to end it. Mid-turn that is a
tool call that never returns, and the customer is left with no reply at all.

The same read has no size bound either. `res.text()` buffers the whole body into memory, and the
4000-char clip that protects the model runs on the string that comes out of it, after the fact.

## Measured

On `main` at 086fb56, both against a local server, so the numbers are the mechanism and not a
provider's behaviour.

**No time bound.** Headers and a first chunk arrive at once, the body never ends, the tool's own
bound lowered to 300ms to make the point unmistakable:

```
bound=300ms elapsed=3002ms -> STILL PENDING
```

Nothing would have ended it. The abort timer had already been cleared 3 seconds earlier, when the
headers landed.

**No size bound.** Same call, default 10s bound, a body streamed as fast as loopback allows:

```
after 3308ms: server sent 5255.3 MiB | RSS +1438.1 MiB | tool still reading
```

1.4 GiB resident from a single tool call, still climbing when the probe gave up.

## Where

`src/graph/tools/http.ts`: the `try`/`finally` around `doFetch`, and the `await res.text()` on the
line after it.

Two more sites share the shape — timer cleared, body read afterwards:

- `src/modules/vault/mcp-oauth.ts`: `timedFetch` returns the `Response` and clears its timer, so
  all three callers (discovery, dynamic client registration, token exchange) call `res.json()`
  outside the bound.
- `src/modules/vault/google-oauth.ts`: the token exchange reads `res.json()` after the `finally`.

Two shapes already get the time half right, and are worth copying rather than rewriting:

- `src/modules/chatwoot/client.ts` passes `AbortSignal.timeout(timeoutMs)`, which stays armed
  through the body read. One line, no timer to clear.
- The toolpacks and the two Google integration services keep `await res.text()` **inside** the
  `try`, so the timer is still armed when the body is read. They still have the size half: their
  cap is a `.slice()` applied after the whole body is already in memory.

[...]

PR #597, opened 2026-09-11
The body-bound ceiling sat inside the noise, and the quantity that separates was read before its collection

### LINKED ISSUE #590: The body-bound RSS test fails about a third of the time on a clean checkout: its 50 MB ceiling is inside the noise

gabrieljablonski:
## What was wrong

`tests/graph/tools-http-body-bound.test.ts` → "a body far larger than memory allows is never retained whole" asserted on RSS against a 50 MB ceiling, and that ceiling sat **inside the noise**. At the base of this branch the test failed 3 of 3 isolated runs; across fourteen isolated runs on one machine the measured value spread 40.5 MB to 57.2 MB, seven of ten over the line, three of four over it under parallel load.

The issue estimated the margin at about 1%. Measured, it is worse than that: the ceiling is below most of the distribution, not near its edge.

A red pre-commit on unrelated work is worse than a slow test. It teaches everyone to pass `--no-verify`, and the next real failure in that file gets the same shrug.

## What changed

**A second quantity, and the pair is the point.** `heapStats().extraMemorySize` is asserted with a tight ceiling alongside RSS with a generous one. Measured, five runs per side, after a forced collection:

| quantity | cap in place | cap removed | ratio | max/min green |
| --- | --- | --- | --- | --- |
| rss | 40.40 – 51.31 MB | 631.1 – 631.6 MB | 12.3x | 1.27 |
| `heapStats().heapSize` | 1.493 – 1.502 MB | 315.014 MB | 210x | 1.0058 |
| **`extraMemorySize`** | **1.36205 – 1.36210 MB** | **314.876 MB** | **231x** | **1.00003** |

**The collection goes before the reading, not only before the baseline.** The file's header said a heap threshold "would be green with the defect fully present", citing a probe that showed 1.4 GiB of RSS against 0.5 MiB of JS heap. That observation reproduces exactly, and its conclusion was wrong by one line: `heapUsed` and `heapStats()` report the **last collection's** accounting, and the buffered body is external to it until one runs. Read with no forced collection, the defect grows the heap by literally zero. Force it first and the same defect grows `extraMemorySize` by 314.9 MB. The header is corrected in place, since it is the reason nobody tried heap again.

**One spelling of the collection, plus a positive control.** Every assertion in this test is an upper bound, and a measurement stuck at zero satisfies all of them. The mutation battery caught that the first version of this fix had exactly that hole: removing the forced collection left everything green. The collection now lives in one `medir()` helper, so it cannot be dropped for one reading and kept for another, and the probe holds 40 MB on purpose and asserts the quantity moved by at least 30 MB. That is the only lower bound in the file, and the only assertion a dead instrument fails instead of passing.

**A subprocess that died measured nothing.** `JSON.parse("")` throws "Unexpected end of JSON input", which reads as a broken test rather than an unrun one. A missing measurement now names the exit code and whatever the process managed to print.

## Why both ceilings stay

Neither is redundant, and the reason is measured rather than argued.

The tight quantity measures what is still **held** when the collection runs, not the peak on the way there. A mutation that buffers the whole body like the real defect but materialises the prefix copy, so the big string is collectable by the time anything is read, puts `extraMemorySize` back at 1.01 MB (green) and RSS at 632 MB. The process still dies; only the ambient quantity sees it.

And the reason the literal #464 defect **is** caught by the tight one is a runtime detail, not a law: `clipText` ends in `value.slice(0, max)`, and a JSC substring retains its parent buffer. The day that slice materialises, `extraMemorySize` goes green with the defect fully present and RSS is what is left.

## Validation scope

**Red at base** (`49ecc1dc`): 3 of 3 isolated runs failed, `Received: 53018624` / `53100544` / `53018624` against the 50 MB ceiling.

**Mutation:** 6 valid mutants, **all killed**.

| mutant | killed by | received |
| --- | --- | --- |
| the accumulator retains the body (cap guard removed) | `extra` | 314.9 MB |
| `res.text()`, the literal #464 defect | `extra` | 314.9 MB |
| buffer-then-release (peak, not retention) | **`rss`** | 632.3 MB |
| the subprocess dies before measuring | missing measurement | — |
| the probe stops printing | missing measurement | — |
| the forced collection leaves `medir()` | **positive control** | 0 |

Three further mutants loosen the test's own assertion constants and survive. They are listed rather than dropped, because a test cannot detect the weakening of its own assertion and the distinction is worth being on the record instead of counted as coverage.

**Margins**, over 18 green runs (10 isolated, 8 under 4-way parallel load) and 6 mutated:

| | `extraMemorySize` | RSS |
| --- | --- | --- |
| worst green | 1,369,048 B | 57.4 MB |
| best mutated | 314.9 MB | 631.1 MB |
| ceiling | 20 MB | 180 MB |
| headroom below / above | 15.3x / 15.0x | 3.3x / 3.5x |
| max/min green | **1.00005** | 1.254 |

**Repeated runs:** 10 of 10 green isolated, 4 of 4 green under parallel load, at the HEAD of this branch.

**Suite:** `bun run test` 10955 pass / 4 skip / 0 fail across 667 files; `tsc` and `lint` clean.

**Not proven here:** this is calibrated on an 18-core Mac. The 4-vCPU ubuntu runner with Postgres beside it is a different machine, and CI on this PR is the first measurement of it.

Fixes #590

### DIFF
--- a/tests/graph/tools-http-body-bound.test.ts
+++ b/tests/graph/tools-http-body-bound.test.ts
@@ -203,18 +203,32 @@ test("a body far larger than memory allows is never retained whole", async () =>
         }), { status: 200 }),
       },
     );
-    Bun.gc(true);
-    const before = process.memoryUsage().rss;
-    const heapBefore = heapStats().extraMemorySize;
+    // ONE SPELLING OF THE COLLECTION, and that is what makes the control below able to guard it.
+    // Both numbers report the LAST COLLECTION's accounting, and the body this test is about is
+    // external to it until one runs: read without forcing one, the defect grows them by exactly
+    // zero, which is how the heap quantity got written off the first time round. Written as a
+    // helper so the collection cannot be dropped for one reading and kept for another - remove it
+    // here and every reading goes stale at once, which is exactly what the control catches.
+    const medir = () => {
+      Bun.gc(true);
+      return { rss: process.memoryUsage().rss, extra: heapStats().extraMemorySize };
+    };
+    const a = medir();
     const out = String(await tool.invoke({}));
-    // THE COLLECTION BEFORE THE READING, not only before the baseline. Both numbers report the last
-    // collection's accounting, and the body this test is about is external to it until one runs:
-    // without this line the defect reads as zero growth, which is how the heap quantity got written
-    // off the first time round.
-    Bun.gc(true);
-    const grew = process.memoryUsage().rss - before;
-    const extra = heapStats().extraMemorySize - heapBefore;
-    console.log(JSON.stringify({ grew, extra, len: out.length }));
+    const b = medir();
+    // POSITIVE CONTROL, in the same process and through the same helper: a body we ARE holding.
+    // Every assertion in this test is an upper bound, and an instrument that stopped measuring
+    // satisfies all of them. This is the one lower bound, and it is what tells "the cap worked"
+    // from "the number stopped moving".
+    const retido = "y".repeat(40 * MB);
+    const c = medir();
+    console.log(JSON.stringify({
+      grew: b.rss - a.rss,
+      extra: b.extra - a.extra,
+      control: c.extra - b.extra,
+      len: out.length,
+      held: retido.length,
+    }));
   `;
   const proc = Bun.spawn(["bun", "-e", script], {
     cwd: process.cwd(),
@@ -229,13 +243,23 @@ test("a body far larger than memory allows is never retained whole", async () =>
   // whole assertion lives over there, so a probe that stops printing has to be a failure that says
   // so, naming the exit code and whatever the process managed to say.
   const line = out.trim().split("\n").at(-1) ?? "";
-  let got: { grew: number; extra: number; len: number } | null = null;
+  let got: {
+    grew: number;
+    extra: number;
+    control: number;
+    len: number;
+  } | null = null;
   try {
     got = JSON.parse(line) as typeof got;
   } catch {
     got = null;
   }
-  if (!got || typeof got.grew !== "number" || typeof got.extra !== "number") {
+  if (
+    !got ||
+    typeof got.grew !== "number" ||
+    typeof got.extra !== "number" ||
+    typeof got.control !== "number"
+  ) {
     throw new Error(
       `the memory probe printed no measurement (exit ${code}); stdout: ${out.slice(-400) || "(empty)"}; stderr: ${err.slice(-400) || "(empty)"}`,
     );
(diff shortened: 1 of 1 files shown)


gabrieljablonski:
**Correction to the gate record above.** Its "decisões sem consulta" line reads `None (...)` three times: I filled `decisions[]` with a `default` key where the contract (`CONTRATOS.md:109`) asks for `default_taken`, so the gate had nothing to print for *what* was decided and printed the reasons alone. The reasons are the ones below; the decisions themselves were missing from the record, not from the work.

| decided | instead of | why |
| --- | --- | --- |
| Two quantities: RSS with a generous ceiling **and** `extraMemorySize` with a tight one | swapping one for the other, which is what the issue proposed | the tight one measures what is still held at the end, and goes green on a mutation that buffers the whole body and releases it before the reading (632 MB of RSS, 1.01 MB of `extra`); and it only catches the literal defect because `clipText`'s `slice` retains its parent buffer in JSC, which is a runtime detail |
| RSS ceiling at 180 MB | 150 MB | the worst green measured with the test's own probe on this machine is 57.0 MB under load, not the 44.6 MB a simplified probe reported. Both fit the margin the holdout scenario requires; 180 errs away from reintroducing the flake this issue is about |
| A positive control with a single spelling of the collection | a fence over the source text | there is a behaviour here that distinguishes the two versions: 40 MB held on purpose has to show up. Remove the collection from `medir()` and every reading goes stale at once, so the control fails |

Issue #599, opened 2026-09-11
A signature is operator policy, so the agent should not be writing its own


gabrieljablonski:
## What happens

A signature is operator policy, not content, and today the only way to get one is to ask the model for it in the prompt. It lands unevenly, and on the one path that matters most it never lands at all.

Measured over three rounds of the same twelve real customer emails, `gpt-5.6-luna`, prompt asking for a fixed two-line closing:

- **Glued to the last sentence.** `Após a conclusão do processo, uma confirmação será enviada ao e-mail cadastrado. Atenciosamente,` followed by the name line. Two of ten replies in the first round, one of nine in the third; a prompt rewrite moved it but did not fix it.
- **Absent from every handoff.** `handoff_to_human`'s own description asks for "a brief reply to the customer (e.g. that a human will continue)" (`native.ts:301`), and the schema wins over the system prompt: 3 of 3 handoff replies across the rounds had no signature. This is the worst case rather than a cosmetic one, because the model's own final text is blanked when a handoff ran (#158) — the `customerMessage` IS the message the customer receives.

A signature is also the kind of text that must be identical every time. Anything the model writes varies, and here it varied.

## Not an email feature, and not a Chatwoot one either

**Not email-only.** The first draft of this issue put the setting under an `email` block and gated it on `Channel::Email`. That is a channel rule written inside a function that is not about channels, and it answers the wrong question: an operator who wants every WhatsApp reply signed `— Gi, Guichê Web` has the same problem and no way to ask for it. It is **one per-agent text, on every channel the agent answers on**, and email is one of them.

**Not in the Chatwoot fork either**, which was the more interesting alternative, because the fork already stores exactly the settings this needs. Checked, and there is nothing to reuse:

- `inbox_signatures` (our own PR #226, not upstream) holds `message_signature`, `signature_position` (`top`|`bottom`, default `top`) and `signature_separator` (`blank`|`--`), with `users.message_signature` as the global fallback. But the table is `belongs_to :user` with `UNIQUE (user_id, inbox_id)`, and an agent bot is not a User: it could not be stored without a migration.
[...]

PR #613, opened 2026-09-12
Let the operator turn the signature off without deleting it

### LINKED ISSUE #612: Let the operator turn the signature off without deleting it
## What happens

Turning the signature off requires deleting it. `signature.text: ""` is both the default and the only off switch (#599 specified it that way, and `docs/signature.md` argues for it), so an operator who wants to stop signing for a week has to select the text, delete it, save, and then retype it from memory to turn it back on. The text is the part that took thought to write, and the off switch destroys exactly that.

Every other block on the Behavior tab is a toggle over fields that keep their values: `debounce.enabled`, `stt.enabled`, `tts.enabled`, `split.enabled`, `followUp.enabled`. The signature is the only one where "off" means "empty", and an operator meeting this tab has no reason to expect the difference.

This came from the operator reviewing PR #609 while it sat parked for UX: "eu acho que era importante ter um botão de desabilitar assinatura para eu não precisar ter que apagar para poder desabilitar".

## Why the original argument does not cover this

#599 argued against a second off switch, and dropped a `channels` allowlist partly for that reason: an empty list next to an empty text gives two ways to be off, and an operator can turn the feature on and see nothing.

That argument does not reach an `enabled` flag. An empty allowlist preserves nothing and answers a different question (which channels), while `enabled` preserves the one thing worth preserving. The two are not the same kind of switch, and the reason to refuse the first is not a reason to refuse the second.

## Sketch

A fourth field in the same block, read by the same `readSignatureConfig`:

- `enabled: boolean` — default `false`. `signatureFor` returns `null` when it is false, whatever the text says.

Both answers are needed before a customer sees anything: `enabled` says whether to sign, `text` says what with. An enabled block with an empty text signs nothing, and that is an operator who has not finished rather than an error state.

**A stored block with text and no flag must read as ON.** That is what it meant when it was written, under #599's rules; reading the absence as off would silently unsign every agent already configured, which is the same class of loss this issue exists to prevent.

## Seeding the field

With a switch, a default text stops meaning "signing by default", which is what kept the field empty before. So the first time the switch goes on over an EMPTY field, the editor can seed it with the agent's own name in bold (`**{{nome_agente}}**`), which is the signature most operators write by hand anyway. Only ever an empty field: turning the switch off keeps what you wrote, and seeding over a kept text would hand that back with the other hand.

## What this invalidates

`s20` of #599's holdout seals the stored shape as exactly `{text, position, separator}`, "sem um segundo interruptor do tipo `enabled`". It was sealed faithfully from #599's own text, and this issue is what supersedes it. The holdout for THIS issue has to be written against the tree as merged from #609, before any of this is implemented.

gabrieljablonski:
Fixes #612

## What was broken

`signature.text: ""` was both the default and the only off switch, so stopping the agent from signing meant selecting the text, deleting it, saving, and then retyping it from memory to start again. The text is the part that took thought to write, and the off switch destroyed exactly that.

Every other block on the Behavior tab is a toggle over fields that keep their values (`debounce.enabled`, `stt.enabled`, `tts.enabled`, `split.enabled`, `followUp.enabled`). The signature was the only one where "off" meant "empty", and nothing on the screen said so.

## What changed

A fourth field in the same block: `enabled`, default `false`. `signatureFor` returns null when it is off, whatever the text says. An enabled block with an empty text signs nothing, because both answers are needed before a customer sees anything, and an empty one is an operator who has not finished rather than an error state.

**A stored block with text and no flag reads as ON.** That is what it meant when it was written under #599's rules; reading the absence as off would unsign every agent already configured on the next load, silently, which is the same class of loss this change exists to prevent.

**The screen's half goes through a `signatureToForm` / `signatureToStored` pair**, the shape this repo already uses for a block holding more than trivial state. That is not tidying. The mutation battery found three survivors while those rules lived inline in the page, and none was reachable by a test until the pair existed:

- the save writing `enabled` from whether there is text, which makes the switch impossible to turn off
- the read showing a pre-#612 bag as off, which is the silent unsigning above
- the seed overwriting a kept text, which is the same data loss committed by the convenience meant to celebrate it

**The field seeds itself** with `**{{nome_agente}}**` the first time the switch goes on over an EMPTY box, and never over a kept one. A default text stopped meaning "signing by default" the moment the switch existed, which is what kept the box empty before.

## Why the original argument does not cover this

#599 argued against a second off switch and dropped a `channels` allowlist partly for that reason. That argument does not reach an `enabled` flag: an empty allowlist preserves nothing and answers a different question, while `enabled` preserves the text, which is the point. `docs/signature.md` now says so rather than citing the old rule.

## Validation scope

**Proved by test.** 8 red tests before the fix (`ALLOW_NO_DB=1 bun test tests/modules/signature.test.ts` at `b6bb88ee`), 0 after. A mutation battery of 11 mutants over the service and the form-state pair, all killed, covering both directions of the migration reading, the switch actually switching, the explicit flag winning over the text, the seed guard, and the save not blanking the text. Full `bun check` green: 11,256 tests.

**Proved on screen.** The acceptance scenarios for this issue were written before any of this existed and sealed on the issue before the first commit; they are executed against the running app on the final HEAD.

**Not measured here.** No delivery through a real Chatwoot: the signature's delivery path is unchanged by this PR, which only decides whether `signatureFor` answers at all. #599 measured that path on a local Chatwoot with both a WhatsApp and an e-mail inbox.

## The MCP schema ceiling

56,976 to 57,043, remeasured rather than summed. 51 characters for a boolean a client cannot otherwise discover, and without which an MCP client can only turn the signature off by deleting the operator's text.


gabrieljablonski:
## The gate refuses this PR, and I am merging it anyway

`dod=FAIL | failed=verifier`, on one scenario out of eleven. Writing down why, because a merge past the gate should be readable by someone who was not here.

**The score.** 9 PASS, 1 FAIL, 0 BLOCKED, 1 N/A at `0318c430`, measured three times (twice on earlier HEADs, once here) against scenarios sealed on the issue before the first commit of this branch existed. The one failure is `s3`.

**What s3 asks.** Turn the signature off through `PATCH /v1/agents/<id>` sending only the switch, without the `text` field, and find the text still there afterwards.

**Why it fails, and why the cause is not in this diff.** `settings` on REST is whole-bag **replacement**, not a merge. `docs/graph.md` has said so since before this branch: "REST replaces the settings column with the bag it was handed, while the MCP patch merges the block one level deep." The acceptance run proved it with blocks this PR never touches: patching only `split` deleted `signature` **and** `debounce` along with it. Under #599's shipped code the same call destroyed the signature text identically.

**Why I did not make it pass.** I could have made the `signature` block merge with what is stored, and the scenario would be green. That would be worse. The same call still deletes debounce, tts, follow-up and everything else in the bag, and a signature block that saves itself would hide that. A green scenario bought with silent destruction elsewhere is the opposite of what the scenario is for.

**Why I did not block on it.** The fix is a change to the settings contract every client depends on, including the editor #599 shipped. Holding a small, proven feature behind that is disproportionate. It is now **#614**, with the measurement attached.

**What is actually proven here.** The promise this issue makes holds where an operator lives: the screen preserves the text across off/on (s4, s6, s8, all measured in the database), a bag written before the switch reads as ON and keeps signing with nobody opening the tab (s2), and an invalid switch is now refused at the write instead of echoed back raw (s9, all three clauses). Before/after screenshots and a live before/after of the API refusal are in the validation record.

**What remains false.** "Turning it off keeps the text" is a promise about the console and the MCP path. An API caller who sends a partial `settings` body loses the whole bag, this block included. That is #614, and until it is closed the honest statement of this feature's scope is the one in this paragraph.

PR #679, opened 2026-09-17
Compute the due set once, so a claim's limit actually holds

### LINKED ISSUE #627: scheduler-lanes OBSERVE claim tests failed once on CI with every due row run
Two tests added by #624 failed once on CI and passed on a rerun of the same commit, so they are timing- or order-dependent as written.

**Where:** `tests/modules/scheduler-lanes.test.ts`
- `scheduler lanes > the shared tick claims OBSERVE beyond the traffic share, and only up to its own limit` (line 415)
- `scheduler lanes > the OBSERVE claim gives up the rounds the fixed batch's provider work already took` (line 481): `expect(ran.OBSERVE).toBe(5)`, received `10`

**Measured:**
- PR #626 at `d2f85a2a` (rebased on `05a31185`, the #623 merge; its diff touches no scheduler file), run [34777420817](https://github.com/fazer-ai/agents/actions/runs/34777420817), job `test (2/4)`: 2 fail, 2892 pass. The rerun of the same job on the same commit: success.
- `bun test --shard=2/4` locally on the same commit (macOS): 2894 pass, 0 fail. `bun check` on the same commit: 11416 pass, 0 fail.
- `main` at `05a31185` passed CI. Adding test files reshuffles every shard (the workflow comment at `.github/workflows/test.yml:77` says so), so a PR that adds files can land these tests next to a different neighbour.

**What the numbers say:** `ran.OBSERVE` counts calls to a handler registered process-wide with `registerJobHandler`. Ten is every OBSERVE row the test enqueued, so either the claim limit was not applied on that tick, or something other than the test's own `runSchedulerTick` ran those handlers while they were registered, such as a scheduler worker or timer another test file left running in the same process. Not verified which; this issue records the failure, not the cause.

**Acceptance:** the two tests cannot count a handler call that did not come from the tick under test, and pass under `--shard=2/4` on CI across repeated runs.

gabrieljablonski:
Closes #627.

## Half the report holds, and the mechanism is not the one it guessed

The issue recorded two `scheduler lanes` tests failing on CI and passing on a rerun of the same commit, and it was careful to say it did not know why: "either the claim limit was not applied on that tick, or something other than the test's own `runSchedulerTick` ran those handlers". It is the first, and the defect is in `src/modules/scheduler/service.ts`, not in the tests.

## What was measured

`claimWhere` asked for its batch like this:

```sql
UPDATE scheduler_jobs SET status = 'CLAIMED', ...
WHERE id IN (
  SELECT id FROM scheduler_jobs WHERE status = 'PENDING' ...
  ORDER BY run_at FOR UPDATE SKIP LOCKED LIMIT $n
)
RETURNING ...
```

`WHERE id IN (SELECT ...)` is a semi-join, and a semi-join keeps the UPDATE's target on the **outer** side, so the subquery is the inner one. When the planner puts no `Materialize` above it, the inner side is re-executed once per outer row; each re-execution `SKIP`s the rows the previous one just locked and returns a **different** `$n`, so every row the outer scan reaches ends up matching. The plan, reproduced both on CI and locally:

```
Update on scheduler_jobs
  ->  Nested Loop Semi Join
        Join Filter: (scheduler_jobs.id = "ANY_subquery".id)
        ->  Seq Scan on scheduler_jobs
        ->  Subquery Scan on "ANY_subquery"
              ->  Limit
                    ->  LockRows
                          ->  Index Scan using scheduler_jobs_status_run_at_idx ...
```

Instrumented on CI, inside one transaction, on the same parameters (run 35152684463, job `test (4/4)`):

```
[P] ctrl  lim=5 -> 5     SELECT id ... ORDER BY run_at LIMIT 5
[P] claim lim=5 -> 10    the UPDATE above
[P] ctrl  lim=8 -> 8
[P] claim lim=8 -> 10
```

The parameter binds; the subquery's `LIMIT` is what does not hold.

That accounts for both failing assertions at once. With the cap gone, each test ran **every** row it had enqueued: ten `INGEST_MESSAGE` where the traffic share is five, ten `OBSERVE` where `observeClaimLimit(2, 3)` is five. `APPOINTMENT_REMINDER` read 3 and passed only because three were enqueued. No second tick was involved, which was the open hypothesis on the issue: `startScheduler` is called from `src/index.ts` alone, nothing in the suite imports it, and the probe's stack traces put all ten handler calls under the test's own `runSchedulerTick`.

## Why it is not a test's problem

That `LIMIT` is what every lane's budget is made of: the traffic share, the observe cap, and `batchSize` itself. A tick that claims every due row and drains it at once is the head-of-line blocking the lanes were built to prevent, and on a busy fleet it is also an unbounded batch of provider work behind one semaphore.

## Why it looked like a flake

Locally the planner picks `Hash Semi Join`, which evaluates the subquery once, so the same code is correct here and wrong there on the same Postgres 17. Four CI runs read as "timing- or order-dependent" because a plan choice is neither. Under the fleet role the RLS quals push the planner back to a safe plan, which is why the failure could not be reached through the claim's own path on this machine.

## The change

The statement is now a CTE joined into the UPDATE:

```sql
WITH due AS MATERIALIZED (
  SELECT id FROM scheduler_jobs WHERE ... ORDER BY run_at FOR UPDATE SKIP LOCKED LIMIT $n
)
UPDATE scheduler_jobs SET ... FROM due WHERE scheduler_jobs.id = due.id RETURNING ...
```

`MATERIALIZED` is the executor-level guarantee of one evaluation. A `FOR UPDATE` CTE is not inlined today, so the keyword is redundant today; it is written out because what this depends on is the single evaluation, and saying so is what stops a later Postgres, or a later edit that drops the `FOR UPDATE`, from quietly restoring the bug.

The SQL moved into an exported `claimSql` so the test can run the statement production runs rather than a retyped copy of it.

## The family: one site

The tree has three `SKIP LOCKED` claims. The other two, `src/modules/flowlog/alert-worker.ts` and `src/modules/webhooks/outbound/worker.ts`, already use `UPDATE ... FROM (SELECT ... LIMIT n) picked`. Measured side by side against ten due rows under the same forced plan:

| shape | rows returned for `LIMIT 5` |
| --- | --- |
| `WHERE id IN (SELECT ... FOR UPDATE SKIP LOCKED LIMIT 5)` | 10 |
| `FROM (SELECT ... FOR UPDATE OF ... SKIP LOCKED LIMIT 5) picked` | 5 |
| `WITH due AS MATERIALIZED (... LIMIT 5)` | 5 |

The difference is structural rather than luck: a semi-join must keep the UPDATE's target on the outer side, so the subquery is necessarily the rescanned inner one, while a plain join leaves the planner free to drive from the small side, which it does. Nothing else changes.

## The proof

`tests/modules/scheduler-claim-limit.test.ts` pins the hazardous plan onto its own connection (libpq `options`, because the statement runs inside a transaction the test does not open) instead of waiting for the planner to choose it, and runs `claimSql` itself.

Red, with the old SQL restored and nothing else changed:

```
(fail) the due set is evaluated once, as a CTE
       Received: "Update on scheduler_jobs\n  ->  Nested Loop Semi Join\n ..."
(fail) the traffic share claims its five of ten, not all ten   Expected: 5  Received: 10
(fail) the fixed batch claims its four of ten, not all ten     Expected: 4  Received: 10
 1 pass  3 fail
```

Green with the change: `4 pass  0 fail`. Full suite: `bun check` → `11596 pass  0 fail`.

The one test that passes in both directions is the rig's own guard, and it is there on purpose: an `options` string the driver silently dropped would leave the other three running on the safe plan and passing for the wrong reason.


gabrieljablonski:
## Correção da evidência de mutação desta PR

Duas das três sobreviventes que registrei na bateria desta rodada não eram sobreviventes: as âncoras casavam antes do alvo, então a mutação caiu em outro lugar e a suíte, corretamente, não reclamou de nada.

O script usava `replace(old, new, 1)` sobre `src/modules/scheduler/service.ts`, e:

| mutação registrada | onde a âncora casou de fato | leitura correta |
| --- | --- | --- |
| `FOR UPDATE SKIP LOCKED` → `FOR UPDATE` | linha 31, um **comentário** | a mutação real na query **sobrevive** |
| `claim_seq = claim_seq + 1` → `claim_seq = claim_seq` | linha 424, o `UPDATE` de **outra função** (a que retira uma arma por dedupeKey) | a mutação real na claim **MORRE** |

Medido de novo com âncoras asseridas únicas, contra `scheduler.test.ts`, `scheduler-claim-token.test.ts`, `scheduler-lanes.test.ts` e `scheduler-claim-limit.test.ts`:

```
m4 SKIP LOCKED fora da CLAIM      SOBREVIVEU
m7 claim nao incrementa claim_seq MORTA :: (fail) scheduler claim token > the claim issues a new
                                           token every time, and the re-arm does not
                                        :: (fail) scheduler claim token > a superseded run cannot
                                           complete the arm that replaced it
```

O `claim_seq` da claim, portanto, **está coberto**, e a frase que escrevi no `validation.json` (que `scheduler-claim-token.test.ts` monta o token à mão) está errada: aquele arquivo chama `claimDueJobs` duas vezes de verdade e afirma `current.claimSeq > stale.claimSeq`.

**O que isso muda e o que não muda.** Não muda o mérito do merge: a correção do `LIMIT` foi provada pela prova red (`5→10` e `4→10` com o SQL antigo restaurado), pela medição na CI dentro de uma transação e pelo controle positivo ao lado dela, e nenhuma dessas leituras depende da bateria. Muda o registro de cobertura: a bateria útil desta rodada foram 6 mutações, não 8, e a contagem `8/3` que ficou no `validation.json` está inflada.

A issue #681, que abri a partir dessa leitura errada, foi corrigida para o único vazio que sobrou.

PR #692, opened 2026-09-17
Vision analyzes every attachment, not only the first (#691)

### LINKED ISSUE #691: Vision analyzes only the first attachment, so a message with several files is answered by asking for what is in the ones it never opened
## What happens

A customer message with several attachments is analyzed **on the first one only**. The other files are downloaded by nobody, described to nobody, and the model never learns they exist — so it asks the customer for data that is sitting in the attachment it did not open.

Reproduced with three synthetic PNGs in one message, empty message body, each image carrying one field that exists nowhere else. The proof is running it twice with the same three files in a different order:

| attachment order | what the agent quoted | what the agent then asked for |
| --- | --- | --- |
| `a` (order no., purchase date, PIX), `b` (PIX id, amount), `c` (holder name, e-mail, phone) | order `40000001`, purchase date, "via Pix" — all of `a` | private note said `CPF/e-mail da compra: não informado`, with the e-mail sitting in `c` |
| `c`, `b`, `a` | greeted the customer by the holder name — from `c` | "Número do pedido; Nome e data do evento; Data da compra" — the fields of `a` |

Same three files, one position swap, and each run reads position 1 and nothing else.

The execution log agrees and is easy to misread: `stage: vision` writes **one record per turn**, `status: ok`, even for a message with 3 or 5 attachments. Nothing reports a skip, because from the pipeline's point of view nothing was skipped.

## Why it matters

On one production mailbox (SAC by e-mail, 30 days): **3,667 conversations arrive with a customer attachment, and 1,856 of them — 50.6% — carry more than one**. Mean 1.98 attachments, max 70. Half the attachment traffic hits this.

The failure mode is the one customers notice immediately: they send the receipt, the ID and the screenshot in one e-mail, and the reply asks them to send the receipt.

## Where it is

Three places each collapse to one, and any single fix leaves the other two:

1. `src/modules/chatwoot/webhook.ts:1922` — `const visual = firstVisualAttachment(n)`. One attachment is extracted, ever. `firstVisualAttachment` (`normalize.ts:899`) returns on the first match by design and by name.
2. `src/modules/chatwoot/messages.ts:168-169` — `metaStringFrom(item.attachments, "image_description")`. Its own comment says it: *"read from the first attachment that carries it"*. So even a message whose attachments all had meta written would surface one.
3. `src/modules/chatwoot/render.ts:155-160` — `else if`: `<imagem>` wins over `<documento>`, so a message with a photo **and** a PDF renders only the photo, even when both extractions exist.

The storage is not the problem: the extraction is already persisted **per attachment**, in each attachment's meta (`image_description` / `extracted_text`, `vision/service.ts:211`). The data model already holds N; three readers throw N-1 away.

## What a fix has to decide

- **A cap.** One vision call per attachment means a 70-attachment message is 70 calls. A per-message cap, with the overflow named to the model ("+N files not analyzed") rather than silently dropped.
- **A budget.** The per-file budget today is 20s image / 60s document. N of those in series is not acceptable on a turn; parallel with a total ceiling is.
- **A boundary in the rendered text.** The model has to see which datum came from which file, so the parts are labelled (file name) rather than concatenated into one blob.
- `firstAudioAttachment` has the same shape. Out of scope here unless someone wants it in.

## Notes

Measured 17/09/2026 against v1.17.3, agent with `vision.enabled`, `provider: openai`, `model` empty (so the `gpt-4o` default). Nothing about the extraction itself is broken: 14 real conversations extracted with `status ok` in 1.4-2.6s each, the operator's `extractionPrompt` obeyed (it described a CPF document and its visible photo without transcribing numbers, in 14/14 with zero document-number leaks). Single-attachment messages read correctly — two with an empty body had the order number read off the image.

gabrieljablonski:
## What was broken

A customer message with several attachments was analyzed **on the first one only**. The other files were downloaded by nobody and described to nobody, so the model answered as if they were not there — and asked the customer for data sitting in the attachment it never opened.

Nothing reported it. `stage: vision` wrote one line per turn with `status: ok`, because from the pipeline's point of view nothing had been skipped.

**Measured on a production mailbox (30 days):** 3,667 conversations arrive with a customer attachment and **1,856 of them, 50.6%, carry more than one**. Mean 1.98, max 70.

The proof is the order swap — the same three synthetic files, empty message body, each carrying one field that exists nowhere else:

| attachment order | what the agent quoted | what it then asked for |
| --- | --- | --- |
| order screenshot first | order no., purchase date, "via Pix" | private note said `CPF/e-mail da compra: não informado`, with the e-mail in the third file |
| holder's data first | greeted the customer by name | "Número do pedido; Nome e data do evento; Data da compra" — the fields of the file now third |

## The fix

Three places each collapsed N to one, and any single fix leaves the other two:

- **`webhook.ts`** took `firstVisualAttachment`, which did what its name said. Now every visual attachment, **in parallel**: the per-file budget is 20s image / 60s document, so five in series is a turn nobody waits for while five at once cost one. A cap of **8** per message bounds the album case, and the overflow is **named to the model** rather than dropped — a model told nothing answers as if the message had those files fewer, which is the bug.
- **`messages.ts`** read the meta of *"the first attachment that carries it"*, in its own comment's words. Now every one, labelled by file name, so two receipts are two orders and not one document contradicting itself.
- **`render.ts`** was an `else if`, so a message carrying a photo **and** a PDF rendered only the photo and the document vanished without a trace.

The storage was never the problem: vision already writes back per attachment. Three readers threw N-1 away.

**One attachment produces byte-identical output to before**, label included — there is nothing to tell apart. That is why the 93 existing tests on these three files pass untouched.

## Validation scope

**Proved by test** (`tests/modules/vision-every-attachment.test.ts`, plus cases in `chatwoot-messages` and `chatwoot-render`):

- the ratio of `vision` lines to attachments, through the real receiver (`processChatwootDelivery`) — a unit test on `visualAttachments` would pass with the call site still taking `[0]`. Measured as a ratio, never an absolute count, because the eager pass has more than one call site per delivery;
- the cap at 8 with its named overflow, and that the notice is written once rather than once per pass;
- the labelled join across three attachments, and the positional fallback when a file has no usable name;
- the single-attachment shape, which must not change;
- a message carrying both an image and a document.

**Five of the six were seen red** against the previous behavior.

**Proved live**: the defect itself, on a real tenant, by the order-swap runs above, and the 50.6% figure by SQL on the production mailbox. The **fix** has not been exercised against a live provider — the test path skips at `no_credential` by design, which is what makes it deterministic and offline.

**Not covered**: `firstAudioAttachment` has the same shape and is untouched here. Several audio attachments in one message still transcribe one; it is a separate issue and far rarer in the traffic measured (16 mp4 + 14 quicktime against 4,068 jpeg in the same 30 days).

Fixes #691


gabrieljablonski:
## What the merge gate refused, and what stays unproven after this merge

The mechanical gate (`ship.py dod`) is **red on two items**, and both are recorded here rather than worked around.

| check | state |
| --- | --- |
| head_sha, tests, red_proof, mutation, review, skip_codes, base | ok |
| **live_e2e** | **not exercised** |
| **verifier** | **absent, and unclosable now** |

**`live_e2e`.** Nothing in this PR has run against a real Chatwoot inbox with real attachments. Doing so requires the fix to be deployed there, which is a release plus a deployment of a production instance: not this round's to perform. There is no skip code that describes this honestly (`PROD_ONLY_DATA` would be false — the state is not production-only, the *deployment* is), so it is written out instead of coded. The live exercise belongs to the release, and until it happens the behaviour of this change on production traffic is proven by tests only.

**`verifier`.** The holdout scenarios have to be sealed *before* the code exists, and this round did not start as a ship round, so there was never a seal. Writing them now would be writing the answer key after the exam, so they are not written.

**What IS proven**, so the two gaps are read against something:

- 8 review rounds, 12 findings, each fix red-first. The last round is clean on the merged HEAD.
- A fresh mutation battery on the final code: **7 mutations, 7 killed, 0 survivors**. Each one switches off a single behaviour this PR introduces (the cap cutting before the cache, the count never returning to zero, the cached extraction being paid for again, the partial page meta winning over the complete aggregate, reading only the first visual attachment on either side, and the unread marker) and at least one test goes red for each.
- Full suite green locally (11667 tests) and CI green on lint, type-check and all four shards.

**Also left out on purpose:** `firstAudioAttachment` has the same first-only shape on the STT path. It is a different budget per file and a different call site, so it gets its own issue rather than riding along here.

Issue #642, opened 2026-09-14
A watching agent cannot see what it already decided, so its classification churns


gabrieljablonski:
## What happens

A watching agent cannot see what it already decided on the conversation it is watching, so it re-decides from scratch on every burst and the classification churns.

Measured on one install, 12 hours, 837 conversations observed:

- 676 category changes across 285 conversations.
- **161 conversations (22%) went A → B → A**, one of them alternating `outros ↔ dúvidas-evento` eight times.
- **42% of the conversations that changed category ended in the category they started in**; among the ones that oscillated, 69% did.
- Median gap between two classifications: **2 minutes** (minimum: 1 second).

The pairs are symmetric, which is the signature of indecision rather than of a subject that evolved: `cancelamento → compra-de-ingresso` 82 times against 63 the other way, `dúvidas-evento → compra-de-ingresso` 77 against 60.

Each change also posts an activity line in the conversation, so a human reading the thread sees the label flip back and forth.

## Why the tick cannot tell

The tick is stateless by design: it rebuilds everything from Chatwoot, which is what makes an observation reproducible from the conversation alone (`docs/chatwoot.md`). The frame already carries that principle — *"o que você já fez nesta conversa está no que está registrado nela, não na sua memória"* — and it already gives the model the labels standing right now (`<current_labels>` on the tool, `labelsBefore` on the line).

What it does NOT give is the RECORD OF THE CHANGES. Chatwoot writes one activity message per label change ("Classificador SAC adicionou cancelamento", "Classificador SAC removeu compra-de-ingresso"), and `usableRow` (`observe/job.ts`) drops every activity row from the window on purpose, because it is "the system narrating, not either side talking". So the model cannot distinguish "this label has been here since the first message" from "I have already changed this twice in the last four minutes", and its instruction to leave things alone when nothing needs to change has nothing to apply to.

## Expected

The observation frame carries the conversation's own label history from the window it is already reading, so the model sees what it (or a human) already decided before deciding again.

## Shape

- Built from the rows the window ALREADY fetched (`readWindowRows` returns activity rows; only the transcript filter drops them), so no extra request.
- **Verbatim, never parsed.** The activity line is a localized sentence with no sender and no `content_attributes` (measured: `message_type=2`, `content_type=0`, `sender_id` null), so any structure we would extract has to come from grammar in the account's language. Forwarding the line keeps the author, the verb and the label intact in every language.
[...]

PR #699, opened 2026-09-17
/reset names the labels it removed, so its own removal line is not read as history

### LINKED ISSUE #645: O /reset não consegue esconder a própria remoção de etiquetas do observador
Descoberto revisando a #644 (rodadas 20, 21 e 22), que tentou três correções e foi derrubada por um fato mais fundo a cada vez.

## O que acontece

`/reset` limpa as etiquetas da conversa. O Chatwoot narra essa remoção como uma linha de atividade ("Fulano removeu compra-de-ingresso"). O bloco `<mudancas-de-etiqueta>` do observador (#642) lê linhas de atividade, então o próximo tick pode ver as etiquetas do episódio que o reset acabou de apagar — e um observador sem memória lê isso como motivo para não recolocá-las, que é o oposto do que o operador foi informado.

## Por que as três tentativas não fecham

1. **Filtrar pelo `reset_at_message_id`** (o que já protege a transcrição e as notas): não basta. O boundary é o id da **mensagem do comando**, e a limpeza acontece uma dúzia de chamadas depois, então a linha nasce acima dele.
2. **Cortar na primeira linha que não é narração depois do boundary**: uma mensagem do cliente chegando no meio da limpeza toma esse lugar. E o corte nunca parava de valer: com a janela já deslizada além do reset, seguia descartando atividade legítima.
3. **Cortar no ack do próprio comando, marcado por `content_attributes`** (o que está na #644 hoje): melhor, mas não é garantia. `LabelActivityMessageHandler#create_label_change_activity` usa `Conversations::ActivityMessageJob.perform_later`, ou seja, a linha é criada por um job do Sidekiq. Com fila atrasada — e esta instalação já teve isso — o job roda depois do ack e a linha fica com id acima do corte.

## O que falta decidir

Nada na linha diz quem a causou: uma mudança de etiqueta não grava `content_attributes` nenhum (diferente de uma mudança de status, que grava `conversation_status_changed`). Então nem ordem nem conteúdo separam a remoção do reset de uma remoção legítima.

Caminhos possíveis, nenhum barato:

- **No fork**: fazer o `ActivityMessageJob` gravar um `activity.type` (ou qualquer marca) quando a mudança vem de uma limpeza. Resolve na origem e serve a qualquer leitor, não só a este.
- **Deste lado**: o `/reset` registrar quais etiquetas removeu, e o observador descartar linhas de remoção que nomeiem só aquele conjunto enquanto a conversa não tiver recebido etiqueta nova. Uma remoção legítima da mesma etiqueta também cairia, o que é uma perda, não uma leitura errada.
- **Aceitar**: a frase é verdadeira e o alcance é conversa em modo de teste. É o que a #644 faz hoje, com o resíduo escrito no corpo da PR e em `docs/chatwoot.md`.

## Alcance

`reset_at_message_id` só é preenchido em conversa onde alguém digitou `/reset`, o que exige modo de teste. Conversa de cliente real nunca passa por isso.

gabrieljablonski:
Fixes #645

## What the observer was reading

`/reset` clears the conversation's labels; Chatwoot narrates that clear as an activity row ("John removeu compra-de-ingresso, orcamento-enviado"), and the observer's `<mudancas-de-etiqueta>` block reads activity rows. So the next tick could read the labels the reset had just erased, **named**, and a stateless observer reads that as a reason not to put them back — the opposite of what the operator was told happened.

The cut that was supposed to hide it (#644, round 20) is an **order** test: drop every activity row up to the acknowledgement's own row, found by the name the command stamps into `content_attributes`. The order is not ours to choose. `LabelActivityMessageHandler#create_label_change_activity` hands the row to `Conversations::ActivityMessageJob.perform_later`, so it appears whenever that Sidekiq job runs; on an install whose queues are backed up the job runs *after* the acknowledgement and the row keeps an id **above** the cut.

Measured against a real Chatwoot (local fork, account 1, conversation 20719), posting the acknowledgement before the clear so the row lands where a delayed job would put it:

```
66052 incoming  "/reset"
66053 outgoing  "🔄 Memória, …"   content_attributes {fazer_ai_send_id: "reset-ack:66052", fazer_ai_reset_cleared: [...]}
66054 activity  "John removeu compra-de-ingresso, orcamento-enviado"   content_attributes {}
66055 activity  "John adicionou cancelamento"                          content_attributes {}
```

The old order cut, on those rows: `["John removeu compra-de-ingresso, orcamento-enviado", "John adicionou cancelamento"]`. Both reach the model.

## What this changes

The command now **records the set the clear removed**, on the conversation's own row (`conversations.reset_cleared_labels`, a new nullable `jsonb` column), read inside the conversation's label queue immediately before the write that empties it and written only after that write returns. The read is best-effort: it is bookkeeping for a later reader, while clearing the labels is what the operator asked for, so a `GET` that fails leaves no claim and the clear goes out anyway. The observer's question stops being *where did the cleanup end* and becomes *is this line about what the cleanup took off* — which has the same answer at any id. Same rows as above, after this change: `["John adicionou cancelamento"]`.

The set rode the acknowledgement's own `content_attributes` until review round 2 measured where that bag goes: `api/v1/widget/messages/index.json.jbuilder` renders `json.content_attributes message.content_attributes` and `Message#push_event_data` returns `attributes.symbolize_keys.merge(...)`, so on a website inbox the bag of a **public** message is handed to the contact. Internal label names are not the customer's. The column is read together with the boundary, from the same row in the same query, so a second `/reset` between two reads cannot pair one reset's boundary with another's set, and it is written only by the reset that still owns that boundary. A clear that threw records nothing (`NULL` is "no claim", `[]` is "it removed nothing"), so the labels still standing keep their lines — and the set is invalidated in the same statement that moves the boundary forward, since a second `/reset` whose cleanup fails would otherwise leave the new boundary beside the previous reset's set.

Only a **removal** line can be the cleanup's own, and the verb is read rather than assumed: `conversations.activity.labels` has exactly two leaves, so the 74 vendored templates are now two tables of 37 (`bun scripts/extract-chatwoot-activity-templates.ts` prints them apart; a test asserts the union is the whole table and that the two do not overlap). A title is **consumed** by the removal line that narrates it, so a label put back after the reset and taken off again keeps both of its lines: the reset removes each title exactly once. The reading is checked against the cleared set rather than the account's catalog, which is stricter for free and settles the pt/pt_BR ambiguity (`"Ana removeu a vip"` is the label `a vip` under one locale and `vip` under the other) with titles that were standing on this conversation seconds ago.

`[added A, removed A]` is **ambiguous by construction** — the rows are identical whether the removal is the reset's, delayed behind an addition that was delayed too, or somebody's real change after the cleanup's own line aged out of the fetched page. So the page's **reach** decides: with the command still in the page, everything Chatwoot wrote since the reset is in it too, the removal is the reset's and an addition never releases a title; with the page truncated above the command, an addition releases the title and the removal is read as real history. Holding a budget for a row that will never be fetched would hide real removals forever; spending it on an addition would leak the cleanup. Each regime takes the error that fits it.

The acknowledgement's own row still cuts everything the command wrote before it, **joined** with the set rather than replaced by it: only that cut covers a *previous* reset's removal line arriving late, which names a title the newer clear no longer found standing and therefore never recorded. It costs what it always cost, and the cost is back with it: a colleague who changed a label between the command and its acknowledgement loses that line.

No fork change: the mark the issue asked the fork to write on the activity row is on a row of ours instead.

## What in the issue does not hold

- **"closing it for certain needs a mark only the fork can write"** — it needs a mark on a row *we* write, not on the row Chatwoot writes. `conversations` is ours, and it already carries the reset's boundary.
- **"Uma remoção legítima da mesma etiqueta também cairia, o que é uma perda"** (the issue's own sketch of this approach, released by "enquanto a conversa não tiver recebido etiqueta nova") — it does not, because the release is **per title** rather than global: the set stops answering for a title the moment its own line is read.
- **"Conversa de cliente real nunca passa por isso"** (the Alcance section) — the gate is `commandActive = command !== null && commandMode === "test"` (`webhook.ts`), where `commandMode` is the **agent's** mode, plus `/teste` activated on that conversation (`shouldRunReset`). The command rides an **incoming** message (`isNewIncomingMessage`), and nothing asks who the contact is: a real customer's conversation reaches this the moment an operator activates test mode on it, which is how an agent gets evaluated. The scope claim is about who typed the command, and the gate does not check that.
- **"Aceitar"** was the third option, and the reason to reject it is not the frequency: test mode is the operator's *evaluation* surface, so a wrong label there is read as the agent's judgement, which is exactly the thing being evaluated.

## Validation scope

**Proved by test** (`bun check`: 11880 pass, 0 fail):
- `tests/modules/observe-transcript.test.ts` — the cleanup's line above the ack is hidden; a title put back and removed again keeps both lines (on a truncated page and on a complete one, where consumption is the only thing that releases it); the same pair `[added A, removed A]` read both ways by how far the page reaches; the filter applies the reset boundary itself; a line naming a cleared title *next to a live one* survives; neither a private note nor a row with its own `activity.type` spends a title; a line too long to scan goes unread on both sides and is counted as an omission (`omitted: 1`); a clear that removed nothing hides nothing; an acknowledgement with no set falls back to the order cut; the two template tables partition the old one, and the verb reaches the reader in a locale that puts it at each end.
- `tests/modules/chatwoot-reset.test.ts` — the reset records the set the conversation had, paired with the boundary of the command that removed them; the labels are read before the POST that empties them (`GET` then `POST`); the acknowledgement's `content_attributes` carries exactly one key, the send id, and nothing about the account's state; a clear that failed, **after a healthy reset**, records `NULL` instead of inheriting the previous set; a snapshot read that failed does not stop the clear, and leaves SQL `NULL` (asked as SQL, because a written json `null` answers FALSE to `IS NULL`); a reset that no longer owns the boundary does not overwrite the newer one's set.
- The column value is read rather than trusted: an array of strings or `null`, with a single element of another type disqualifying the whole value.
- **Mutation battery**: 19 mutations, 0 survivors (consumption removed, `every`→`some`, the null fallback dropped, each of the three guards removed, the bag omitted when only labels are given, `every`→`some` in the reader, the read moved after the clear, the assignment moved before the write, `[]`→`undefined` on a failed clear, the verb ignored when consuming, an addition never releasing and always releasing, the page's reach inverted, the boundary filter dropped, the two verb tables swapped at compile time, the boundary fence dropped from the column write, the column value accepted half-right, the set put back on the public message, the fence that invalidates the set when the boundary advances, the snapshot read back to aborting the step, the column written without having been read, and the two cuts unjoined).

**Proved live** (local Chatwoot fork, Puma at :3000 on `chatwoot_main_dev`, Sidekiq running, account 1, inbox 1884, AgentBot 70):
- Chatwoot persists an **array** inside `content_attributes` verbatim on create and echoes it back on `GET /conversations/:id/messages`. That is what the first design relied on, and it is also why it had to move: the same bag is rendered to the contact by the widget endpoint (source read on `chatwoot-pro-main`, above).
- A clear of two labels renders **one** activity row with the titles joined by `", "`, and the row carries `content_attributes: {}` — no sender, no type, which is why no structural test separates it from a human's removal.
- The real rows, through the real parser and the real filter: the window past the command is `[66053, 66054, 66055]`, the filter keeps `[66053, 66055]`, and the observer's block reads `{lines: ["John adicionou cancelamento"], omitted: 0}` where the order cut read both lines. Measured while the set still rode the acknowledgement; the filter's input is the same list either way, and the column is covered by the DB-backed reset tests.
- Residue removed afterwards (`DELETE /contacts/9440`; conversation and messages gone).

**Not validated**:
- The case both regimes leave open: a **truncated** page whose re-add line has also aged out still hides the first removal it sees of a cleared title. Detecting it would need the cleanup's own row, which is exactly what is missing. Not reproduced.
- The **complete**-page regime chooses to read `[added A, removed A]` as the reset's delayed removal, so a re-add and removal that both happen while the cleanup's row is still unwritten lose the removal line. It needs the label activity job to be behind the whole cleanup. Not reproduced.
- A clear whose POST applied while its response was lost: the acknowledgement then declares `[]` and the removal line shows. Not reproduced (it needs a partition mid-POST), and it fails in the direction this block fails everywhere else — a true sentence showing.
- The column write itself was not exercised against the live Chatwoot: the live run predates review round 2, which moved the set out of the message. What it proves is unchanged (the row Chatwoot writes, its text, its position and how our filter reads it); what it no longer covers is where the set is stored, which the DB-backed reset tests do.


gabrieljablonski:
## What the acceptance gate does not cover here

Merged with the mechanical gate reporting `FAIL` on one check, and the reason is the round's process rather than the change. Recording it so the next reader does not have to reconstruct it.

**The holdout scenarios do not predate the diff.** The acceptance scenarios for this issue were written, and sealed on the issue, *after* the implementation was measured and this PR was opened: the seal is stamped `2026-09-17T22:35:35Z` and the branch's first commit landed `2026-09-17T22:13:06Z`. Worse than the timestamp, the agent that wrote them was pointed at `docs/chatwoot.md` in the working tree, which already described the chosen design, so the set is not blind to the fix (it reported that on its own). The ten scenarios are good and their execution is real, but the claim a holdout exists to make — that the scenarios were written before the change — is false for this round, and there is no way to make it true after the fact.

**Two of the ten did not come back PASS, and neither is a defect in this change.**

- `s7` fails one clause: after a *second* `/reset`, it expects the block to show the addition of a label the operator applied *before* that reset. It does not, and the cause is older than this PR — `reset_at_message_id` excludes everything below the boundary, which is the whole point of "the erased episode is not part of this one" (#642, migration of 30 Aug). Confirmed with a control on the same conversation: the same label applied *above* the boundary shows normally. The clause asks for behaviour the product deliberately does not have.
- `s3` is `BLOCKED`: reproducing the delayed-activity arrangement needs Chatwoot's Sidekiq stopped, and it is shared with other work on this machine. Measured across the four resets it did run: the removal row is written 87-101 ms *before* the acknowledgement, so inverting that order would mean holding `ActivityMessageJob` for more than 100 ms. The case it could not build is the one measured by hand and reported in the PR body (removal `66054` above acknowledgement `66053`, the block reading only the genuine addition, and both lines reading through when the set is absent).

Everything else the gate asks for is on record: red proof 1→0, 20 mutations with no survivors, `bun check` green, seven review rounds with the last one clean at this head, CI green on all four shards, and the base combination proved in a throwaway worktree against `252b6eb`.

PR #710, opened 2026-09-18
set_labels names the delta, so a label nobody named is never touched

### LINKED ISSUE #695: set_labels deletes by omission, so protecting a label means hiding it from the model
## What is wrong

`set_labels` takes the **complete final list** of labels and diffs it against what the model was shown: shown-and-omitted is a removal, asked-and-not-shown is an addition. Removal therefore happens **by omission**, which makes two properties inseparable:

**Protecting a label and hiding it are the same lever.** `settings.setLabels.protected` works precisely because a protected label is never shown, and what is not shown can never be omitted, so it is never removed. There is no way to express "the model may see this label but not touch it".

That coupling has already cost us a measured defect. On 2026-09-11, on a live fork, an agent whose taxonomy was fenced with `protected` could not see the canonical values the other classifier owned, and **invented** `duvidas-evento` — a label that then lives in the account's catalogue forever. The workaround we settled on was not to fix the fence but to remove the tool from that agent's allowlist entirely.

**And a model that simply forgets a label deletes it.** Silently, with no way to detect it afterwards, because the deletion is indistinguishable from an intentional removal. The only guard is a hand-maintained allowlist, which on one production tenant has grown 23 → 28 → 35 → 36 entries, each time because somebody noticed a label had been left out. There is no mechanism that notices.

## Proposed change

Replace the single `labels` argument with `add` and `remove`:

```
set_labels(add: string[], remove: string[], scope: "conversation" | "contact")
```

- Nothing is removed unless named. **Protection becomes the default**, not a configuration.
- The model can be shown every label on the conversation, including ones it does not own, because seeing is no longer the same as being able to remove. That is what closes the invention defect above.
- Swapping a value stays expressible and becomes explicit: `remove: ["dúvidas-evento"], add: ["cancelamento"]` instead of an omission the reader has to infer.
- The failure mode inverts from destructive to recoverable: forgetting to remove leaves a stale label, which is visible and fixable; forgetting to include one currently **erases data**.
- Each call carries the delta rather than the whole list, so it is also cheaper.

`settings.setLabels.protected` survives with a better meaning: from "hide from the model" to "refuse to remove, whoever asks". It can then be shown to the model, and it stops being the only thing standing between a forgotten name and a deleted label.

## What this breaks, and what has to move with it

- **The tool contract changes for every agent in the fleet.**
- **Operator prose describes the current contract in five free-text fields** (`system_prompt`, `toolGuidance.<tool>`, `handoff.instructions`, `kanban.instructions`, `followUps[].instructions`). #604 established that a migration rewrites the tool name only in `system_prompt`, so the prose migration here has to be deliberate, or agents ship with instructions contradicting their own tool.
- **The observer's verdict path** builds the final list today; it has to name what it removes. It already performs the swap correctly under the current contract, so this is about stating an intent that used to be implicit, not about new behaviour.
- The label write queue (`withConversationLabels`, #477) is unaffected: it serializes writers, which is orthogonal to what each write means.

## Why now rather than later

Every new agent inherits this contract. The cost of changing it grows with the fleet, and the failure it produces — a label deleted by omission — is the kind that leaves no trace to debug.

gabrieljablonski:
Closes #695.

## What was wrong

`set_labels` replaced the scope's whole label list. Removal therefore happened by **omission**, and the only defence against erasing a label another writer owns was `settings.setLabels.protected` — a hand-kept list, capped at 50 and truncated in silence, which **hid** the protected label from the model as the price of protecting it. Hiding and protecting were the same mechanism, because being shown a label was the first half of being able to delete it by leaving it out.

Two incidents on a production tenant:

- **2026-09-11** — the responder, unable to see the values the classifier writes, invented `duvidas-evento` outside the catalogue.
- **2026-09-17** — the e-mail agent went live on an inbox where four other systems own labels. `itaqua-dados-pedidos` (401 uses) survived only because somebody remembered to add it to the list; the observer, whose own list had been built for another inbox, did remove `aguardando-resposta`, `email-voltou` and `itaqua-rodeio-fest`.

## What changed

The model names `add` and `remove`. The three rules are now properties rather than mechanisms:

| | |
|---|---|
| not named | not touched, whoever put it there and whenever |
| named in `remove` | removed, even if the model was never shown it |
| named in `add` | added; naming one already there is a no-op, not an error |

A label a person, an automation, the observer or n8n added between the turn's read and the call is safe for free, and so is one past the 40-label display ceiling.

`protected` keeps meaning "this tool may neither add nor remove it", **both halves**, because a tenant relies on the add half to keep one agent's taxonomy out of another's reach. What it stops meaning is "never sees": the label is printed in the tool description, in `<current_labels>` and in the observer's prompt, and **a refusal is reported by name**. The model sees it, will therefore ask, and answering "already as requested" would be a false statement it reads back out of its own transcript one call later.

There is a **fourth** model-facing site, and it stays filtered on purpose: the label-change history block that narrates who added or removed what (`labelHistoryFromRows(..., cfg.protectedLabels, ...)`). Seeing the label is what the model needs to reason and to be refused by name; reading that a human moved it is an invitation to undo the human, and no part of this issue asks for that.

The retired shape is **refused by name**, not stripped. A strict schema would drop the key, leave an empty delta, answer "already as requested", and let the model record a classification that was never written.

## Operator prose this PR does NOT migrate

Five free-text fields still describe the replace contract on every tenant that has not rewritten them, and each is operator-owned, so this PR does not touch any of them:

- `agents.system_prompt`
- `settings.toolGuidance.set_labels`
- `settings.handoff.instructions`
- `settings.kanban.instructions`
- `settings.followUp.steps[*].instructions`

**Residual risk, stated plainly:** a model obeying stale prose can put the whole current list into `add`. That call is accepted — every name in it is a legitimate add — and the effect is that an operator's removal is silently undone, because `remove` was never named. The by-name refusal only catches the literal `{labels: [...]}` shape, not a correct-shaped call carrying a replace-contract intent. Migrating those five fields is fleet-wide work and belongs to its own change.

**How an operator finds out: they do not.** There is no console warning, no import warning, no log line and no configuration-health check that reads those five fields and reports that their text still describes the retired contract. That is the cost of not migrating them here, and it is deliberate rather than overlooked: a check that reads free-text prose and judges it is its own piece of work, and getting it wrong means nagging every tenant whose wording happens to resemble the old one.

What the PR **does** fix is the prose the product itself ships into that box: `editor.labelInstructionsPlaceholder`, the example the console suggests inside `settings.toolGuidance.set_labels`, told the operator to "leave the others out" and to "keep `vip` and `agente-off` wherever they already are". Under the delta the first half removes nothing and the second is an active instruction to put back what an operator has just taken off. Rewritten in both locales, and in the `t()` default beside them.

## Validation scope

**Proved by test.** The 31 that asserted the replace contract were rewritten against the delta rather than deleted, and each keeps the harm it was written for: the concurrent write mid-turn, the guarded label, the two calls of one batch, the fence withdrawn inside the queue, the display ceiling, and the observer's line counts. New ones cover the retired `labels` list refused by name, an empty delta refused, a guarded label shown in all three model-facing places and still surviving the write, and the task scope reading the card fresh so a label added mid-turn is not overwritten. Full suite 11,760 pass / 4 skip / 0 fail, `biome check` and `tsc --noEmit` clean, all by exit code.

**Baseline redone.** The one recorded for this round was wrong: the clean-HEAD worktree had no test database, so all 698 files aborted in `tests/setup.ts` and nothing had run. With the database created, the clean HEAD passes 223/223 across the four files this PR touches, so the 31 failures were the contract change and nothing inherited.

**Proved live**, against a local Chatwoot (account 5, inbox 8, a monitoring agent on `gpt-5.6-luna` with `set_labels` and nothing else), reading the Chatwoot request log and the flow log:

- a conversation on `compra-de-ingresso` whose verdict turns to `cancelamento` produces **one** `POST /labels` whose body carries the new category and not the old one. It is never both and never neither, not even for an instant; the flow log records `added 1, removed 1, toolCalls 1`.
- a second tick on the same conversation, verdict unchanged, produces **zero** writes, and the flow log reports neither an add nor a removal.
- with `agente-off` standing on the conversation and **not** in `protected`, a verdict that does not name it leaves it in place: 3 labels before, 3 after, one swapped.

**Not proved live: the `task` scope.** The rehearsal fork does not carry the kanban routes (`GET .../kanban/tasks` answers 404 `No route matches`), so every assertion about the card is deterministic, against a test client. The one thing that a deterministic test cannot settle is the SHAPE of `GET /kanban/tasks/:id`, and it matters: the new code reads `.labels` off that response, so an envelope would make it `undefined` and an `add` would write the card with only the added label, zeroing the rest. Settled by reading the source of the fork that serves the route rather than by memory — chatwoot-pro `main` at `23ad9c6cd3`, where `tasks/show.json.jbuilder` is `json.partial! 'task', task: @task` (bare object, no envelope) and `_task.json.jbuilder:21` renders `json.labels task.cached_label_list_array` (plain array of strings). Written into the client beside the call. A live call against a kanban-carrying fork is still the stronger evidence and is not in this PR.

**A shape this PR does not change, and makes routine.** The guard applies to each half of a call on its own, so a swap with one half guarded lands the other half: with `cancelamento` guarded and `compra-de-ingresso` not, `add: [cancelamento], remove: [compra-de-ingresso]` leaves the conversation with **no category**, and the mirror configuration leaves it with **both**. A `protected` holding the whole exclusive set refuses both halves and writes nothing, which is the configuration that is actually correct. Measured identical on `5ae9af39`, so it is not introduced here — but a guarded label used to be hidden from every place the model acts, so the model did not ask for it; now it is shown on purpose, which is what makes this path common. Pinned by three tests so it cannot drift silently, declared in `docs/chatwoot.md`, and the product decision (should any refusal make the whole call atomic?) is #712.

**The matched pair, same conversation shape and same model.** Eight labels on the conversation, six of them owned by other systems, the verdict flipping one category into another:

| | tool call the model made | what the six unrelated labels depended on |
|---|---|---|
| base `5ae9af39` | `labels: [8 items]` | the model repeating all six |
| this PR `9e0842d5` | `add: [1], remove: [1]` | nothing the model does |

The resulting write is identical in both. **The defect did not reproduce live**: on two runs the base repeated the full list correctly, so what the rehearsal shows is not the base failing, it is the base's correctness resting on the model's memory while this PR takes the six labels out of the model's reach entirely. The failure itself is the one the red proof pins as an outcome, and the field evidence for it is the two incidents above.


gabrieljablonski (commented):
### 🔍 Codex Review

No actionable regressions were found. Type checking and all 150 focused tests passed; database-backed tests could not be verified because the test database was unavailable.

No actionable findings in this pass.

<sub>**Reviewed commit:** `f6bee837f2` · `gpt-6-astra` · effort `high`</sub>

PR #751, opened 2026-09-20
The follow-up fence dates the silence by whoever spoke last, not only by the customer

### LINKED ISSUE #750: A cerca do follow-up data o silêncio pela fala do cliente, e ignora a conversa que o agente acabou de responder
A cerca de ativação do follow-up mede o evento errado, e por isso não cobre conversa que o agente acabou de responder.

## A cerca hoje

`src/modules/followups/handlers.ts:186`, na seleção da varredura:

```sql
AND c.last_inbound_at IS NOT NULL
AND a.follow_up_armed_at IS NOT NULL
AND c.last_inbound_at >= a.follow_up_armed_at
```

O comentário em cima dela diz o propósito, e ele está certo: *"only episodes of silence that BEGAN after follow-up was armed... Without it, flipping an agent to production with follow-up on would blast every eligible conversation in the historical backlog at once."*

O problema é o proxy. A cerca usa **a última fala do cliente** para datar o episódio de silêncio. Mas o silêncio que um follow-up cobra é o que **nós** abrimos: o agente respondeu, pediu um dado, e o cliente não voltou. Esse episódio começa na nossa resposta, não na mensagem dele.

## Onde isso quebra

Conversa antiga que o agente atende hoje, seja por `conversation_reengage`, seja porque a conversa voltou do time, seja porque o cliente sumiu e alguém religou.

Medido em 20/09/2026, numa caixa de e-mail em produção, 19 conversas de 9 a 11 dias religadas pelo `reengage`: depois do turno, `last_replied_message_id` está preenchido nas 19 (então `isFollowUpLive` passa e `ourSideHasSpoken` está satisfeito), 13 ficaram em `pending` com o agente pedindo documento, e **nenhuma das 19 entra na varredura**. Duas razões, e cada uma bastaria:

- `last_inbound_at` é **NULL**: a linha do mirror nasceu de um evento que não é mensagem, então o campo nunca foi preenchido, e o `IS NOT NULL` já corta.
- Mesmo preenchido, o valor seria a data velha da mensagem do cliente, anterior ao `follow_up_armed_at`, e a cerca cortaria de novo.

Resultado: 13 pessoas receberam um pedido de documento que ninguém vai cobrar e uma conversa que nunca encerra. O agente fez a parte dele e a escada não existe.

## Proposta

Datar o episódio pela **nossa última fala**, não pela do cliente:

```sql
AND a.follow_up_armed_at IS NOT NULL
AND c.last_replied_at >= a.follow_up_armed_at
```

Isso preserva o propósito original inteiro: ligar o follow-up não dispara o backlog, porque no backlog **nós não falamos** depois de armar. E passa a cobrir o caso que hoje escapa, que é justamente aquele em que o agente trabalhou.

O watermark logo acima (`last_follow_up_at IS NULL OR last_inbound_at > last_follow_up_at`) tem o mesmo defeito pelo mesmo motivo, e precisa do mesmo eixo: "já cobrei este episódio?" se responde comparando com o início do episódio.

Falta a coluna: hoje existem `last_replied_message_id` (id, não instante), `chatwoot_first_reply_at` (a PRIMEIRA, não a última) e `last_event_at` (qualquer evento, inclusive do cliente). Nenhuma responde "quando nós falamos pela última vez". A mudança é um `last_replied_at` gravado no mesmo ponto em que `last_replied_message_id` já é.

## Alternativa que não resolve

Preencher `last_inbound_at` no religamento. O valor correto é a data real da mensagem do cliente, que é antiga, e a cerca continuaria cortando. Escrever ali a data de hoje seria mentir sobre quando o cliente falou, e outros leitores dependem desse campo.

gabrieljablonski:
## What was broken

The follow-up activation fence dated the silence by the **customer's** last message:

```sql
AND c.last_inbound_at >= a.follow_up_armed_at
```

`last_inbound_at` answers "when did the customer last write", which is the wrong event for a conversation **we** have just answered. Two populations fall through the fence, and neither is exotic:

1. **The customer spoke before follow-up was armed.** The agent replied afterwards, the conversation is genuinely waiting on the customer, and the ladder skips it because the only date it looks at predates the arming.
2. **The row has no inbound timestamp at all.** A mirror row created by a webhook that is not a message (`mirror.ts` creates the conversation from any event) is born with `last_inbound_at` and `last_replied_message_id` NULL. Observed on a live deployment while recovering conversations older than the mirror: every row created that way had both columns NULL, and the sweep never enqueued a single one of them — no 24h nudge, no 48h close, silently.

The same axis is read in three places (the sweep SQL, the step-0 re-check in the handler, and the countdown estimate in `conversations/service.ts`), so a conversation could also pass the first and be dropped by the second.

## The fix

`conversations.last_replied_at` records **when** our side spoke. It is stamped in the debounce claim, beside the existing `last_replied_message_id`, and unconditionally where that id write is conditional: the id must not go backwards, but the timestamp of "we just spoke" is true either way.

`silenceStartedAt(lastInboundAt, lastRepliedAt)` returns the more recent of the two and is the single expression all three readers use. In SQL that is `GREATEST(...)`, **not** `COALESCE`: coalescing would let a stale reply fence out a conversation whose customer spoke after arming, which is a regression in the opposite direction. Postgres `GREATEST` ignores NULLs, so the fallback to `last_inbound_at` is exactly what every row written before this column had.

The ladder cannot feed itself on the new column: the nudge does not call `claimReplyBurst`, so only a real reply to the customer moves `last_replied_at`, and `isNewFollowUpEpisode` still ends the episode when the **customer** speaks.

## Validation scope

**Proved by test** (`bun run check`: 12235 pass, 4 skip, 0 fail, 46106 expect() calls, 723 files):

- `(4b)` the sweep enqueues a conversation whose customer spoke *before* arming and whose reply came *after*;
- `(4c)` inbound NULL plus a reply after arming is enqueued, and the customer's column stays NULL;
- `(4d)` the step-0 handler, not just the sweep, serves a conversation whose only date is our reply;
- `(4e)` with both dates present, the more recent one decides, in the handler;
- the debounce claim stamps `last_replied_at`, including when the message id is **below** the mark.

**Mutation battery**, 7 mutants, 7 killed, no survivors: the SQL fence reverted to `last_inbound_at`, the `IS NOT NULL` clause narrowed to the customer's column, `silenceStartedAt` collapsed to each of its two arguments, the handler's re-check reverted, the episode test reverted, and the timestamp stamp deleted.

**Proved live** (before the fix, on a deployed instance): a status-change webhook creates the mirror row with `last_inbound_at` NULL; those conversations were never enqueued by the sweep.

**Not validated**: the ladder firing end to end on a live conversation recovered this way — that needs this change deployed plus the configured 24h wait. Also unexercised here: nothing changes for a conversation whose row predates the column (both fences fall back to `last_inbound_at`, which is the current behaviour).

**Adjacent, deliberately out of scope**: on an official WhatsApp inbox, a conversation with `last_inbound_at` NULL cannot prove the 24h service window is open, so the nudge takes the conservative branch (template, else private note) rather than a free-form send the provider would reject. Test `(4d)` asserts the note for that reason. On channels without a service window the same path sends the message.

Fixes #750


gabrieljablonski:
## The decisions this round took without asking

1. **`GREATEST`, not `COALESCE`, in the fence.** Coalescing would take a stale reply and fence out a conversation whose customer spoke *after* arming, a regression in the opposite direction. `GREATEST` ignores NULL in Postgres, so the fallback to `last_inbound_at` is exactly what every row written before this column already had.
2. **The timestamp is stamped unconditionally, beside an id write that is conditional.** The id is a watermark and must not move backwards; the instant is not a watermark, and a claim below the mark is still our side speaking.
3. **The ladder cannot feed itself on the new column.** The nudge does not call `claimReplyBurst`, so only a real reply moves it, and `isNewFollowUpEpisode` still ends the episode when the customer speaks.
4. **The 24h service window is out of scope.** On an official WhatsApp inbox a NULL inbound cannot prove the window is open, so the nudge takes the conservative branch. Not a demonstrated defect, so no issue was opened.
5. **Rejected, twice: stamping only after a confirmed delivery.** `runtime.ts` documents that the claim is taken before the send and never given back, and that releasing it on failure belongs to an issue of its own. `lastRepliedMessageId` beside this column already says "we spoke" for a failed send; moving only the instant would hand the three fence readers two contradicting answers about one event. The reasoning now sits at the stamp itself.
6. **The console yields to a fresh episode only when our own reply opened it.** The customer-reply half predates this change and is filed as #752.

## The one question that is not mine to answer (s7)

The holdout's s7 asks that a re-engagement **without write scope** be refused and open no episode. What was measured, both ways:

- A tenant token with role `AGENT` — which the product treats as read-only (it gets `mcp:read` and is refused by `/api/v1/api-keys`) — **entered** `POST /v1/conversations/:id/reengage` and reached the write service, stopping only at an unreachable Chatwoot. The route carries `requireAuth: true` and nothing finer.
- But every tenant API key is minted with a fixed `TENANT_ADMIN` role (`FIXED_ROLE`, "fine-grained scopes deferred"), REST only accepts an API key as Bearer, and the read-only principal the product actually issues (MCP with `mcp:read`) **is** refused: the tool is not even listed, and the call returns `isError`.

Clauses (b), (c) and (d) of the scenario were measured satisfied, the new column stayed NULL included. Deciding between the two readings means deciding whether a hand-written key row with role `AGENT` counts as the token the scenario's `given` describes. That is policy, not measurement, which is why this PR is parked rather than merged or failed.

